PR #92, opened 2015-09-01
Added Skaia War Fork Grist Cost and Combination Recipes

trielt1:


kirderf1:
With the damage it does right now, the grist cost should be similar to that of the emerald sword and unbreakable katana. (And less than Fear no anvil, as that does the same damage along with a chance of applying slowness to the enemy) Perhaps somewhat less than those since you can block with a sword and mine with a hammer, but there is no such special feature for forks right now. I would suggest a grist cost of 90 Build, keep the quartz and diamond amount, and another grist amount, such as... 15 Amethyst? (I'd also suggest that you change quartz or diamond to another grist type to make it less similar to the cost for unbreakable katana which uses build, quartz, diamond and ruby. Perhaps change from diamond to gold, since that grist is of the same rarity as diamond. (90 build, 30 quartz, 11 gold, and 15 amethyst)

trielt1:
Good idea. I'll add the spork along with the new grist cost (actually I've already added the spork, only didn't send it yet).

PR #97, opened 2016-01-14
Updated for Forge 1.8.9 and Added JEI Support

tgarr0:
Admittedly, the code I added for JEI support might be a bit Frankenstein'd together, and it's likely that something better could have been done with the changes made to BasicTreeDecorator (due to a WorldGenTrees parameter requirement change),  but everything appears to work perfectly.


kirderf1:
So, this would update the mod to 1.8.9, making it unusable with the standard 1.8, adds JEI support, while temporarily disabling NEI. (I guess because you did that because it haven't updated to 1.8.9 yet) The change with the tree decorator should prevent it from controlling what tree type it spawns a tree with. Sure there isn't any other method with more relevant parameters? Also, it seems that you accidentally left in a gradlew file. That should be removed or added to .gitignore

Aside from that, I think this should wait with updating it. The compiler at ci.mraof.com is not properly built for taking care of several different branches at the same time right now, and I'd rather let 1.8 be the active version to develop on for a while.

kirderf1:
I'll keep this pull request open until it get's time to either update to 1.8.9 or to an even higher minecraft version, depending on what comes first.

tgarr0:
That's totally understandable.  Most of my reasoning behind updating just involved wanting to play with a few other mods that had rushed to 1.8.9 anyway.

PR #99, opened 2016-05-04
Stuff from seriousCoolness's fork

kirderf1:
This includes:
*A more properly drawn mark on the alchemiter
*Grist costs for sord...... and Caledscratch
*Black dye || wood sword -> sord...........
*Katana && Clock -> Caledscratch

Things I want to be changed before merging:
*Looks like one pixel of the top border of the alchemiter side got painted over. So the border should be restored
*Grist costs seem a bit too high
*I do not understand the logic in having black dye in the sord....'s combination recipe.


seriousCoolness:
This should probs be deleted, It's super outdated

PR #149, opened 2017-11-15
Merge pull request #2 from mraof/large_machines

16msnow6:
Added direction state for large punch designix and started with clean…


16msnow6:
update

PR #182, opened 2019-01-29
1.4 Bug Fixes

Cibernet83:
Fixed the following nasty bugs:
-frogsGeneralShop dialogue appears to be missing
-you can't use the captcharoid while facing up/down
-the captcharoid ignores meta data
-Decor Blocks were able to connect to fences
-frog, petrification and lush desert lands don't have a planet render
-the minestuck advancements tab title is advancements.minestuck.root.title, submitted by: @Robot_Face

PS: i left in some commits by accident and i don't know how to take them out D:
     The bug fixes are supposed to be the last three
PPS: i moved the JBE stuff and Frog Temple stuff to their respective branches, so there shouldn't be anything lost when pulling my commits (i hope :p)


kirderf1:
There are a lot of other things happening in this pull request that doesn't have anything to do with this pull request. The prime target is the commit to remove unfinished features, which really just causes trouble by removing files that doesn't have to be removed, and that would just be readded again later. Then we've got the unfinished JBE stuff commit, which while I'd rather see it on another branch, it's not a big deal to include. Next we have the frog temple stuff, which while you "removed" it, it'd be more iddal to make it so that it wasn't added to the branch in the first place. Also, I'm wondering about your changes in the latest commit.
I'd suggest making sure you have a different branch on the commit before the commit that you remove the frog temple stuff, then rebranch 1.12 to the unfinished JBE stuff commit, or more preferably, the one before that, and then redo the bugfixes from that commit. I'll review your changes properly once you've done that.

PR #188, opened 2019-04-12
Bucket Fix!

SarahKirksey:
Because hard-coded stuff is bad for addons!


kirderf1:
Is this really necessary? Both methods for unlocalized names and for creative tabs are public, and will overwrite the existing values if used outside the constructor. Then we have that our own bucket class isn't really part of our api towards other mods, so there's no point in adding that new constructor for what little convenience it gives other mods that reuse our bucket class.

This addition should not be considered a fix.

PR #189, opened 2019-04-13
Minor changes

SarahKirksey:


kirderf1:
The creative tab field should be fine to restore some functionality that didn't exist before. Not ideal that the bucket fix commit is there and then reverted in the other commit, but that's a thing to keep in mind for next times.

PR #212, opened 2019-08-15
1.14 Updated block textures, music discs and voxel shapes

kirderf1:


kirderf1:
I have no opinion on the textures, and I don't know if the VoxelShapes are fine, but I'm thinking that this can only be a step forward.

PR #213, opened 2019-08-15
1.14 Loot tables that probably should've been made using a data provider but whatever we've got them here now so might as well merge them

kirderf1:


kirderf1:
I'm... just gonna merge it without checking if those files are actually fine. _Surely its fine._

PR #219, opened 2019-10-22
Ciber's (probably) last 1.12 commit

Cibernet83:
I gave a few things Grist Costs and Alchemy Recipes, including concrete, totems of undying and shulker boxes, lmk if you want any of them changed/removed.

-Cibernet

### DIFF
--- a/src/assets/minestuck/advancements/recipes/misc/blender.json
+++ b/src/assets/minestuck/advancements/recipes/misc/blender.json
@@ -1,0 +1,43 @@
+{
+  "parent": "minecraft:recipes/root",
+  "rewards": {
+    "recipes": [
+      "minestuck:blender"
+    ]
+  },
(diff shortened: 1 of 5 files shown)


kirderf1:
There are a few things, such as the surprisingly high cost for shulker shells, the possibility that use damage value flags aren't setup correctly for alchemy recipes, a lack of recipe advancement for the blender, and that you used reflection for grout instead of `ForgeRegistries.ITEMS.getValue()`. (Yeah, I know that we're using reflection for nearly all other mod recipes, I have just never taken the oppurtunity to change it)

However, in all I don't care too much about these things here because this will be dead-end code that wont pass on to 1.14+. If you want to fix anything of it (a lack of recipe advancement if anything), do so and then tell me when you're satisfied and I'll merge this.

PR #235, opened 2020-04-08
1.14

scorp1onSlayer:
Fluid fog fixed


src/main/java/com/mraof/minestuck/event/ClientEventHandler.java
```diff
@@ -77,11 +90,30 @@ public static void addCustomTooltip(ItemTooltipEvent event)
 	@SubscribeEvent
 	public static void onFogRender(EntityViewRenderEvent.FogDensity event)
 	{
-		if (event.getInfo().getFluidState().getFluid() == MSFluids.ENDER.get())
+		if (event.getInfo().getFluidState().getFluid() == MSFluids.ENDER.get() || event.getInfo().getFluidState().getFluid() == MSFluids.FLOWING_ENDER.get())
```
kirderf1, line 93: If you have the time, I'd suggest using the ender fluid tag since you're checking both fluid types

src/main/java/com/mraof/minestuck/event/ClientEventHandler.java
```diff
@@ -77,11 +90,30 @@ public static void addCustomTooltip(ItemTooltipEvent event)
...
 		{
 			event.setCanceled(true);
 			event.setDensity(Float.MAX_VALUE);
 			GlStateManager.fogMode(GlStateManager.FogMode.EXP);
 		}
 	}
+	@SubscribeEvent
+	public static void addFogColor(EntityViewRenderEvent.FogColors event)
+	{
```
kirderf1, line 102: Now here you're changing the fog color based on all blocks based on block.getColor(). However, we shouldn't be changing or influencing the fog color for blocks other than our own fluids. The simple solution would be to check if the fluid is in acollection of our own fluids, but I think it'd make more sense to create an interface with a fog color getter with whatever parameters we think is relevant, similarly to `AlchemizedColored`, `GristWildcardHolder`, `IColored `or `IOwnable`. Either way, we shouldn't modify the fog color for _just about any_ fluid block that end up returning something non-null.

PR #236, opened 2020-04-10
I made a wooden carrot, woohoo.

Nhezak:
Ehh, I don't know if this'll make it so that it's committed to Kirderf or not but I'm sure as fuck gonna try.


src/main/java/com/mraof/minestuck/item/MSFoods.java
```diff
@@ -20,6 +20,7 @@
     public static final Food FRENCH_FRY = (new Food.Builder()).hunger(1).saturation(0.1F).build();
     public static final Food STRAWBERRY_CHUNK = (new Food.Builder()).hunger(4).saturation(0.5F).build();
     public static final Food IRRADIATED_STEAK = (new Food.Builder()).hunger(4).saturation(0.4F).effect(new EffectInstance(Effects.WITHER, 100, 1), 0.9F).build();
+    public static final Food WOODEN_CARROT = (new Food.Builder()).hunger(0).saturation(0.0F).effect(new EffectInstance(Effects.INSTANT_DAMAGE, 1, 0), 1.0F).effect(new EffectInstance(Effects.INSTANT_DAMAGE, 1, 255), 0.01F).setAlwaysEdible().build();
```
kirderf1, line 23: 255 instant damage is way too much. Consider something else instead.

src/main/java/com/mraof/minestuck/data/MinestuckEnUsLanguageProvider.java
```diff
@@ -480,6 +480,9 @@ protected void addTranslations()
 		addTooltip(MSItems.ROCK_COOKIE, "It's as edible as a rock");
 		addStore(MSItems.ROCK_COOKIE, "Grey-Colored Cookie");
 		addStoreTooltip(MSItems.ROCK_COOKIE, "This here is a true wonder-cookie! Totally! And they are even cheaper than regular cookies!");
+		add(MSItems.WOODEN_CARROT, "Wooden Carrot");
+		addTooltip(MSItems.WOODEN_CARROT, "It's as edible as a plank.");
+		addStoreTooltip(MSItems.WOODEN_CARROT, "You're not sure if this thing is edible but you sure as fuck don't want to try...... probably...");
```
kirderf1, line 485: You probably would want to narrate the store tooltip as if it came from the consort selling it (not that the carrot is currently added to stores). Also, while we do have a few already in our lang file, I'd advice against using "fuck" or similar words.

src/main/java/com/mraof/minestuck/data/recipe/MinestuckCombinationsProvider.java
```diff
@@ -490,12 +490,13 @@ protected void registerRecipes(Consumer<IFinishedRecipe> consumer)
 		CombinationRecipeBuilder.of(MSItems.ROCK_COOKIE).input(Items.COOKIE).and().namedInput(Items.STONE).build(consumer);
 		CombinationRecipeBuilder.of(MSItems.ROCK_COOKIE).input(Items.COOKIE).and().namedInput(Items.COBBLESTONE).build(consumer);
 		CombinationRecipeBuilder.of(MSItems.ROCK_COOKIE).input(Items.COOKIE).and().namedInput(Items.GRAVEL).build(consumer);
+		CombinationRecipeBuilder.of(MSItems.WOODEN_CARROT).input(Items.CARROT).and().namedInput(ItemTags.PLANKS).build(consumer);
```
kirderf1, line 493: I'd say add a recipe that takes in logs too. You might also do it with sticks, wooden cactus and other wooden items, but they are not as obvious as logs.

kirderf1: Oh, and you should rerun `runData` so it generates the recipe json files.

src/main/java/com/mraof/minestuck/data/recipe/MinestuckGristCostsProvider.java
```diff
@@ -405,6 +405,7 @@ protected void registerRecipes(Consumer<IFinishedRecipe> recipeSaver)
 		GristCostRecipeBuilder.of(MSItems.ONION).grist(IODINE, 3).build(recipeSaver);
 		GristCostRecipeBuilder.of(MSItems.IRRADIATED_STEAK).grist(IODINE, 12).grist(URANIUM, 1).build(recipeSaver);
 		GristCostRecipeBuilder.of(MSItems.ROCK_COOKIE).grist(BUILD, 10).grist(MARBLE, 5).build(recipeSaver);
+		GristCostRecipeBuilder.of(MSItems.WOODEN_CARROT).grist(BUILD, 2).grist(ARTIFACT, 1).build(recipeSaver);
```
kirderf1, line 408: I disagree with the use of artifact grist here. Sure, it's less than useful, but I don't think it should classify as shitty in the way usually related to artifact grist. (also you can't _normally_ find artifact grist currently, making the grist cost less suited for that reason as well)

src/main/java/com/mraof/minestuck/item/MSFoods.java
kirderf1: Spontaneous idea: how about also increasing the saturation by a small amount? (0.1f or 0.2f)

src/main/generated/resources/data/minestuck/recipes/grist_costs/wooden_carrot.json
```diff
@@ -0,0 +1,10 @@
+{
+  "type": "minestuck:grist_cost",
+  "ingredient": {
+    "item": "minestuck:wooden_carrot"
+  },
+  "grist_cost": {
+    "minestuck:build": 2,
+    "minestuck:artifact": 1
```
kirderf1, line 8: Looks like you didn't runData after changing the grist cost

PR #240, opened 2020-04-18
Stone Tablet Fuctionality and Holopad

Cibernet83:
that's basically it.
ok, but for real, here's the changelog

-Stone Slab renamed to Stone Tablet to avoid any confusion
-You can now Right-Click the Stone Tablet to open it like a written book, holding a Carving Tool in your offhand will let you edit its contents
-Added the Holopad, it can be obtained through edit mode by creating a chain with at least 2 players or completing one of any length, it costs 4000 build grist to deploy and you can put any punched card into it to display its contents


src/main/java/com/mraof/minestuck/data/MinestuckEnUsLanguageProvider.java
```diff
@@ -603,7 +604,8 @@ protected void addTranslations()
 		addStoreTooltip(MSItems.CARVING_TOOL, "You gotta BE the marble!");
 		add(MSItems.CRUMPLY_HAT, "Crumply Hat");
 		add(MSItems.STONE_EYEBALLS, "Stone Eyeballs");
-		add(MSItems.STONE_SLAB, "Stone Slab");
+		add(MSItems.STONE_SLAB, "Stone Tablet");
+		addExtra(MSItems.STONE_SLAB, "carved", TextFormatting.GRAY+"Looks like something has been carved into it!");
```
kirderf1, line 608: I'd suggest setting the text formatting to the text component in `StoneTabletItem.addInformation()` instead of here

src/main/java/com/mraof/minestuck/item/StoneTabletItem.java
```diff
@@ -0,0 +1,59 @@
...
+	
+	@Override
+	public ActionResult<ItemStack> onItemRightClick(World worldIn, PlayerEntity playerIn, Hand handIn)
+	{
+		ItemStack stack = playerIn.getHeldItem(handIn);
+		if(worldIn.isRemote)
+		{
+			boolean canEdit = playerIn.getHeldItem(handIn == Hand.MAIN_HAND ? Hand.OFF_HAND : Hand.MAIN_HAND).isItemEqual(new ItemStack(MSItems.CARVING_TOOL));
+			String text = hasText(stack) ? stack.getTag().getString("text") : "";
+			Minecraft.getInstance().displayGuiScreen(new StoneTabletScreen(playerIn, handIn, text, canEdit));
```
kirderf1, line 48: This might cause problems with dedicated servers since you refer to a screen class and the minecraft class here. Have you tested this with dedicated servers? If it does cause a problem, consider moving the gui display to MSScreenFactories.

PR #245, opened 2020-06-08
Changed PogoMotion property based on Efficiency enchantment

LunyaticCat:


src/main/java/com/mraof/minestuck/item/weapon/PogoWeaponItem.java
```diff
@@ -36,8 +38,7 @@ public boolean hitEntity(ItemStack stack, LivingEntity target, LivingEntity play
 	
 	private double getPogoMotion(ItemStack stack)
 	{
-//		return 0.5 + EnchantmentHelper.getEnchantmentLevel(Enchantment.efficiency.effectId, stack)*0.1;
-		return pogoMotion;
+		return 0.5 + EnchantmentHelper.getEnchantmentLevel(Enchantments.EFFICIENCY, stack)*0.1;
```
kirderf1, line 41: This loses the use of `pogoMotion`. Instead of reinstating the old equation with efficiency, I'd suggest creating a new equation that combines the efficiency level with the pogoMotion factor.

LunyaticCat:
```java
return pogoMotion + (EnchantmentHelper.getEnchantmentLevel(Enchantments.EFFICIENCY, stack)*0.1);
```
is that better?

kirderf1:
I'd say so, yeah. It combines pogoMotion and the efficiency enchantment in a reasonable way, and it even retains the current pogo motion levels.
Also, I noticed a different thing that need to be fixed. As a consequence to weapon capabilities currently being a slight mess, there are two different classes for pogo weapons. So far you've done things in the `PogoWeaponItem`, however you would also need to fix this in `PogoFarmineItem`.

kirderf1:
One idea is to move the efficiency modifier to be applied during the static functions `hitEntity()` and `onItemUse()` instead of from the item calling these, as that would be less code duplication, and the stack is already being passed to those static functions.

LunyaticCat:
I don't really understand why `PogoFarmineItem` exist, the only properties it have are the same as in `PogoWeaponItem`.
Anyway `onItemUse()` do not have stack in `PogoFarmineItem`

kirderf1:
`PogoFarmineItem` currently exists because our current weapon features are implemented through class inheritance, so you had to reimplement either the pogo functionality in a class extending the farmine item, or the other way around.
PogoFarmineItem does have a stack through its context (and it actually picks it out when passing the stack to the static function. Anyways, if you take on the idea from my last comment about instead adding the enchantment modifier in the static functions, you'd fix this for both types without modifying PogoFarmineItem at all.

PR #249, opened 2020-07-08
Added more vanilla recipes

LunyaticCat:
Sorry for all those commits in the pull request, didn't manage to make it look nice


LunyaticCat:
hm, it should be "Planks && charcoal -> coal
a crossbow is visually a bow on a stick
I think of a bone as a white stick
a cobweb kinda is something that "glue" you, like the slimeblock does to blocks
and for sand && gravel, it's because glay is composed of silicium (sand) and other minerals that could be found in gravel (probably, im not geologist XD)

kirderf1:
Planks && charcoal -> coal
I don't think you'd go from charcoal to coal that way

Bow && stick -> crossbow
I think that motivation would work for a crafting recipe where items are put together, but not as much for a combination recipe

Stick || white dye -> bone
It is kind of a leap from white stick to a bone. It would probably maek more sense to keep it at a white stick, or possibly a white crayon

Slime block || string -> cobweb
I suppose that makes sense

Sand && gravel -> clay
The largest critique is that if we were to try and keep the behaivor of bit-wise operations, it would not work together with the Sand && cobblestone -> gravel recipe (because the latter recipe would imply that sand && gravel -> gravel)
One might also argue that minecraft clay would also need some sort of watery influence, though I don't think that specific argument holds enough merit here

LunyaticCat:
well, yeah we could do sand && water bucket then
the coal was odd, so I checked my note and it doesn't seems it's in it, I will remove it

kirderf1:
I'd assume that it was supposed to be
Planks && coal -> charcoal
(a recipe like that already exists but with logs)

LunyaticCat:
What do you think of this :
stick || skeleton skull -> bone
for the crossbow, I can't think of any other thing, so I may suppress this craft

kirderf1:
Yep,
stick || skeleton skull -> bone
sounds good

LunyaticCat:
then if you're ok with it, we may be able to merge it?

kirderf1:
Well, the comment about grist costs are still standing, and I don't think we got any closure on the `Sand && gravel -> clay` recipe

LunyaticCat:
oh right, for the clay I tought we said sand && water
and I had no idea we had a grist cost generator ^^' can I just remove the changes in the grist file and it will auto generate a price for the new items?

kirderf1:
Ah, forgot about that, but yeah, sand && water should be fine (make sure to change that and rerun runData)
Yeah, it should be fine for everything that's craftable as long as the ingredients get costs (though I suggest that you check the generated costs with jei for yourself to make sure that it works for the items that you want to have a grist cost for)

PR #253, opened 2020-07-31
Fixed some item bugs.

bconlon1:
- Edited DrinkableItem to add drinking behavior. This makes Faygos drinkable now.
- Made it so if an entity is killed in one hit with a fire weapon their drops will be cooked.


kirderf1:
1.14 is no longer being developed.

PR #256, opened 2020-12-05
Update uranium ore textures and generation

Dweblenod:
This is the same as the similarly named 1.15 pull request. base uranium vein size is now 5 and the veins per chunk is now 9


kirderf1:
Looks like this pull request (unintentionally) contains stuff from your other branch
What do you want to do with that?

Dweblenod:
A shame, I hadnt seen it with the commit window on IntelliJ. The content should be easy enough to resubmit

PR #258, opened 2020-12-16
Aspect weapons

Dweblenod:
Added Lorentz Distransformationer, an axe that utilizes a new weapon class called TeleportingWeaponItem that teleports the player like they used a chorus fruit when they hit an enemy while doing a jump attack


kirderf1:
Oh, tried doing a thing and squash the changes from this in intellij (since merge conflicts prevented me from doing it here), and ended up getting my name on it. Well, either way, the changes are now merged

PR #262, opened 2021-01-11
Sendificator block

Dweblenod:
Adds Sendificator block made by Caldw3ll and further help from Riotmode. Shift-right clicking the block item while in your mainhand (with nothing currently on your head) puts it on your head, killing you and dropping your characters head. Does nothing if it is on your head in Creative mode. Currently does not have item teleporting properties. Can be found rarely in general dungeon loot.


Dweblenod:
Bug whose source I cannot figure out how to fix: Right clicking on the placed block with nothing in your hand is meant to send a message once, but two messages appear

src/main/java/com/mraof/minestuck/block/machine/SendificatorBlock.java
```diff
@@ -0,0 +1,36 @@
...
+	
+	public SendificatorBlock(Properties properties, CustomVoxelShape shape)
+	{
+		super(properties, shape);
+	}
+	
+	@Override
+	@SuppressWarnings("deprecation")
+	public ActionResultType onBlockActivated(BlockState state, World worldIn, BlockPos pos, PlayerEntity player, Hand handIn, BlockRayTraceResult hit) {
+		if(!player.isSneaking() && player.getHeldItemMainhand().isEmpty()) {
```
kirderf1, line 26: You probably want to do this on server-side only

src/main/java/com/mraof/minestuck/block/machine/SendificatorBlock.java
```diff
@@ -0,0 +1,36 @@
...
+	
+	@Override
+	@SuppressWarnings("deprecation")
+	public ActionResultType onBlockActivated(BlockState state, World worldIn, BlockPos pos, PlayerEntity player, Hand handIn, BlockRayTraceResult hit) {
+		if(!player.isSneaking() && player.getHeldItemMainhand().isEmpty() && !worldIn.isRemote) {
+			//ItemStack itemStackIn = player.getHeldItem(handIn);
+			//Function will store information about item and pass it along to TileEntity, currently useless
+			player.sendMessage(new TranslationTextComponent(getTranslationKey() + "." + ACTIVATION_MESSAGE));
+			
+			return ActionResultType.SUCCESS;
```
kirderf1, line 31: You should probably return success client-side too, to mark that you actually do override any right-click behaivor

kirderf1 (commented):
Should generally be fine on the mechanics part, but are you considering adding a combination recipe, or at least a grist cost?
Also, just looking at the files, the texture seem unreasonably large. The model is probably fine, but I don't like how the texture and uv's are set up.

Dweblenod:
> Should generally be fine on the mechanics part, but are you considering adding a combination recipe, or at least a grist cost?
> Also, just looking at the files, the texture seem unreasonably large. The model is probably fine, but I don't like how the texture and uv's are set up.

I am hoping to make it a rarer item that cannot be duplicated, if I create a cost I would like it to be exorbitant

PR #268, opened 2021-01-19
Hard to test it by myself but this is one of my reccomendations

Dweblenod:
Just added a sample Shield Breaking OnHitEffect. I imagine there is a much better way to go about this. Wasnt sure how to test its use by myself. You may delete this request at any time!


kirderf1:
`item.canDisableShield()`is a thing, so I'd suggest making it a boolean in the builder/weapon class instead

PR #272, opened 2021-01-25
Throwable weapons

Dweblenod:
Adding three types of throwing weapons: Consumable, Returning, and Bouncing

Consumable projectiles(includes shuriken, sbahjarangs, and suitarangs) can be held in a stack. Right clicking causes one item from the stack to be removed and summons an entity with gravity that turns back into an item when it strikes a block or entity where it can be collected. 

Returning projectiles(chakram and umbral infiltrator) can be right clicked, causing the nonstackable item to lose durability and enter a cooldown period. The projectile entity does not have gravity, it strikes a surface and then returns from exactly the location it came from with around the same velocity. If this entity hits its thrower or a surface near the thrower the entity disappears and the items cooldown returns to near 0(almost ready to use again)

Bouncing projectiles(sorcerer's pinball) is similar to returning projectiles except it reflects off of blocks and entities 

Umbral Infilitrator has void specific properties that allows it to pass through blocks


src/main/java/com/mraof/minestuck/entity/item/ItemRenderedProjectileEntity.java
```diff
@@ -0,0 +1,80 @@
...
+	public void writeAdditional(CompoundNBT compound)
+	{
+		super.writeAdditional(compound);
+	}
+	
+	@Override
+	public void readAdditional(CompoundNBT compound)
+	{
+		super.readAdditional(compound);
+	}
```
kirderf1, line 47: Shouldn't need the write and read functions if you don't have any data to store

src/main/java/com/mraof/minestuck/entity/item/ItemRenderedProjectileEntity.java
```diff
@@ -0,0 +1,80 @@
...
+	
+	@Override
+	protected void onImpact(RayTraceResult result)
+	{
+		if(this.getThrower() instanceof PlayerEntity)
```
kirderf1, line 52: Why should everything in here only be executed if there is a thrower which is a player?

src/main/java/com/mraof/minestuck/entity/item/ItemRenderedProjectileEntity.java
Dweblenod: Will remove!

src/main/java/com/mraof/minestuck/entity/item/ItemRenderedProjectileEntity.java
Dweblenod: I think I had gotten some sort of issue with crashes, but I have since made some overhauls through the addition of another entity class so let me send that for review

src/main/java/com/mraof/minestuck/entity/MSEntityTypes.java
```diff
@@ -101,6 +104,9 @@ public static void registerEntities(final RegistryEvent.Register<EntityType<?>>
 		register(registry, EntityType.Builder.<SbahjPosterEntity>create(SbahjPosterEntity::new, EntityClassification.MISC).size(0.5F, 0.5F).setShouldReceiveVelocityUpdates(false).setTrackingRange(10).setUpdateInterval(Integer.MAX_VALUE), "sbahj_poster");
 		register(registry, EntityType.Builder.<ShopPosterEntity>create(ShopPosterEntity::new, EntityClassification.MISC).size(0.5F, 0.5F).setShouldReceiveVelocityUpdates(false).setTrackingRange(10).setUpdateInterval(Integer.MAX_VALUE), "shop_poster");
 		register(registry, EntityType.Builder.<HologramEntity>create(HologramEntity::new, EntityClassification.MISC).size(0.25F, 0.25F).setShouldReceiveVelocityUpdates(false).setTrackingRange(10).setUpdateInterval(Integer.MAX_VALUE), "hologram");
+		
+		register(registry, EntityType.Builder.<ConsumableProjectileEntity>create(ConsumableProjectileEntity::new, EntityClassification.MISC).size(0.25F, 0.25F).setTrackingRange(4).setUpdateInterval(10), "consumable_projectile");
+		register(registry, EntityType.Builder.<ReturningProjectileEntity>create(ReturningProjectileEntity::new, EntityClassification.MISC).size(0.25F, 0.25F).setTrackingRange(6).setUpdateInterval(1), "returning_projectile");
```
kirderf1, line 109: What's the background to the tracking range and update interval values?

src/main/java/com/mraof/minestuck/entity/item/ConsumableProjectileEntity.java
```diff
@@ -0,0 +1,73 @@
...
+			if(!this.world.isRemote && result.getType() == RayTraceResult.Type.ENTITY)
+			{
+				Entity entity = ((EntityRayTraceResult) result).getEntity();
+				entity.attackEntityFrom(DamageSource.causeThrownDamage(this, this.getThrower()), 2);
+			}
+			if(!throwerPlayer.isCreative() && rand.nextFloat() < 0.99F)
+			{
+				ItemEntity itemEntity = new ItemEntity(this.world, this.getPosX(), this.getPosY(), this.getPosZ(), this.getItem());
+				world.addEntity(itemEntity);
+			}else if(!throwerPlayer.isCreative() && rand.nextFloat() >= 0.99F)
```
kirderf1, line 54: I don't think it's intentional to generate a new float value here. I'd assume that you want something like
`if(not creative) {if(drop probability) drop item; else play break sound;}`

src/main/java/com/mraof/minestuck/entity/item/ReturningProjectileEntity.java
```diff
@@ -0,0 +1,105 @@
...
+	@Override
+	protected void onImpact(RayTraceResult result)
+	{
+		PlayerEntity throwerPlayer = (PlayerEntity) this.getThrower();
+		if(throwerPlayer != null){
+			if(this.world.getEntitiesWithinAABB(PlayerEntity.class, getBoundingBox().grow(5)).contains(throwerPlayer)){
+				resetThrower();
+			}
+			++bounce;
+			this.setMotion(this.getMotion().x*-1.1, this.getMotion().y*-1.1, this.getMotion().z*-1.1);
```
kirderf1, line 48: Could be simplified to `setMotion(getMotion().scale(-1.1))`

src/main/java/com/mraof/minestuck/entity/item/ReturningProjectileEntity.java
```diff
@@ -0,0 +1,105 @@
...
+import net.minecraft.nbt.CompoundNBT;
+import net.minecraft.network.IPacket;
+import net.minecraft.util.DamageSource;
+import net.minecraft.util.SoundCategory;
+import net.minecraft.util.SoundEvents;
+import net.minecraft.util.math.*;
+import net.minecraft.world.World;
+import net.minecraftforge.fml.network.NetworkHooks;
+
+public class ReturningProjectileEntity extends ProjectileItemEntity implements RendersAsItem
```
kirderf1, line 19: There are some standard formatting inconsistencies here, so make sure to reformat the code once you're done

src/main/java/com/mraof/minestuck/entity/item/ReturningProjectileEntity.java
```diff
@@ -0,0 +1,105 @@
...
+	public ReturningProjectileEntity(EntityType<? extends ReturningProjectileEntity> type, LivingEntity livingEntityIn, World worldIn)
+	{
+		super(type, livingEntityIn, worldIn);
+	}
+	
+	@Override
+	protected void onImpact(RayTraceResult result)
+	{
+		PlayerEntity throwerPlayer = (PlayerEntity) this.getThrower();
+		if(throwerPlayer != null){
```
kirderf1, line 43: Again, why a player thrower requirement for all of the below?

src/main/java/com/mraof/minestuck/entity/item/ReturningProjectileEntity.java
```diff
@@ -0,0 +1,105 @@
...
+import net.minecraft.util.SoundCategory;
+import net.minecraft.util.SoundEvents;
+import net.minecraft.util.math.*;
+import net.minecraft.world.World;
+import net.minecraftforge.fml.network.NetworkHooks;
+
+public class ReturningProjectileEntity extends ProjectileItemEntity implements RendersAsItem
+{
+	private int cooldownTicks;
+	private int bounce;
```
kirderf1, line 22: Are these uninteresting to save? Or is this entity not meant to be serialized?

src/main/java/com/mraof/minestuck/entity/item/ReturningProjectileEntity.java
```diff
@@ -0,0 +1,105 @@
...
+				if(entity != throwerPlayer)
+					entity.attackEntityFrom(DamageSource.causeThrownDamage(this, this.getThrower()), 5);
+				if(entity == throwerPlayer){
+					resetThrower();
+				}
+			} else if (!this.world.isRemote && result.getType() == RayTraceResult.Type.BLOCK){
+				this.world.playSound(null, this.getPosX(), this.getPosY(), this.getPosZ(), SoundEvents.ITEM_SHIELD_BLOCK, SoundCategory.PLAYERS, 0.6F, 4.0F);
+			}
+			
+			if(throwerPlayer instanceof ServerPlayerEntity){
```
kirderf1, line 61: Why check for if they are a server player when you don't use the fact that they are a server player? It seems like this expression was meant to be a `!world.isremote`

src/main/java/com/mraof/minestuck/entity/item/ReturningProjectileEntity.java
kirderf1: Serialized as in saved to/loaded from the world file

src/main/java/com/mraof/minestuck/entity/item/ReturningProjectileEntity.java
```diff
@@ -0,0 +1,105 @@
...
+	protected void onImpact(RayTraceResult result)
+	{
+		PlayerEntity throwerPlayer = (PlayerEntity) this.getThrower();
+		if(throwerPlayer != null){
+			if(this.world.getEntitiesWithinAABB(PlayerEntity.class, getBoundingBox().grow(5)).contains(throwerPlayer)){
+				resetThrower();
+			}
+			++bounce;
+			this.setMotion(this.getMotion().x*-1.1, this.getMotion().y*-1.1, this.getMotion().z*-1.1);
+			if(!this.world.isRemote && result.getType() == RayTraceResult.Type.ENTITY)
```
kirderf1, line 49: There's a number of if-statements that check that the world is not remote, such that it might be reasonable to just extract it into one if-statement encompassing the rest

src/main/java/com/mraof/minestuck/entity/item/ReturningProjectileEntity.java
```diff
@@ -0,0 +1,105 @@
...
+					entity.attackEntityFrom(DamageSource.causeThrownDamage(this, this.getThrower()), 5);
+				if(entity == throwerPlayer){
+					resetThrower();
+				}
+			} else if (!this.world.isRemote && result.getType() == RayTraceResult.Type.BLOCK){
+				this.world.playSound(null, this.getPosX(), this.getPosY(), this.getPosZ(), SoundEvents.ITEM_SHIELD_BLOCK, SoundCategory.PLAYERS, 0.6F, 4.0F);
+			}
+			
+			if(throwerPlayer instanceof ServerPlayerEntity){
+				cooldownTicks = throwerPlayer.getCooldownTracker().hashCode();
```
kirderf1, line 62: I can't tell what this cooldownTicks code block is meant to do. What is the hashcode here for?

src/main/java/com/mraof/minestuck/item/MSItems.java
```diff
@@ -792,8 +799,15 @@ public static void registerItems(final RegistryEvent.Register<Item> event)
...
+		
 		registry.register(new WeaponItem(ItemTier.IRON, 7, -2.9F, 1.0F, MSItemTypes.MISC_TOOL, new Item.Properties().group(MSItemGroup.WEAPONS)).setRegistryName("bident"));
-
+		registry.register(new ConsumableProjectileWeaponItem(new Item.Properties().group(MSItemGroup.WEAPONS)).setRegistryName("shuriken"));
+		registry.register(new ConsumableProjectileWeaponItem(new Item.Properties().group(MSItemGroup.WEAPONS)).setRegistryName("clubs_suitarang"));
+		registry.register(new ConsumableProjectileWeaponItem(new Item.Properties().group(MSItemGroup.WEAPONS)).setRegistryName("diamonds_suitarang"));
+		registry.register(new ConsumableProjectileWeaponItem(new Item.Properties().group(MSItemGroup.WEAPONS)).setRegistryName("hearts_suitarang"));
+		registry.register(new ConsumableProjectileWeaponItem(new Item.Properties().group(MSItemGroup.WEAPONS)).setRegistryName("spades_suitarang"));
+		registry.register(new ReturningProjectileWeaponItem(new Item.Properties().group(MSItemGroup.WEAPONS).maxDamage(250)).setRegistryName("chakram"));
```
kirderf1, line 809: Perhaps add some property or something else to differentiate the consumable projectile weapons?

src/main/java/com/mraof/minestuck/entity/item/ReturningProjectileEntity.java
```diff
@@ -0,0 +1,105 @@
...
+			if(bounce > 15){
+				resetThrower();
+			}
+		}
+	}
+	
+	public void resetThrower(){
+		PlayerEntity throwerPlayer = (PlayerEntity) this.getThrower();
+		if(throwerPlayer != null){
+			throwerPlayer.getCooldownTracker().setCooldown(MSItems.CHAKRAM, 5);
```
kirderf1, line 77: The item is hardcoded. Perhaps you could use `func_213882_k().getItem()` instead?

src/main/java/com/mraof/minestuck/entity/item/ReturningProjectileEntity.java
```diff
@@ -0,0 +1,105 @@
...
+import net.minecraft.item.Item;
+import net.minecraft.nbt.CompoundNBT;
+import net.minecraft.network.IPacket;
+import net.minecraft.util.DamageSource;
+import net.minecraft.util.SoundCategory;
+import net.minecraft.util.SoundEvents;
+import net.minecraft.util.math.*;
+import net.minecraft.world.World;
+import net.minecraftforge.fml.network.NetworkHooks;
+
```
kirderf1, line 18: I am not entierly clear on how this entity is meant to behave. Perhaps you could write a javadoc on the code describing the projectiles intended behavior?

kirderf1: If you want to, you could write one for the other entity class as well
But I think it's primarily valuable to have for this class

kirderf1 (changes requested):

src/main/java/com/mraof/minestuck/entity/MSEntityTypes.java
Dweblenod: The change in tracking range is probably unnecessary, but the update interval values seem to affect how visible it is. A high update interval seems to cause it to visually pass through blocks or appear invisible making it hard to keep track of

src/main/java/com/mraof/minestuck/entity/item/ReturningProjectileEntity.java
Dweblenod: Good to know! Will make this happen

src/main/java/com/mraof/minestuck/entity/item/ReturningProjectileEntity.java
Dweblenod: The game crashes on impact without it, or at least thats what I remember happens

src/main/java/com/mraof/minestuck/entity/item/ReturningProjectileEntity.java
Dweblenod: Its another way of checking for the remaining cooldown ticks the player has for this item, the exact way I did it may not be optimal but it was just a way it happened to work

src/main/java/com/mraof/minestuck/item/MSItems.java
Dweblenod: I was having issues making a builder or adding properties, I will likely need your help setting that up so I can change velocity, accuracy, bounce(if returning projectile), and damage independently

src/main/java/com/mraof/minestuck/entity/item/ReturningProjectileEntity.java
Dweblenod: I could make a description on the discord in showcase if you would like!

src/main/java/com/mraof/minestuck/entity/MSEntityTypes.java
Dweblenod: Should I make any changes to the values I made?

kirderf1: Eh, it was more of noticing how the values didn't seem to be picked from any similar minestuck or vanilla entity type, so I want to make sure that they weren't just arbitrarily picked.

src/main/java/com/mraof/minestuck/entity/item/ReturningProjectileEntity.java
kirderf1: It appear to only be used as a nonnull player for the cooldownTicks part (and resetThrower(), but it gets and checks the player on its own).
Ideally I'd think that the entity should still work in most manners, even with a null thrower or a non-player thrower. But what are your thoughts on this?
Independently of if you let more stuff happen with a non-player thrower or not, you should probably check that the type of the thrower is an instance of EntityPlayer before casting it to one.

src/main/java/com/mraof/minestuck/entity/MSEntityTypes.java
kirderf1: Eh, it was more of noticing how the values didn't seem to be picked from any similar minestuck or vanilla entity type, so I want to make sure that they weren't just arbitrarily picked.

src/main/java/com/mraof/minestuck/entity/item/ReturningProjectileEntity.java
kirderf1: How does `hashCode()` get the remaining cooldown ticks here?

src/main/java/com/mraof/minestuck/entity/item/ReturningProjectileEntity.java
Dweblenod: Just now I tried to make it work independently of a player but I could not get it to work fully. It would make sense to have it work with non-players, but because many of the properties are dependent on the player and the specific item, I dont seem to have the coding prowess necessary to solve the issue. It failed to render among other things

src/main/java/com/mraof/minestuck/item/MSItems.java
Dweblenod: I figured out how to make them parameters!

src/main/java/com/mraof/minestuck/entity/item/ReturningProjectileEntity.java
Dweblenod: The functionality I thought cooldownTicks gave was nonexistent and attributable to other code. Removed

src/main/java/com/mraof/minestuck/entity/item/ReturningProjectileEntity.java
Dweblenod: I do not understand the importance well enough to know

Dweblenod: Serialized!

src/main/java/com/mraof/minestuck/entity/item/ReturningProjectileEntity.java
Dweblenod: With help, player dependency has been removed!

PR #275, opened 2021-02-08
Adding more variability to AspectBasedWeaponItem

Dweblenod:
Added parameter that allows for an effect to be given to the target in addition to the attacker, as well as a critical hit condition which means that crits cause the effect to occur more often. Nosferatu spoon now gives the target the hunger effect, while the Clown Club and Angel Apocalypse weapons just give an inert version of the same effect a successful attack grants the attacker(Gives 0 second 0 amplifier effect of Strength and Resistance respectively). 

Effect on target may(not confirmed) cancel preexisting effects of the same type if it is of the same amplifier level(so if a target has base level resistance and is hit with Angel Apocalypse, the inert resistance effect angel apocalypse gives may remove the original effect).


kirderf1:
It seems like the clown club and angel apocalypse aren't really meant to have enemy effects. Perhaps you could make the enemy effect nullable, and then pass null in for those two? Also, consider possibly implementing this for the weapon properties branch instead as these weapon classes are being replaced.

Dweblenod:
> It seems like the clown club and angel apocalypse aren't really meant to have enemy effects. Perhaps you could make the enemy effect nullable, and then pass null in for those two? Also, consider possibly implementing this for the weapon properties branch instead as these weapon classes are being replaced.

I had originally made it so it passed in null parameters. However I likely misunderstood how to use them. The use of null caused crashes on effect application in the worse set of usage and gave an error on the best iteration of code. Would I use something like @Nullable to remove that error?

kirderf1:
`@Nullable` is just a marker. If for example the supplier might be null, you need to check if it's null before deciding to use it. For example, `if(enemyEffect != null) target.addPotionEffect(enemyEffect.get());`

Dweblenod:
> `@Nullable` is just a marker. If for example the supplier might be null, you need to check if it's null before deciding to use it. For example, `if(enemyEffect != null) target.addPotionEffect(enemyEffect.get());`

I had done a check for that, as far as I remember it was almost exactly that! Could I have done something strange with how I put the null value in the item registry?

kirderf1:
Well, what was the error?

Dweblenod:
I will try merging it into weapon properties, with a null value, and potentially scrap this request and work from there so that we can see what issues there are while I reattempt

PR #276, opened 2021-02-09
Frog quests

Dweblenod:
A work in progress that adds new progress towards completing Frog(Space aspect) quests, while furthering the effort to build land mythology. The request may eventually include Frog Temples and/or primitive ectobiology equipment if they are not already added by another commit. As of sending request, the commits include several frog blocks, frog eggs, and expository consort land dialogue.


kirderf1:
Looks pretty good so far, and I'm interested to see where it goes from here

PR #281, opened 2021-02-21
Reworked Aspect powers

Dweblenod:
LIKELY TO BE REWORKED FROM SCRATCH

Creating a better range of Aspect powers, that will eventually incorporate class so that each classpect has unique talents

Three keybinds used: UserEffect, Positive[Target]Effect, Negative[Target]Effect


Dweblenod:
This rework itself is intended to be reworked before release, so there is a much more solid foundation of code to work from

PR #284, opened 2021-03-02
End terrain land terrain changes

Dweblenod:
End Grass:
- New tall end grass was added and so old end grass name was changed to end grass block
- Tall end grass and end grass blocks now appear all over rough terrain with rarer spots in normal terrain
- End grass blocks no longer grow to overtake end stone
- Tall end grass has particle effects similar to end grass blocks as well as random teleportation

Features/structures:
- Added chorus plants, which now appear rarely in rough terrain
- Removed coarse dirt/gravel veins and added coarse end stone veins
- Created Surface Disk feature which allows patches of blocks to appear on the surface


src/main/java/com/mraof/minestuck/world/lands/terrain/EndLandType.java
```diff
@@ -85,21 +81,28 @@ public void setProperties(LandProperties properties)
...
 	{
-		if(biome.type == BiomeType.NORMAL)
+		if(biome.type != BiomeType.OCEAN)
 		{
 			biome.addFeature(GenerationStage.Decoration.VEGETAL_DECORATION, MSFeatures.END_TREE.withConfiguration(EndTree.getDefaultConfig().build()).withPlacement(Placement.COUNT_EXTRA_HEIGHTMAP.configure(new AtSurfaceWithExtraConfig(2, 0.1F, 1))));
+		}
+		
+		if(biome.type == BiomeType.NORMAL)
+		{
+			biome.addFeature(GenerationStage.Decoration.TOP_LAYER_MODIFICATION, MSFeatures.SURFACE_DISK.withConfiguration(new SphereReplaceConfig(MSBlocks.END_GRASS_BLOCK.getDefaultState(), 5, 1, Lists.newArrayList(blocks.getBlockState("surface"), MSBlocks.END_GRASS_BLOCK.getDefaultState()))).withPlacement(Placement.COUNT_TOP_SOLID.configure(new FrequencyConfig(1))));
```
kirderf1, line 91: Because it's placing a grass block, it might be more appropriate to make a feature that strictly places blocks if they have an air block/non-solid above them (or as an alternative, gets its positions for placing blocks as the topmost block in an xz-area), since the surface block is not just used for the surface. Though I suppose it will only matter in some unusual cases as long as the upper blocks differ from the surface blocks. So if you choose to leave it be, make sure to leave a TODO-comment about it.

src/main/java/com/mraof/minestuck/world/lands/terrain/EndLandType.java
```diff
@@ -46,7 +41,8 @@ public EndLandType()
 	public void registerBlocks(StructureBlockRegistry registry)
 	{
 		registry.setGroundState(Blocks.END_STONE.getDefaultState(), MSFillerBlockTypes.END_STONE);
-		registry.setBlockState("surface", MSBlocks.END_GRASS.getDefaultState());
+		registry.setBlockState("surface", Blocks.END_STONE.getDefaultState());
```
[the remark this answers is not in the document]
kirderf1: Oh, another thing to consider: might ores spawn on the surface with this, and is that acceptable?

src/main/java/com/mraof/minestuck/block/MSBlocks.java
```diff
@@ -274,7 +275,7 @@ public static void registerBlocks(final RegistryEvent.Register<Block> event)
 		registry.register(new Block(Block.Properties.create(Material.ROCK).hardnessAndResistance(1.5F, 6.0F).harvestTool(ToolType.PICKAXE).harvestLevel(0)).setRegistryName("flowery_mossy_cobblestone"));
 		registry.register(new Block(Block.Properties.create(Material.ROCK).hardnessAndResistance(1.5F, 6.0F).harvestTool(ToolType.PICKAXE).harvestLevel(0)).setRegistryName("flowery_mossy_stone_bricks"));
 		registry.register(new Block(Block.Properties.create(Material.ROCK, MaterialColor.SAND).hardnessAndResistance(3.0F, 9.0F).harvestTool(ToolType.PICKAXE).harvestLevel(0)).setRegistryName("coarse_end_stone"));
-		registry.register(new EndGrassBlock(Block.Properties.create(Material.ROCK, MaterialColor.PURPLE).hardnessAndResistance(3.0F, 9.0F).harvestTool(ToolType.SHOVEL).tickRandomly()).setRegistryName("end_grass"));
+		registry.register(new EndGrassBlock(Block.Properties.create(Material.ROCK, MaterialColor.PURPLE).hardnessAndResistance(3.0F, 9.0F).harvestTool(ToolType.SHOVEL).tickRandomly()).setRegistryName("end_grass_block"));
```
kirderf1, line 278: Since you're changing a registry name here, and I'd say that we want to keep some level of backwards-compability at this time, I'd like you to map the old name back to the block during the MissingMappings event, similarly to what we're already doing in MSBiomes.

kirderf1 (changes requested):

src/main/java/com/mraof/minestuck/world/lands/terrain/EndLandType.java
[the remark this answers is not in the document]
Dweblenod: Coarse end stone is in a strange place right now, agreed. I think it would be preferable if it was the filler block of the world and that the underground was made of it, with a thin layer of end stone and then the end grass blocks. But this would mean modifying the end ore textures so it might be worth debating. If the underground were made of coarse end stone, it would get rid of the issue you are describing with the potentially visible-from-the-surface-ore

src/main/java/com/mraof/minestuck/block/MSBlocks.java
Dweblenod: I will try and do this!

src/main/java/com/mraof/minestuck/world/lands/terrain/EndLandType.java
Dweblenod: I have not noticed any major issues with this so far, but it is definitely worth making the change

src/main/java/com/mraof/minestuck/world/lands/terrain/EndLandType.java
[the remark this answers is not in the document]
kirderf1: Personally I don't consider end stone ore -> coarse end stone ore switch to be a big deal, but I suppose you could check what other people think as well.

src/main/java/com/mraof/minestuck/world/gen/feature/SurfaceDiskFeature.java
```diff
@@ -0,0 +1,57 @@
...
+ * A version of the {@link net.minecraft.world.gen.feature.SphereReplaceFeature}, but without the need to be placed in water
+ */
+public class SurfaceDiskFeature extends Feature<SphereReplaceConfig>
+{
+	public SurfaceDiskFeature(Function<Dynamic<?>, ? extends SphereReplaceConfig> configFactoryIn)
+	{
+		super(configFactoryIn);
+	}
+	
+	public boolean place(IWorld worldIn, ChunkGenerator<? extends GenerationSettings> generator, Random rand, BlockPos pos, SphereReplaceConfig config) {
```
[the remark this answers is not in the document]
Dweblenod: Done!

src/main/java/com/mraof/minestuck/world/lands/terrain/EndLandType.java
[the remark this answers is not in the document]
Dweblenod: Ground filler block is now coarse end stone, the ore textures have been made to match

src/main/java/com/mraof/minestuck/block/MSBlocks.java
Dweblenod: The MissingMappings-backwards compatibility does not appear to be working, so a name revert may be the best solution for the moment

src/main/java/com/mraof/minestuck/world/lands/terrain/EndLandType.java
Dweblenod: I cannot see a way to check for both the given block and if the block above it is air within the SphereReplaceConfig, I may leave this as a TODO

kirderf1: With the surface block now being the same as the upper block, this has become more important. Also, it wouldn't happen in SphereReplaceConfig but rather in a feature class. For example a class similar to SurfaceDiskFeature, or in SurfaceDiskFeature but with a boolean parameter like PillarFeature.

Dweblenod: Changes made! Appears to work well

kirderf1:
Looking at the current version of the SurfaceDiskFeature, there are some problems with how it is written that I'd like to have fixed. Gonna make a pull request as I feel that it's non-obvious how to explain it.

src/main/java/com/mraof/minestuck/world/lands/terrain/EndLandType.java
```diff
@@ -83,23 +79,30 @@ public void setProperties(LandProperties properties)
 	}
 	
 	@Override
-	public void setBiomeSettings(LandWrapperBiome biome, StructureBlockRegistry blocks)
+	public void setBiomeSettings(LandWrapperBiome biome, StructureBlockRegistry blocks) //TODO Try and find a method of checking that a non solid block exists above SurfaceDisk features as their use here is to lay down grass
```
kirderf1, line 82: Ah yeah, this has been solved now, so that comment should not stick around.

PR #288, opened 2021-03-20
Void aspect mace and Breath aspect cane

Dweblenod:
Introduces two new aspect based weapons

-Desolator Mace: Void aspect mace that ignores armor when damaging(if Void), alchemy recipe = mace || flint

-Zephyr Cane: Breath aspect cane that propels the player in the direction they are facing(if Breath), alchemy recipe = iron tipped cane || feather or iron tipped cane || phantom membrane

As of creation of pull request the newly created PropelEffect class that the zephyr cane utilizes has a frustrating but mild bug that prevents velocity from changing the first instance of the item's use upon entering the world(still makes a sound, takes away item durability, etc)


src/main/java/com/mraof/minestuck/event/ServerEventHandler.java
```diff
@@ -154,6 +155,13 @@ public static void onEntityAttack(LivingHurtEvent event)
 					double modifier = PlayerSavedData.getData(player).getEcheladder().getUnderlingDamageModifier();
 					event.setAmount((float) (event.getAmount() * modifier));
 				}
+				
+				Title title = PlayerSavedData.getData(player).getTitle();
+				boolean isVoid = title != null && title.getHeroAspect() == EnumAspect.VOID;
+				if (player.getHeldItemMainhand().getItem() == MSItems.DESOLATOR_MACE && isVoid)
```
kirderf1, line 161: I'm not sure if this accurately catches only damage dealt by the weapon, as all we've confirmed about the damage source is that the player caused it. There could be other ways for players to be a source of damage other than direct weapon attacks.

Also, it'd be better design to go through some property on the weapon item instead of hardcoding this here.

src/main/java/com/mraof/minestuck/event/ServerEventHandler.java
```diff
@@ -154,6 +155,13 @@ public static void onEntityAttack(LivingHurtEvent event)
...
+				{
+					event.getSource().setDamageBypassesArmor();
```
kirderf1, line 163: I'm a bit wary of modifying the damage source like this, as it is normally set during initialization of a damage source. It's possible that this may cause unexpected behavior in special circumstances. It might be fine most of the time, but to be on the safe side I'd prefer a different method.

src/main/java/com/mraof/minestuck/item/weapon/PropelEffect.java
```diff
@@ -0,0 +1,68 @@
...
+import net.minecraft.util.math.Vec3d;
+import net.minecraft.world.World;
+
+public class PropelEffect implements ItemRightClickEffect
+{
+	public static final PropelEffect BREATH_PROPEL = new PropelEffect(3, EnumAspect.BREATH);
+	
+	private final double velocity;
+	private final EnumAspect aspect;
+	private Title title;
```
kirderf1, line 19: This field is not appropriate for this type of class. Use of it should not be necessary, and there's a high risk of unexpected behavior such as leakage between onRightClick() calls (in this case between sides). You should keep it as a local variable to avoid this.

kirderf1 (changes requested):

src/main/java/com/mraof/minestuck/event/ServerEventHandler.java
Dweblenod: What if it checks if the active item is the same as well?

I would make it a weapon property, although I did not want to have to override the damage value and I did not want to make it give damage in addition to what the player is already inflicting, I wanted it to override the damage correctly. If there is a succinct way of changing it, it is absolutely worth doing so.

src/main/java/com/mraof/minestuck/item/weapon/PropelEffect.java
Dweblenod: This is referring to the Title variable?

kirderf1: Yes

src/main/java/com/mraof/minestuck/event/ServerEventHandler.java
Dweblenod: Issue mitigated by the use of a OnHitEffect damage bonus instead of controlling the action damage action

src/main/java/com/mraof/minestuck/item/weapon/PropelEffect.java
Dweblenod: It is now a local variable, and the reason for its previous use in the field has been solved!

src/main/java/com/mraof/minestuck/item/weapon/OnHitEffect.java
```diff
@@ -167,6 +167,39 @@ static OnHitEffect setOnFire(int duration)
 		return (itemStack, target, attacker) -> target.setFire(duration);
 	}
 	
+	static OnHitEffect armorBypassingDamageMod(float additionalDamage, EnumAspect aspect)
+	{
+		return (stack, target, attacker) -> {
+			DamageSource source;
+			float damage = additionalDamage * 3.3F;
+			
+			if(attacker instanceof PlayerEntity)
```
kirderf1, line 176: Maybe change this to a `ServerPlayerEntity` since that's what you're using in the cast below. (and maybe add a `ServerPlayerEntity player = (ServerPlayerEntity) attacker;` so that you don't have to do multiple casts.

src/main/java/com/mraof/minestuck/item/weapon/PropelEffect.java
```diff
@@ -0,0 +1,75 @@
...
+	
+	private double getVelocityMod()
+	{
+		return velocity;
+	}
+	
+	void propelAction(PlayerEntity player, ItemStack stack, double velocity, Hand hand)
+	{
+		Title title = null;
+		if(player instanceof ClientPlayerEntity)
```
kirderf1, line 44: Since the `ClientPlayerEntity` class doesn't exist on a dedicated server, you might want to replace this with `player.world.isRemote`

src/main/java/com/mraof/minestuck/item/weapon/PropelEffect.java
```diff
@@ -0,0 +1,75 @@
...
+		{
+			title = ClientPlayerData.getTitle();
+		}
+		
+		if(player instanceof ServerPlayerEntity)
```
kirderf1, line 49: Not a big deal, but you could make this an else-if because this should (and either way will) only happen when the if-statement below doesn't happen

src/main/java/com/mraof/minestuck/item/weapon/PropelEffect.java
```diff
@@ -0,0 +1,75 @@
...
+		{
+			title = PlayerSavedData.getData((ServerPlayerEntity) player).getTitle();
+			if(player.getCooldownTracker().getCooldown(stack.getItem(), 1F) <= 0 && title != null && title.getHeroAspect() == aspect)
+				propelActionSound(player.world, player);
```
kirderf1, line 53: Might I ask why this is here, and not in the code block below?

kirderf1 (changes requested):
Gotta appreciate how you're managing damage and damage source in armorBypassingDamageMod().
Still some things I want to have changed.

src/main/java/com/mraof/minestuck/item/weapon/PropelEffect.java
Dweblenod: It is not important, I chose to put it here because I thought the sound only plays on server side

src/main/java/com/mraof/minestuck/item/weapon/PropelEffect.java
Dweblenod: Done

src/main/java/com/mraof/minestuck/item/weapon/PropelEffect.java
Dweblenod: Switched out and still works

src/main/java/com/mraof/minestuck/item/weapon/PropelEffect.java
kirderf1: I think it might be better code form to have it in the code block below together with the other effects, separated from the code to obtain the title. Though I suppose it's not a big deal, and I'm fine merging it as it is if you don't want to bother with it.

Dweblenod: I think it helps mildly, in that its obvious in which conditions it would be activating. So I am fine with merging

PR #293, opened 2021-03-31
Shieldbreaking and sweep changes

Dweblenod:
This request removes SwordItem as the class our weapons (WeaponItem) extend in order to eliminate the unconditional sweep effect all our weapons have. Instead, two new OnHitEffects are created that handle the technical nuance of sweeping weapons(swords in vanilla) and shield disabling weapons(axes in vanilla) called SHIELD_DISABLE and SWEEP respectively.

Currently our swords and fans will have sweep while our axes and sickles will have shield disabling. These should help balance gameplay use of the weapons in addition to the changes made with the merging of the weapon_properties branch damage and speed adjustments. The final integral merge commit that will need to come after this is weapon grist cost rebalancing


src/main/java/com/mraof/minestuck/item/weapon/OnHitEffect.java
```diff
@@ -145,6 +148,41 @@
 		target.attackEntityFrom(source, rng);
 	};
 	
+	OnHitEffect SHIELD_DISABLE = (stack, target, attacker) -> {
```
kirderf1, line 151: As said elsewhere, forge seem to already have a hook for this accessible by implementing/overriding `canDisableShield()` in the item class. If that hook works as intended, I'd suggest you make use of that one instead, perhaps by adding a "disables shield" boolean in the weapon properties builder.

kirderf1 (changes requested):

src/main/java/com/mraof/minestuck/item/weapon/WeaponItem.java
```diff
@@ -42,14 +48,17 @@
 	@Deprecated
 	public WeaponItem(IItemTier tier, int attackDamage, float attackSpeed, float efficiency, @Nullable MSToolType toolType, Properties properties)
 	{
-		this(new Builder(tier, attackDamage, attackSpeed).efficiency(efficiency).set(toolType), properties);
+		this(new Builder(tier, attackDamage, attackSpeed).efficiency(efficiency).disableShield().set(toolType), properties);
```
kirderf1, line 51: Adding disable shield in this constructor does not seem right. However this constructor is not meant to be used anymore, so instead of removing just `disableShield()` here, you might as well remove the entire constructor.

PR #294, opened 2021-04-08
Dweble grist rebalance

Dweblenod:
Rebalancing the cost of weapons and potentially other items, this will also include final tweaks to items for 4/13 release

Grist costs are based off of a calculation factoring damage, speed, special properties, shield disabling, sweeping, durability, material, rarity, tool type functionality, knockback, and ranged capabilities. All weapons are standardized into a universal value that different grist types are then applied to with their respective universal value considered

It now also includes bug fixes that could only be checked in a multiplayer. As well as last minute playtesting suggestions and and a custom armor model for items(currently only used on crumpled hat)


Dweblenod:
Playtesting has finished and changes due to it have been incorporated

PR #300, opened 2021-05-19
Custom Bows

Dweblenod:
This request results in the creation of bows with custom properties. Such as a bow with dedicated flame properties, bows that are faster and longer range, and bows that shoot an additional uncollectable arrow. It still uses vanilla mechanics but creates some improved mechanic variety


src/main/java/com/mraof/minestuck/item/weapon/CustomBow.java
```diff
@@ -0,0 +1,133 @@
...
+public class CustomBow extends BowItem
+{
+	private final Type type;
+	
+	public enum Type
+	{
+		FLAME, LONG, NO_GRAVITY, QUICK, RANDOM_SECOND
+	}
+	
+	public CustomBow(Item.Properties builder, Type type)
```
Dweblenod, line 24: At the moment, this uses the same code as the traditional BowItem class. All it adds is the Type enum where custom material is inserted

src/main/java/com/mraof/minestuck/item/weapon/CustomBow.java
```diff
@@ -0,0 +1,133 @@
...
+	{
+		if(entityLiving instanceof PlayerEntity)
+		{
+			PlayerEntity playerentity = (PlayerEntity) entityLiving;
+			boolean canShootFlag = playerentity.abilities.isCreativeMode || EnchantmentHelper.getEnchantmentLevel(Enchantments.INFINITY, stack) > 0;
+			ItemStack itemstack = playerentity.findAmmo(stack);
+			
+			int i = this.getUseDuration(stack) - timeLeft;
+			i = net.minecraftforge.event.ForgeEventFactory.onArrowLoose(stack, worldIn, playerentity, i, !itemstack.isEmpty() || canShootFlag);
+			if(i < 0) return;
```
Dweblenod, line 53: Because it is the default code, its readability needs improving, will change the name of these undescriptive variables

Dweblenod:
Closing because of the messy implementation, a future pull request may instead take notes from Cibernet's implementation of these same bows into Minestuck Universe

PR #304, opened 2021-07-25
lunchtop retexture

amiwyn:
https://trello.com/c/J59qjMXy/194-remodel-lunchtop-complete

image


src/main/resources/assets/minestuck/models/block/lunchtop_bsod.json
```diff
@@ -1,2415 +0,0 @@
...
-		"ground": {
-			"scale": [1, 1, 1],
-			"translation": [0, 4, 0]
-		},
-		"gui": {
-			"scale": [0.625, 0.625, 0.625],
-			"rotation": [30, 225, 0],
-			"translation": [0, 0, 0]
-		}
-	}
```
Dweblenod, line 2414: In the GUI it appears flat, facing directly towards the screen instead of in an isometric view as most blocks are (alternatively we could use an item sprite if you wished). It also floats and appears large when viewed in third person. Blockbench has a feature where you can apply a "block" preset to display which should fix this

Dweblenod (changes requested):

src/main/resources/assets/minestuck/blockstates/lunchtop.json
```diff
@@ -8,9 +8,9 @@
         "facing=east,state=on":      { "model": "minestuck:block/lunchtop_on", "y": 90 },
         "facing=south,state=on":     { "model": "minestuck:block/lunchtop_on", "y": 180 },
         "facing=west,state=on":      { "model": "minestuck:block/lunchtop_on", "y": 270 },
-        "facing=north,state=broken": { "model": "minestuck:block/lunchtop_bsod" },
-        "facing=east,state=broken":  { "model": "minestuck:block/lunchtop_bsod", "y": 90 },
-        "facing=south,state=broken": { "model": "minestuck:block/lunchtop_bsod", "y": 180 },
-        "facing=west,state=broken":  { "model": "minestuck:block/lunchtop_bsod", "y": 270 }
```
Dweblenod, line 14: Is this change from "bsod" to "off" intentional?

Dweblenod:
Overall this update is an excellent first submission! Once the minor graphical issues are resolved this appears ready for merging

Dweblenod (changes requested):
It appears that the block shape does not match the model. You need to go into the block class ComputerBlock(block/machine/ComputerBlock) and change "LUNCHTOP_SHAPE" to match the model.

src/main/resources/assets/minestuck/blockstates/lunchtop.json
amiwyn: theres no screen, so no bsod texture anymore. Ive put the off state instead

kirderf1: For this, I think it's more appropriate that the "on" model is used for the broken state, since a "broken" computer is a variant of the "on" state when it comes to functionality.

kirderf1 (changes requested):
I think it needs a small change for the "broken" state (that you get by inserting music disc 11 into the computer), but after that I think it's good to go as well!

amiwyn:
image

Dweblenod (approved):

kirderf1 (approved):

PR #305, opened 2021-07-31
Tiered dungeons

Dweblenod:
NOW OUTDATED. #345 will be the actively updated PR while this one remains to preserve review notes before 345 is completed

Expanding on skills and code developed with Frog Temples to create a tiered dungeon system to the medium. This system will not replace imp dungeons, those will become a type of untiered dungeon like planned terrain + aspect + consort dungeons

New content:

1. Items
**-Dungeon keys:** appears in land loot, used to grant access to a tiered dungeon. With this, tiered dungeons will require players to have explored the previous tier, forcing a linear progression

2. Blocks
**-Dungeon Door Interface:** requires a specific key, removes door blocks close to it
**-Dungeon Door:** Normal block that acts as visual barrier to entry at the beginning of a dungeon and occasionally for subsequent puzzles
**Various puzzle/redstone blocks**: Check Puzzle blocks pull request for information on these(https://github.com/mraof/Minestuck/pull/313)
**-Loot Block:** Gives players the contents of a loot table on right click, at which point it becomes empty/inactive

3. Dungeons

4. Effects
-Creative Shock: 
5.


src/main/java/com/mraof/minestuck/util/PlacementFunctionsUtil.java
```diff
@@ -0,0 +1,233 @@
...
+				}
+				 */
+			}
+			
+			if(yIterator >= maxBlockPosIn.getY())
+			{
+				break;
+			}
+		}
+	}
```
Dweblenod, line 103: Im not sure how to fix the issue with the staircase generating in a fragmented manner. It does seem related to bounding boxes but if thats the case I will probably need help

src/main/java/com/mraof/minestuck/block/EnumKeyType.java
```diff
@@ -0,0 +1,36 @@
...
+	
+	public static EnumKeyType fromInt(int ordinal) //converts int back into enum, used primarily in KeyItem
+	{
+		for(EnumKeyType type : EnumKeyType.values())
+		{
+			if(type.ordinal() == ordinal)
+				return type;
+		}
+		return null;
+	}
```
Dweblenod, line 35: Would it be smarter to work with a single conversion type for storing nbt as opposed to switching between instances of strings or ints with the enum?

src/main/java/com/mraof/minestuck/item/weapon/InventoryTickEffect.java
```diff
@@ -48,4 +50,16 @@ static InventoryTickEffect passiveAspectEffect(EnumAspect aspect, Supplier<Effec
...
+		if(isSelected && entityIn instanceof ServerPlayerEntity)
+		{
+			ServerPlayerEntity serverPlayerEntity = (ServerPlayerEntity) entityIn;
+			ItemStack otherStack = !serverPlayerEntity.getHeldItemMainhand().isItemEqual(stack) ? serverPlayerEntity.getHeldItemMainhand() : serverPlayerEntity.getHeldItemOffhand(); //it gives the stack in the hand that doesnt hold the tick weapon
+			if(ItemStack.areItemsEqual(MSItems.PRISMARINE_BASHER.getDefaultInstance(), otherStack))
+			{
+				MSCriteriaTriggers.SURF_N_TURF.trigger(serverPlayerEntity);
+			}
+		}
+	};
```
Dweblenod, line 64: This method works, but I am opening to optimization or alternatives. I tried picking what is likely the less popular weapon for the tick effect so that it is less likely to cause performance issues

src/main/java/com/mraof/minestuck/item/KeyItem.java
```diff
@@ -0,0 +1,75 @@
...
+		return stack;
+	}
+	
+	@Override
+	public void addInformation(ItemStack stack, @Nullable World worldIn, List<ITextComponent> tooltip, ITooltipFlag flagIn)
+	{
+		super.addInformation(stack, worldIn, tooltip, flagIn);
+		EnumKeyType keyTypeFromInt = EnumKeyType.fromInt(getKeyType(stack));
+		tooltip.add(new TranslationTextComponent("item.minestuck." + this + ".key", keyTypeFromInt));
+	}
```
Dweblenod, line 74: currently lists key type with the underscores present, how would I remove those?

src/main/java/com/mraof/minestuck/item/KeyItem.java
```diff
@@ -0,0 +1,75 @@
...
+import net.minecraft.world.dimension.Dimension;
+import net.minecraftforge.common.util.Constants;
+
+import javax.annotation.Nullable;
+import java.util.List;
+
+public class KeyItem extends Item
+{
+	public EnumKeyType keyID;
+	//public Dimension dimension = null;
```
Dweblenod, line 24: Im hoping to add dimension specificity so keys only work on the land they were generated for. May need help working on that

src/main/java/com/mraof/minestuck/event/ServerEventHandler.java
```diff
@@ -277,13 +281,70 @@ public static void onPlayerTickEvent(TickEvent.PlayerTickEvent event)
...
+			PlayerEntity player = (PlayerEntity) entityLiving;
+			
+			if(player instanceof ServerPlayerEntity && !player.isCreative() && effect == MSEffects.CREATIVE_SHOCK.get())
+			{
+				ServerPlayerEntity serverPlayerEntity = (ServerPlayerEntity) player;
+				player.abilities.allowEdit = !serverPlayerEntity.interactionManager.getGameType().hasLimitedInteractions();
+				
+				StopCreativeShockEffectPacket packet = new StopCreativeShockEffectPacket(serverPlayerEntity.interactionManager.getGameType());
+				MSPacketHandler.sendToPlayer(packet, serverPlayerEntity);
+			}
```
Dweblenod, line 323: Any idea how to make it send a packet it will accept?

src/main/java/com/mraof/minestuck/item/KeyItem.java
kirderf1: Can say that it should be stored in the item stack nbt, so you should probably write functions for the dimension and key type to read from/write to nbt. In 1.15, dimensions are referred to with an `DimensionType`.

src/main/java/com/mraof/minestuck/block/EnumKeyType.java
kirderf1: Yeah, I think so

src/main/java/com/mraof/minestuck/item/KeyItem.java
kirderf1: You would be more explicit in how the key type is turned into text. There are a few options:
- Override `toString()` in `EnumKeyType` and return the string you want
- Turn the type into a string before it is passed into the text component
- Turn the type into some form of text component (such as a translatable text component)

src/main/java/com/mraof/minestuck/event/ServerEventHandler.java
kirderf1: The packet is considered invalid because it is not registered in `MSPacketHandler`

src/main/java/com/mraof/minestuck/item/weapon/InventoryTickEffect.java
kirderf1: Since you're comparing an item stack and an item and not two itemstacks, you could compare the items directly (`MsItems.PRISMARINE_BASHER == otherStack.getItem()`)
Next, it looks like `isSelected` is unreliable for items not in the main inventory, so I'd suggest you check directly if it's in the right place (with `serverPlayerEntity.getHeldItemMainhand() == stack` and `serverPlayerEntity.getItemStackFromSlot(EquipmentSlotType.OFFHAND) == stack`)
For this specific purpose, its fine to use `==` instead of `stack.isItemEqual(stack2)`, since we're looking for the exact stack, and not just an equivalent stack.

src/main/java/com/mraof/minestuck/block/LootBlockTables.java
```diff
@@ -0,0 +1,60 @@
...
+import net.minecraft.item.ItemStack;
+import net.minecraft.item.Items;
+import net.minecraft.world.World;
+
+import java.lang.reflect.Array;
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
+
+public class LootBlockTables
```
Dweblenod, line 14: Not planning on using this as of the moment but it seems like a useful idea

src/main/java/com/mraof/minestuck/world/gen/feature/structure/tiered/tier1/TierOneDungeonSecondaryPiece.java
```diff
@@ -0,0 +1,638 @@
...
+		breathPuzzleDoorTemplate = templates.getOrCreate(new ResourceLocation(Minestuck.MOD_ID, "dungeons/breath_puzzle_door"));
+		breathFirstSideRoomTemplate = templates.getOrCreate(new ResourceLocation(Minestuck.MOD_ID, "dungeons/breath_first_side_room"));
+		breathSecondSideRoomBarrierTemplate = templates.getOrCreate(new ResourceLocation(Minestuck.MOD_ID, "dungeons/breath_second_side_room_barrier"));
+		breathAreaEffectChamberTemplate = templates.getOrCreate(new ResourceLocation(Minestuck.MOD_ID, "dungeons/breath_area_effect_chamber"));
+		bloodSymbolTemplate = templates.getOrCreate(new ResourceLocation(Minestuck.MOD_ID, "aspect/blood_symbol_no_background"));
+		bloodFirstSideRoomTemplate = templates.getOrCreate(new ResourceLocation(Minestuck.MOD_ID, "dungeons/blood_first_side_room"));
+		bloodSecondSideRoomTemplate = templates.getOrCreate(new ResourceLocation(Minestuck.MOD_ID, "dungeons/blood_second_side_room"));
+		bloodWallFountainTemplate = templates.getOrCreate(new ResourceLocation(Minestuck.MOD_ID, "dungeons/blood_wall_fountain")); //getOrCreate was getTemplateDefaulted
+	}
```
Dweblenod, line 120: Considering how many lines of code will be needed for the 12+ dungeon templates I think this init system should be changed

PR #307, opened 2021-08-15
Rung rebalancing

Dweblenod:
Changing the rung system to allow reaching of max rung. 

Now most imps stop giving xp around rung 8, ogres around 18, basilisks around 32, while liches and giclops never stop giving xp.


kirderf1 (approved):
Some classes that were otherwise untouched had some formatting changes. It's not a big deal here, but for future reference I'd suggest making sure you don't commit such changes if you don't do any other changes in the class. (in other words, in the future avoid committing classes where you only did formatting changes)

PR #308, opened 2021-08-15
Dialogues

amiwyn:
new dialogue gui


src/main/java/com/mraof/minestuck/client/gui/DialogueScreen.java
```diff
@@ -0,0 +1,202 @@
...
+public class DialogueScreen extends Screen
+{
+	private static final ResourceLocation BACKGROUND = new ResourceLocation("minestuck", "textures/gui/dialogue.png");
+	
+	private static final int GUI_WIDTH = 256;
+	private static final int GUI_HEIGHT = 106;
+	private static final int PORTRAIT_SIZE = 32;
+	
+	private DialogueCard[] dialogueCards;
+	private String[] responseOptions;
```
kirderf1, line 25: I'd suggest using lists instead of arrays for these throughout the code

src/main/java/com/mraof/minestuck/client/gui/DialogueScreen.java
```diff
@@ -0,0 +1,204 @@
...
+	{
+		this.renderBackground();
+		
+		int xOffset = (width - GUI_WIDTH) / 2;
+		int yOffset = (height - GUI_HEIGHT) / 2;
+		int leftStart = xOffset + PORTRAIT_SIZE + 16;
+		
+		RenderSystem.color4f(1.0F, 1.0F, 1.0F, 1.0F);
+		this.minecraft.getTextureManager().bindTexture(BACKGROUND);
+		this.blit(xOffset, yOffset, 0, 0, GUI_WIDTH, GUI_HEIGHT);
```
Dweblenod, line 54: The bottom section(black box with yellow/red outline) of the dialogue is cut off, is this intentional?

Dweblenod (changes requested):
This is a matter for sprite/model artists and obviously planned but by the end, a more well rounded set of dialogue sprites for consorts will be needed

src/main/java/com/mraof/minestuck/client/gui/DialogueScreen.java
Dweblenod: As well, it is important to note that the light gray background crashes with the lighter color texts of salamanders and iguanas. Something such as an alternate backdrop color should be done to provide more contrast

amiwyn: I wanted to keep consistency for the backgrounds so I tweaked the colors a bit but its not perfect. I can add it back if its something we want.

amiwyn: For the dialogue section, I would have to make it scalable and I removed it for simplicity to concentrate on the integration

PR #314, opened 2021-12-21
added mixins & optimised ServerEventHandler

LunyaticCat:


kirderf1:
Alright, gonna read up a little bit on mixins before properly reviewing this PR.

In the meantime, there are a few things I'd like to bring up already:
- I don't think this classifies as an optimization. Can you elaborate on your wording?
- What is the benefit of this method compared to how it was before?
- In the discord, I've suggested forge's capability system as an alternative. What are your thoughts on it?
- In `ServerEventHandler`, you're doing an unchecked cast to `ILivingEntity`. I'd suggest pairing it with an `instanceof` check.

LunyaticCat:
This is mostly a training for the cursor model replacing the regular player model I will do in the futur where mixins will be mendatory to work.
It still is an optimization though.
This method is injecting in minecraft living entity classes a getter for our xp system. This way we can call it on death without any need to check what entity it is as we where doing before.

LunyaticCat:
My code injection work for every LivingEntity classes.
```java
@Cancelable
public class LivingDeathEvent extends LivingEvent
{
    private final DamageSource source;
    public LivingDeathEvent(LivingEntity entity, DamageSource source)
    {
        super(entity);
        this.source = source;
    }

    public DamageSource getSource()
    {
        return source;
    }
}
```
Because LivingDeathEvent as a LivingEntity as entity, I don't think doing a instance of is relevant

LunyaticCat:
To learn about mixins, I suggest the github wiki of SpongePowered.
https://github.com/SpongePowered/Mixin/wiki

kirderf1:
Sure, but the code does not make it clear that it depends on the code injection to function properly. Also, if the code injection stops working due to software regression, you'd be left with a nondescript `ClassCastException`. What you could do instead, is to manually throw a more informative exception, such as `new IllegalStateException("Expected mixin to inject interface ILivingEntity into LivingEntity, but the entity was not an instance of ILivingEntity")`.

LunyaticCat:
I mean, ILivingEntity is a mixin interface in a folder called "mixin interface".
To me thinking that code injection would stop working due to software regression is the same as thinking SubscribeEvents will stop working due to that... It's very unlikely.
```java
	@SubscribeEvent(priority=EventPriority.LOWEST, receiveCanceled=false)
	public static void onEntityDeath(LivingDeathEvent event)
	{
		if(event.getEntity() instanceof IMob && event.getSource().getEntity() instanceof ServerPlayerEntity)
		{
			ServerPlayerEntity player = (ServerPlayerEntity) event.getSource().getEntity();
			int exp = 0;
			try {
					exp = ((ILivingEntity) event.getEntity()).getMinestuckXpValue();
			}
			catch(IllegalStateException ignored){
				throw new IllegalArgumentException("Expected mixin to inject interface ILivingEntity into LivingEntity, but the entity was not an instance of ILivingEntity");
			}

			if(exp > 0)
				Echeladder.increaseProgress(player, exp);
		}
		if(event.getEntity() instanceof ServerPlayerEntity)
		{
			TitleSelectionHook.cancelSelection((ServerPlayerEntity) event.getEntity());
		}
	}
```
I could change it to that though if you're afraid it might occur

kirderf1:
I also don't think its particularly likely to fail that way. Rather, the main point is code clarity, for which I don't think that "mixin interface" as a package name is sufficient here.
```java
try {
	exp = ((ILivingEntity) event.getEntity()).getMinestuckXpValue();
}
catch(ClassCastException ignored){
	throw new IllegalArgumentException("Expected mixin to inject interface ILivingEntity into LivingEntity, but the entity was not an instance of ILivingEntity");
}
```
is fine, but since the exception has a simple condition, I have a weak preference towards
```java
if (event.getEntity() instanceof ILivingEntity)
	exp = ((ILivingEntity) event.getEntity()).getMinestuckXpValue();
else
	throw new IllegalArgumentException("Expected mixin to inject interface ILivingEntity into LivingEntity, but the entity was not an instance of ILivingEntity");
```
or once we update to java 16

```java
if (event.getEntity() instanceof ILivingEntity entity)
	exp = entity.getMinestuckXpValue();
else
	throw new IllegalArgumentException("Expected mixin to inject interface ILivingEntity into LivingEntity, but the entity was not an instance of ILivingEntity");
```.

LunyaticCat:
I changed it to the way you suggested

kirderf1 (changes requested):
Alright, first I want to say that I'm sorry that I couldn't get back sooner.
Now, I still think that the PR has a fundamental problem in answering the question "why should we have this?".
From your replies, I seem to scrape three reasons:
- To optimize the code (for performance?)
- Modularity, to be able to add echeladder xp definitions independently (for addon support?)
- Moving the "entity -> xp" definitions out from ServerEventHandler
- In preparation for other contributions using mixin

My current stance on these are as follows:
- I can't make an accurate comparison to determine if this would perform better or worse. But I do believe that regardless of leaning, it would be close enough that calling this an "optimization" is misleading.
- If the goal is to let addons introduce echeladder xp for other mods, I think that capabilities are a better choice, primarily because it would handle conflicts in a safer way. Capabilities essentially involve attaching mod interfaces to things, which seems well suited for this use case. I also think that capabilities hold merit as an alternative that does not involve any form of code injection on our part.
- This specific point has little to do with mixins, and essentially boils down to hiding the actual definitions behind some sort of function, be it some static function or a function on an interface.
- Since adding a mixin dependency is a relatively small change, it could easily be added to a future PR where it is needed if this PR falls through. As such, it's not really something we need to add now unless your mixin stays.
If I have misinterpreted your reasons or haven't considered something crucial, feel free to add it.
Overall, I currently think it's appropriate to close this PR and instead create a solution with capabilities.

LunyaticCat:
Because  we are directly adding minestuck xp into the living entity class, any mod adding an entity using this class would also return minestuck xp. It goes beyond simple addons as we would have automatic comptatibilities with other mods mobs.

LunyaticCat:
that's why I would also suggest that by default, the minestuck xp reward should depend on how much minecraft xp the mob is rewarding.
this way, we could have different xp rewards for stronger mobs without knowing what the mob could be.

kirderf1:
You seem to be talking about having a default method of defining minestuck xp. To me, that doesn't seem relevant here, as that could easily be done for any method that has been mentioned (mixins, capabilities, and even the original method with instanceof checks)

kirderf1:
As said, I believe that we can easily have a default method for minestuck xp no matter which of the three alternatives we use.
With the original method, it'd be called if no `instanceof` check succeeds.
With your entity interface, the default method could be called for entities without the interface, or as you've chosen to do, the default method is mixin'ed into `LivingEntity`.
With capabilities, the default method could be used if the capability is not present on the entity. If attached capabilities can be attached in an orderly fashion, it would also be possible to attach the default behavior as a capability to any `LivingEntity`.

PR #319, opened 2022-03-19
Add smithing recipe grist support

jwright159:
frick reflection

### DIFF
--- a/src/main/java/com/mraof/minestuck/item/crafting/alchemy/generator/DefaultInterpreter.java
+++ b/src/main/java/com/mraof/minestuck/item/crafting/alchemy/generator/DefaultInterpreter.java
@@ -7,8 +7,12 @@ import net.minecraft.item.Item;
 import net.minecraft.item.ItemStack;
 import net.minecraft.item.crafting.IRecipe;
 import net.minecraft.item.crafting.Ingredient;
+import net.minecraft.item.crafting.SmithingRecipe;
+import net.minecraft.util.NonNullList;
 import net.minecraftforge.registries.ObjectHolder;
+import org.apache.logging.log4j.LogManager;
 
+import java.lang.reflect.Field;
 import java.util.Collections;
 import java.util.List;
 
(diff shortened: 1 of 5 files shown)


src/main/java/com/mraof/minestuck/item/crafting/alchemy/generator/DefaultInterpreter.java
```diff
@@ -33,9 +37,37 @@ public GristSet generateCost(IRecipe<?> recipe, Item output, GenerationContext c
...
-		
+
 		GristSet totalCost = new GristSet();
-		for(Ingredient ingredient : recipe.getIngredients())
+
+		// Frick reflection frick reflection frick reflection frick reflection i hate java
+		// this was a whole minecraft oversight anyway
+		// so if they eventually fix this, remove this bit ok?
+		NonNullList<Ingredient> ingredients;
+		if (recipe instanceof SmithingRecipe)
```
kirderf1, line 47: Can you write this as a new recipe interpreter for smithing recipes instead of extending the default interpreter?

src/main/java/com/mraof/minestuck/item/crafting/alchemy/generator/DefaultInterpreter.java
```diff
@@ -33,9 +37,37 @@ public GristSet generateCost(IRecipe<?> recipe, Item output, GenerationContext c
...
+		{
+			try
+			{
+				Field baseField = SmithingRecipe.class.getDeclaredField("base");
```
kirderf1, line 51: Because of how minecraft mappings are handled, this won't work in production as the fields won't be called "base" and "addition". You should use `ObfuscationReflectionHelper` instead. (That class will also call `setAccessible()` for you)
You'll need the srg names for that which for "base" is "field_234837_a_" and for "addition" is "field_234838_b_".

kirderf1 (changes requested):

src/main/java/com/mraof/minestuck/item/crafting/alchemy/generator/DefaultInterpreter.java
```diff
@@ -7,8 +7,12 @@
 import net.minecraft.item.ItemStack;
 import net.minecraft.item.crafting.IRecipe;
 import net.minecraft.item.crafting.Ingredient;
+import net.minecraft.item.crafting.SmithingRecipe;
+import net.minecraft.util.NonNullList;
 import net.minecraftforge.registries.ObjectHolder;
+import org.apache.logging.log4j.LogManager;
 
+import java.lang.reflect.Field;
```
kirderf1, line 15: There are now unused imports from before. Revert the changes to the file/remove the imports.

src/main/java/com/mraof/minestuck/item/crafting/alchemy/generator/SmithingInterpreter.java
```diff
@@ -0,0 +1,75 @@
...
+
+	@Override
+	public GristSet generateCost(IRecipe<?> recipe, Item output, GenerationContext context)
+	{
+		// SmithingRecipes don't list their ingredients as ingredients so use this as workaround
+		try
+		{
+			GristSet totalCost = new GristSet();
+
+			Ingredient base = (Ingredient)ObfuscationReflectionHelper.findField(SmithingRecipe.class, "field_234837_a_").get(recipe);
```
kirderf1, line 33: I'd suggest either caching the field object in a static field, or change to using `ObfuscationReflectionHelper.getPrivateValue()`. The latter is easier to write, but the former is more appropriate under these circumstances.

kirderf1 (approved):

PR #320, opened 2022-03-20
Add /classes to .gitignore

jwright159:
I have about 3000 unversioned files reported by git in the `/classes` directory that are generated by the build process. They really don't need to be able to be included afaik.


jwright159:
Oh. No, that's a *me* issue, probably left over from working on 1.12 or something.

PR #321, opened 2022-03-27
added Edison's Fury

Badadamadaba:


src/main/java/com/mraof/minestuck/data/MinestuckEnUsLanguageProvider.java
```diff
@@ -706,6 +706,8 @@ protected void addTranslations()
 		addTooltip(MSItems.GOLDEN_SPORK, "A spoon and fork with all the awfulness of gold tools? Sign me up!");
 		add(MSItems.BIDENT, "Bident");
 		addTooltip(MSItems.BIDENT, "Double trouble for all your bubble scuffle needs");
+		add(MSItems.EDISONS_FURY, "Edison's Fury");
+		add(MSItems.EDISONS_SERENITY, "Edison's Fury");
```
Dweblenod, line 710: This is dependent on preference, but it would be smart to match the tooltip name to the registry name if you want the serenity/fury naming scheme to be identifiable by players and be understood in the code

Dweblenod (commented):
This is good! There are no errors or components to clean up with this request. Just some points of personal opinion regarding naming schemes and the need for grist cost/alchemy registry as discussed in Discord

src/main/java/com/mraof/minestuck/item/MSItems.java
```diff
@@ -1009,6 +1011,8 @@ public static void registerItems(final RegistryEvent.Register<Item> event)
...
 		registry.register(new WeaponItem(new WeaponItem.Builder(ItemTier.GOLD, 5, -2.5F).efficiency(1.0F).set(MSItemTypes.SHOVEL_TOOL), new Item.Properties().tab(MSItemGroup.WEAPONS)).setRegistryName("golden_spork"));
 		registry.register(new WeaponItem(new WeaponItem.Builder(ItemTier.IRON, 7, -2.9F).efficiency(1.0F).set(MSItemTypes.MISC_TOOL), new Item.Properties().tab(MSItemGroup.WEAPONS)).setRegistryName("bident"));
+		registry.register(new WeaponItem(new WeaponItem.Builder(ItemTier.IRON, 3, -2.4F).efficiency(1.0F).set(MSItemTypes.MISC_TOOL).set(ItemRightClickEffect.switchTo(() -> EDISONS_SERENITY)).add(OnHitEffect.DROP_FOE_ITEM).add(InventoryTickEffect.DROP_WHEN_IN_WATER).add(OnHitEffect.playSound(() -> MSSoundEvents.EVENT_ELECTRIC_SHOCK, 0.6F, 1.0F)), new Item.Properties().tab(MSItemGroup.WEAPONS)).setRegistryName("edisons_fury"));
+		registry.register(new WeaponItem(new WeaponItem.Builder(ItemTier.IRON, 3, -2.4F).efficiency(1.0F).set(MSItemTypes.MISC_TOOL).set(ItemRightClickEffect.switchTo(() -> EDISONS_FURY)), new Item.Properties().tab(MSItemGroup.WEAPONS)).setRegistryName("edisons_serenity"));
```
Dweblenod, line 1015: The weapon speed should actually be -2.6F instead of -2.4F if you want to keep it consistent with other standard forks

kirderf1:
I suspect that dweblenod has already mentioned it to you, but just in case:
Never modify or add files in "main/generated" manually. That folder should only have files generated from the "runData" configuration. If you want to add resources manually, add them in "main/resources" instead of "main/generated/resources". However in your case I'd recommend adding the recipes through code in the appropriate data providers (that you can find in the data package).

PR #324, opened 2022-04-09
Blocksets

HellZDarkArkon:
I was asked by Dweb to implement these blocks.
I've got most of the blocksets ready by now.


src/main/resources/assets/minestuck/models/block/copy_block.json
```diff
@@ -0,0 +1,7 @@
+{
+  "parent": "block/cube_all",
+  "textures":
+  {
+    "all": "minestuck:block/minestuck_missingtex"
+  }
+}
```
Dweblenod, line 7: this commit still has the template files

Dweblenod (commented):
Excellent work overall! Some stuff to change but this contributions are valuable. Make sure to include art credits for all the textures

Dweblenod:
Created new pull request with specifically everything needed so this can be wrapped up

PR #325, opened 2022-04-12
New blocks - HellZ

Dweblenod:
Finishing up some work that HellZ did the majority of, submitting it in this pull request so that it would get in for 4/13.

Multiple artists collaborated to create new textures or retexture existing blocks(with some gaps) in the categories of:
-Flowery Mossy Stone
-Frost
-Shade
-Mycelium
-Black stone
-Chalk
-Coarse stone
-Pink stone(added column but made no retextures)

Includes new slabs and stair variants for those sets as well as brown stone/green stone/cast iron(just one stair, no slab)


src/main/resources/assets/minestuck/models/block/flowery_mossy_stone_bricks.json
```diff
@@ -1,6 +0,0 @@
-{
-    "parent": "block/cube_all",
-    "textures": {
-        "all": "minestuck:block/flowery_mossy_stone_bricks"
-    }
-}
```
kirderf1, line 6: When replacing this with the numbered models, you forgot to update the model used for the double-slab block in the `flowery_mossy_stone_brick_slab` blockstate file (which still refers to the `flowery_mossy_stone_bricks` model)
Since you changed the slab models to use the same textures as the `flowery_mossy_stone_bricks1`model, it makes sense to fix this by changing the model in the blockstate to `flowery_mossy_stone_bricks1`.

kirderf1 (commented):
I am not seeing any new stonecutting recipes, which I thought would be included with these new blocks. Also, are all the new blocks getting grist costs from the grist cost generator?

There seems to be some old previously unused assets that you accidentally started using without updating. You should have a look at the blockstates file for `pink_stone_slab` and the model `pink_stone_double_slab` and update them accordingly.

kirderf1 (approved):

kirderf1:
If you could compile a list of changes in the pr description, that'd help later with the changelog.

PR #330, opened 2022-04-27
Puzzle block extensions

Dweblenod:
Adding additional puzzle blocks and making improvements to the ones that are already in the mod

New blocks:
-Toggler: toggles a specific blockstate that many puzzle blocks have been modified to possess or been given(if they are new), can be toggled itself to instead shut off powered redstone blocks, has an alchemy recipe of piston || lever
-Remote Comparator: checks for blocks at a specific distance and gives off redstone if it matches the type that is directly behind it, can be toggled to additionally matches blockstates, has an alchemy recipe of block pressure plate || plutonium core
-Structure Core: records data to specific structures and can permanently shut down specific blocks functionality(summoners and area effect blocks) if otherwise set to read data from said structures, intended for actual use in the Tiered Dungeons update
-Pushable Block: can be pushed in a horizontal direction if right clicked, imagined with sokoban/zelda/portal style puzzles in mind, can absorb platform blocks but cannot be pushed into their way, destroys small plants and any blocks listed in the block tag "pushable_block_replacable", has an alchemy recipe of stone(the block tag) || gravel, has a stonecutting recipe from stone bricks
-Block Pressure Plate: will give off redstone if a player(who is not crouching) or a block with a full faced bottom side are directly above it, has an alchemy recipe of observer || stone pressure plate

New functionality/changes: 
-Area effect block: can be shut down by the structure core when in READ_AND_WIPE mode. The block now works as an offset, allowing it to change the area of effect upon rotation. GUI now also has a way to set the effect type/amplifier and change whether all mobs or just players are affected
-Item Magnet: can now be machine toggled to push entities away instead of pulling them closer. Can also move the new pushable blocks upwards(turning them into a falling block entity)
-Platform block/platform generator: now use new blockstate properties to reduce the amount of block updates they caused, with the platform blocks being able to check whether they are being supported and the generators able to more reliable create and remove the platform blocks, including a block tag for blocks that should absorb the platforms. Now has the ability to replace blocks like grass/fire/liquid
-Summoner: now has a gui for improved capacity to set the summoned entity among other small utility
-Remote Observer: the area at which it can observe is now variable. Now has a blacklist entity type tag for entities it shouldnt be able to detect. IS_ENTITY_IN_WATER has been split into a type for if an entity is submerged in water and a type for it the player is wet(in bubble column/water/rain), and now sprinting + entities on ground can be checked for
-Stat storer: alchemy recipe now uses computer parts instead of comparator. If the Toggler block is in Discharge mode it will also reset the active stat type value to 0, increasing its capacity for automation
-Variable Solid Switch: now has an alchemy recipe of Solid Switch || Comparator
-Redstone Clock: now has an alchemy recipe of Clock && Comparator. Fixed issue with the clock tick speed resetting every time it was unloaded
-Fragile Stone: alchemy recipe now uses && instead of ||. has a stonecutting recipe from stone
-Retractable Spikes: now has an alchemy recipe of Spikes && Sticky Piston
-Wireless Redstone Transmitter: now works as an offset, allowing it to power different receivers depending on rotation. now uses computer parts instead of quartz in its crafting recipe
-Rotator: expanded the list of blocks it can rotate(now includes repeaters and comparators for advanced redstone). now has a crafting recipe


src/main/java/com/mraof/minestuck/tileentity/redstone/SummonerTileEntity.java
```diff
@@ -97,7 +98,7 @@ public void setSummonedEntity(EntityType<?> entityTypeIn, @Nullable PlayerEntity
 	{
 		this.summonType = entityTypeIn;
 		
-		if(playerEntityIn != null)
+		if(playerEntityIn != null) //used when setting via spawn egg, does not play message when set through gui
 			playerEntityIn.displayClientMessage(new TranslationTextComponent(SUMMON_TYPE_CHANGE, summonType.getRegistryName()), true);
```
kirderf1, line 102: How about moving this function call to `SummonerBlock.use()`? That'd remove the need for the player parameter and the null check.

src/main/java/com/mraof/minestuck/tileentity/redstone/SummonerTileEntity.java
```diff
@@ -26,6 +26,7 @@ public class SummonerTileEntity extends TileEntity implements ITickableTileEntit
 {
 	private EntityType<?> summonType;
 	private int cooldownTimer;
+	private int summonRange; //default is 8, but can be set(via gui) between 1 and 64
```
kirderf1, line 29: If the default is 8, why is it not initialized to 8 here?

src/main/java/com/mraof/minestuck/client/gui/SummonerScreen.java
```diff
@@ -0,0 +1,114 @@
...
+	
+	private TextFieldWidget entityTypeTextField;
+	
+	
+	SummonerScreen(SummonerTileEntity te)
+	{
+		super(new StringTextComponent("Summoner"));
+		
+		this.te = te;
+		this.summonRange = te.getSummonRange() > 0 ? te.getSummonRange() : 8; //if its defaulted on creation to 0, set it to the intended default of 8
```
kirderf1, line 44: I think that it's better to handle the default summon range in the tile entity.

src/main/java/com/mraof/minestuck/client/gui/SummonerScreen.java
```diff
@@ -0,0 +1,114 @@
...
+	public Button incrementButton;
+	public Button decrementButton;
+	private Button unTriggerableButton;
+	
+	private TextFieldWidget entityTypeTextField;
+	
+	
+	SummonerScreen(SummonerTileEntity te)
+	{
+		super(new StringTextComponent("Summoner"));
```
kirderf1, line 41: Translation key candidate?

src/main/java/com/mraof/minestuck/client/gui/SummonerScreen.java
```diff
@@ -0,0 +1,114 @@
...
+	
+	private static final int guiWidth = 150;
+	private static final int guiHeight = 98;
+	
+	private final SummonerTileEntity te;
+	private boolean isUntriggerable;
+	private int summonRange;
+	
+	public Button incrementButton;
+	public Button decrementButton;
```
kirderf1, line 33: Why are these buttons public? I'd expect them to be private.

src/main/java/com/mraof/minestuck/client/gui/SummonerScreen.java
```diff
@@ -0,0 +1,114 @@
...
+			addButton(unTriggerableButton = new ExtendedButton(this.width / 2 - 65, yOffset + 70, 85, 20, new StringTextComponent("TRIGGERABLE"), button -> cycleUntriggerable()));
+		addButton(new ExtendedButton(this.width / 2 + 25, yOffset + 70, 40, 20, new StringTextComponent("DONE"), button -> finish()));
+	}
+	
+	/**
+	 * Attempts to increase or decrease the range at which the entity can be summoned if that value is between the min and max
+	 */
+	private void changeRange(int change)
+	{
+		summonRange = MathHelper.clamp(summonRange + change, 1, 64);
```
kirderf1, line 72: How about also modifying `this.incrementButton.active` and `this.decrementButton.active` depending on if the summon range is at its maximum or minimum?

src/main/java/com/mraof/minestuck/client/gui/SummonerScreen.java
```diff
@@ -0,0 +1,114 @@
...
+		this.isUntriggerable = te.getBlockState().getValue(SummonerBlock.UNTRIGGERABLE);
+	}
+	
+	@Override
+	public void init()
+	{
+		int yOffset = (this.height / 2) - (guiHeight / 2);
+		
+		addButton(incrementButton = new ExtendedButton(this.width / 2 + 20, (height - guiHeight) / 2 + 12, 20, 20, new StringTextComponent("+"), button -> changeRange(1)));
+		addButton(decrementButton = new ExtendedButton(this.width / 2 - 40, (height - guiHeight) / 2 + 12, 20, 20, new StringTextComponent("-"), button -> changeRange(-1)));
```
kirderf1, line 54: Since the summon range is in 1-64, it may be nicer to be able to reach a range farther away quicker. I've got two different suggestions for it:
1. Replace the buttons with a text field that the user writes the desired range in
2. Give some means of changing the range by larger amounts with the buttons. I think the most intuitive alternative is to increase the amount if the buttons were clicked with shift held down. Other alternatives include increasing the amount by holding down ctrl, or by using a different mouse button. With the range being 1-64, I think it makes sense for the increased amount to be 8. (So in summary, change the range by 8 when shift-clicking the buttons, change by 1 otherwise)

src/main/java/com/mraof/minestuck/client/gui/SummonerScreen.java
```diff
@@ -0,0 +1,114 @@
...
+		
+		this.entityTypeTextField = new TextFieldWidget(this.font, this.width / 2 - 60, yOffset + 40, 120, 18, new StringTextComponent("Current Entity Type"));	//TODO Use translation instead, and maybe look at other text fields for what the text should be
+		this.entityTypeTextField.setValue(EntityType.getKey(te.getSummonedEntity()).toString());
+		addButton(entityTypeTextField);
+		
+		if(isUntriggerable)
+			addButton(unTriggerableButton = new ExtendedButton(this.width / 2 - 65, yOffset + 70, 85, 20, new StringTextComponent("UNTRIGGERABLE"), button -> cycleUntriggerable()));
+		else
+			addButton(unTriggerableButton = new ExtendedButton(this.width / 2 - 65, yOffset + 70, 85, 20, new StringTextComponent("TRIGGERABLE"), button -> cycleUntriggerable()));
```
kirderf1, line 63: I'd suggest simplifying this (and `cycleUntriggerable()`) with a function:
```java
private ITextComponent getTriggerableButtonMessage()
{
	return this.isUntriggerable ? new StringTextComponent("UNTRIGGERABLE") : new StringTextComponent("TRIGGERABLE");
}
```
These button messages could also be candidates for getting translation keys.

src/main/java/com/mraof/minestuck/client/gui/SummonerScreen.java
```diff
@@ -0,0 +1,114 @@
...
+			unTriggerableButton.setMessage(new StringTextComponent("TRIGGERABLE"));
+	}
+	
+	/**
+	 * Returns the current entity type, with imps as the generic entity type
+	 */
+	private EntityType<?> getEntityType(String stringInput)
+	{
+		Optional<EntityType<?>> attemptedEntityType = EntityType.byString(stringInput);
+		return attemptedEntityType.orElse(MSEntityTypes.IMP);
```
kirderf1, line 93: It may make sense to change the gui so that it rejects nonexisting entity types (draws a message on the gui and doesn't allow the user to press the "DONE" button until a valid type is written), instead of silently changing it to an imp.

src/main/java/com/mraof/minestuck/client/gui/SummonerScreen.java
```diff
@@ -0,0 +1,114 @@
...
+import net.minecraftforge.fml.client.gui.widget.ExtendedButton;
+
+import java.util.Optional;
+
+public class SummonerScreen extends Screen
+{
+	private static final ResourceLocation guiBackground = new ResourceLocation("minestuck", "textures/gui/generic_medium.png");
+	
+	private static final int guiWidth = 150;
+	private static final int guiHeight = 98;
```
kirderf1, line 26: Static final constants in java typically use SCREAMING_SNAKE_CASE.

src/main/java/com/mraof/minestuck/client/gui/SummonerScreen.java
```diff
@@ -0,0 +1,114 @@
...
+	@Override
+	public void render(MatrixStack matrixStack, int mouseX, int mouseY, float partialTicks)
+	{
+		this.renderBackground(matrixStack);
+		RenderSystem.color4f(1F, 1F, 1F, 1F);
+		this.minecraft.getTextureManager().bind(guiBackground);
+		int yOffset = (this.height / 2) - (guiHeight / 2);
+		
+		this.blit(matrixStack, (this.width / 2) - (guiWidth / 2), yOffset, 0, 0, guiWidth, guiHeight);
+		font.draw(matrixStack, Integer.toString(summonRange), (width / 2) - 5, (height - guiHeight) / 2 + 16, 0x404040);
```
kirderf1, line 105: You have already defined yOffset. Use it instead of `(height - guiHeight) / 2` (and consider if you want to write `yOffset = (height - guiHeight) / 2` instead of how yOffset is written right now)

src/main/java/com/mraof/minestuck/tileentity/redstone/SummonerTileEntity.java
```diff
@@ -108,12 +109,23 @@ public EntityType<?> getSummonedEntity()
 		return this.summonType;
 	}
 	
+	public void setSummonRange(int rangeIn)
+	{
+		this.summonRange = rangeIn;
```
kirderf1, line 114: Consider if you want to enforce a range here with `MathHelper.clamp()` for safety.

kirderf1 (changes requested):
Reviewed just the Summoner block. More to come. Make sure that this doesn't become another "catch-all" pull request like the puzzle blocks PR before it.

src/main/java/com/mraof/minestuck/client/gui/SummonerScreen.java
Dweblenod: I think I understand but I am not sure what the benefit is or exactly how it would work

src/main/java/com/mraof/minestuck/client/gui/SummonerScreen.java
Dweblenod: For suggestion 2, would it be something like using the function keyPressed() while in the changeRange function still?

src/main/java/com/mraof/minestuck/client/gui/SummonerScreen.java
Dweblenod: Oh wait, you mean visually removing the buttons once it is no longer possible to increment in that direction dont you

kirderf1: By setting `button.active` to false, the button will darken and become unclickable. It's a good indicator for the player when the range can't go further in one direction. You could also change `button.visible` to false to visually remove it, but I think that `button.active` is more appropriate here.

src/main/java/com/mraof/minestuck/tileentity/redstone/SummonerTileEntity.java
```diff
@@ -108,12 +100,23 @@ public EntityType<?> getSummonedEntity()
...
+		return MathHelper.clamp(this.summonRange, 1, 64);
+	}
+	
 	@Override
 	public void load(BlockState state, CompoundNBT compound)
 	{
 		super.load(state, compound);
 		
 		cooldownTimer = compound.getInt("cooldownTimer");
+		summonRange = compound.getInt("summonRange");
```
kirderf1, line 119: Because "summonRange" didn't exist before, I suggest adding a `if(compound.contains("summonRange", Constants.NBT.TAG_ANY_NUMERIC))` check, so that the range is only read if it is in nbt. (`compound.getInt()` will default to 0 if the range tag is not there, which is not the default you want)

src/main/java/com/mraof/minestuck/tileentity/redstone/WirelessRedstoneTransmitterTileEntity.java
```diff
@@ -19,8 +21,10 @@
 
 public class WirelessRedstoneTransmitterTileEntity extends TileEntity implements INameable, ITickableTileEntity
 {
-	private BlockPos destBlockPos;
+	private BlockPos offsetPos;
+	private Direction facing;
 	private int tickCycle;
+	private static final int WIRELESS_CONSTANT = 6;
```
kirderf1, line 27: `WIRELESS_CONSTANT` isn't a very informative name in this context. What is it for?

src/main/java/com/mraof/minestuck/tileentity/redstone/WirelessRedstoneTransmitterTileEntity.java
```diff
@@ -127,10 +150,10 @@ public void load(BlockState state, CompoundNBT compound)
...
 		
 		tickCycle = compound.getInt("tickCycle");
-		int destX = compound.getInt("destX");
-		int destY = compound.getInt("destY");
-		int destZ = compound.getInt("destZ");
-		this.destBlockPos = new BlockPos(destX, destY, destZ);
+		int offsetX = compound.getInt("offsetX");
+		int offsetY = compound.getInt("offsetY");
+		int offsetZ = compound.getInt("offsetZ");
+		this.offsetPos = new BlockPos(offsetX, offsetY, offsetZ);
```
kirderf1, line 156: Can you include backwards-portability here? It'd involve detecting if you're loading the new format with "offset" coords, or the old format with "dest" coords, and in the case of "dest" coords convert that position to an offset position.

src/main/java/com/mraof/minestuck/tileentity/redstone/RemoteObserverTileEntity.java
```diff
@@ -23,6 +23,8 @@ public class RemoteObserverTileEntity extends TileEntity implements ITickableTil
 	private int tickCycle;
 	@Nonnull
 	private ActiveType activeType;
+	private int observingRange; //default is 16, but can be set(via gui) between 1 and 64
```
kirderf1, line 26: The default value should be handled by the tile entity.

src/main/java/com/mraof/minestuck/tileentity/redstone/AreaEffectTileEntity.java
```diff
@@ -143,15 +202,15 @@ public void load(BlockState state, CompoundNBT compound)
...
+		this.minAreaOffset = new BlockPos(minAreaOffsetX, minAreaOffsetY, minAreaOffsetZ);
 		
-		int maxEffectPosX = compound.getInt("maxEffectPosX");
-		int maxEffectPosY = compound.getInt("maxEffectPosY");
-		int maxEffectPosZ = compound.getInt("maxEffectPosZ");
-		this.maxEffectPos = new BlockPos(maxEffectPosX, maxEffectPosY, maxEffectPosZ);
+		int maxAreaOffsetX = compound.getInt("maxAreaOffsetX");
+		int maxAreaOffsetY = compound.getInt("maxAreaOffsetY");
+		int maxAreaOffsetZ = compound.getInt("maxAreaOffsetZ");
+		this.maxAreaOffset = new BlockPos(maxAreaOffsetX, maxAreaOffsetY, maxAreaOffsetZ);
```
kirderf1, line 213: Can you make this backwards-compatible? If the nbt tag has an effect pos, it would be loaded and converted to an area offset.

src/main/java/com/mraof/minestuck/data/loot_table/MSBlockLootTables.java
```diff
@@ -380,6 +380,37 @@ protected void addTables()
...
+		dropSelf(FALL_PAD);
+		dropSelf(FRAGILE_STONE);
+		dropSelf(RETRACTABLE_SPIKES);
+		dropSelf(PORTABLE_BLOCK);
+		dropSelf(AND_GATE_BLOCK);
+		dropSelf(OR_GATE_BLOCK);
+		dropSelf(XOR_GATE_BLOCK);
+		dropSelf(NAND_GATE_BLOCK);
+		dropSelf(NOR_GATE_BLOCK);
+		dropSelf(XNOR_GATE_BLOCK);
```
kirderf1, line 412: Looks like you accidentally duplicated the group of drops above.

src/main/java/com/mraof/minestuck/client/gui/StatStorerScreen.java
```diff
@@ -8,25 +8,23 @@
...
-	private static final ResourceLocation guiBackground = new ResourceLocation("minestuck", "textures/gui/generic_medium.png");
+	private static final ResourceLocation GUI_BACKGROUND = new ResourceLocation("minestuck", "textures/gui/generic_medium.png");
 	
-	private static final int guiWidth = 150;
-	private static final int guiHeight = 98;
+	private static final int GUI_WIDTH = 150;
+	private static final int GUI_HEIGHT = 98;
 	
-	private static final String divideValueMessage = "Divide power output by:";
+	private static final String DIVIDE_VALUE_MESSAGE = "Divide power output by:";
```
kirderf1, line 22: Translation key candidate

kirderf1 (changes requested):

src/main/java/com/mraof/minestuck/tileentity/redstone/SummonerTileEntity.java
Dweblenod: Since it would also be optimal not to allow even players with commands to set the block to 0, perhaps I could create a check in getSummonRange() to convert any 0 values to the default? If not, I could make sure that getSummonRange is used anyways as it has a clamp. This would cover both issues with both older mod versions and attempts to set "illegal" bounds

src/main/java/com/mraof/minestuck/client/gui/SummonerScreen.java
Dweblenod: I will try to keep this approach in mind for the future, but opted to instead add two buttons that increment or decrement by 10 with the belief that it would be easier to implement and easier for player comprehension

src/main/java/com/mraof/minestuck/client/gui/SummonerScreen.java
Dweblenod: Did this, but instead of a message I turned the text in the text field red and prevented it from sending the packet/exiting

src/main/java/com/mraof/minestuck/tileentity/redstone/WirelessRedstoneTransmitterTileEntity.java
Dweblenod: Do you consider my added comment to have resolved this?

src/main/java/com/mraof/minestuck/tileentity/redstone/SummonerTileEntity.java
kirderf1: Putting it in `getSummonRange()` does work, but as an alternative I'd like to suggest putting it in `setSummonRange()`, and using it when reading from nbt. `this.setSummonRange(compound.getInt("summonRange"));`
If you do want to keep the clamp in the getter and not move it to the setter, you must make sure that you always use the getter to get the range, and never access `summonRange` directly like you currently do in `summonEntity()`.
Either way, the range would then instead end up at 1 when loaded for an old version. It works, but it could be better. I'd still suggest that you have a check for in the load function. In addition, to match the range before this PR, the default here could be 16 instead of 8.
```java
if(compound.contains("summonRange", Constants.NBT.TAG_ANY_NUMERIC))
	this.setSummonRange(compound.getInt("summonRange"));
else this.setSummonRange(16);
```

src/main/java/com/mraof/minestuck/tileentity/redstone/WirelessRedstoneTransmitterTileEntity.java
kirderf1: Alright, with the new comment, I'd suggest renaming the constant to something like `TICK_SPEED`, `TICK_DURATION`, `TICK_FREQUENCY` or `TICK_DELAY`

src/main/java/com/mraof/minestuck/tileentity/redstone/WirelessRedstoneTransmitterTileEntity.java
```diff
@@ -42,20 +46,42 @@ public void tick()
...
+		if(offsetPos == null)
+			offsetPos = new BlockPos(0, 0, 0);
+		if(facing == null)
+			facing = getBlockState().getValue(WirelessRedstoneTransmitterBlock.FACING);
+		
+		Direction stateFacing = getBlockState().getValue(WirelessRedstoneTransmitterBlock.FACING);
+		
+		//in cases where the block has been rotated after the coordinates were set(rotator blocks or placement in structure templates), this will transform the relative coordinates by rotating them around the block
+		if(facing != stateFacing)
+			this.offsetPos = offsetPos.rotate(MSRotationUtil.rotationBetween(this.facing, stateFacing));
```
kirderf1, line 60: I'd suggest removing the need for a `facing` field by doing one of:
- Let the offset position always be the unrotated offset, and in this function get a rotation from `stateFacing` to them apply with `this.getBlockPos().offset(offsetPos.rotate(rotation))`
- Introduce functionality so that the rotator block can call some function to rotate `offsetPos`

src/main/java/com/mraof/minestuck/tileentity/redstone/WirelessRedstoneTransmitterTileEntity.java
```diff
@@ -127,10 +153,18 @@ public void load(BlockState state, CompoundNBT compound)
...
 		
 		tickCycle = compound.getInt("tickCycle");
-		int destX = compound.getInt("destX");
-		int destY = compound.getInt("destY");
-		int destZ = compound.getInt("destZ");
-		this.destBlockPos = new BlockPos(destX, destY, destZ);
+		int offsetX = compound.getInt("offsetX");
+		int offsetY = compound.getInt("offsetY");
+		int offsetZ = compound.getInt("offsetZ");
+		this.offsetPos = new BlockPos(offsetX, offsetY, offsetZ);
```
kirderf1, line 159: Could you move offset loading into an if-else statement with the dest loading?
It makes it clearer that dest coords is an alternative format to offset coords if you write it like this:
```java
if (should load offset coords)
{
	load offset;
} else
{
	load dest;
}
```
or
```java
if (should load dest coords)
{
	load dest;
} else
{
	load offset;
}
```

src/main/java/com/mraof/minestuck/tileentity/redstone/WirelessRedstoneTransmitterTileEntity.java
```diff
@@ -42,20 +46,42 @@ public void tick()
...
+		return getBlockPos().offset(offsetPos);
 	}
 	
-	public void setDestinationBlockPos(BlockPos destinationPosIn)
+	public void setOffsetFromDestinationBlockPos(BlockPos destinationPosIn)
 	{
-		this.destBlockPos = destinationPosIn;
+		if(level != null) //level is null when gotten from Load
+		{
+			this.facing = level.getBlockState(worldPosition).getValue(WirelessRedstoneTransmitterBlock.FACING);
```
kirderf1, line 72: Get the block state as an argument from the `load()` function instead.

src/main/java/com/mraof/minestuck/network/RemoteObserverPacket.java
```diff
@@ -59,6 +63,7 @@ public void execute(ServerPlayerEntity player)
 					((RemoteObserverTileEntity) te).setActiveType(activeType);
 					if(entityType != null)
 						((RemoteObserverTileEntity) te).setCurrentEntityType(entityType);
+					((RemoteObserverTileEntity) te).setObservingRange(observingRange);
```
kirderf1, line 66: At this point, I'd suggest using a `RemoteObserverTileEntity observer = (RemoteObserverTileEntity) te;`

src/main/java/com/mraof/minestuck/tileentity/redstone/RemoteObserverTileEntity.java
```diff
@@ -123,9 +127,22 @@ public EntityType<?> getCurrentEntityType()
...
 	}
 	
-	public void setCurrentEntityType(EntityType<?> currentEntityType)
+	/**
+	 * Checks for entity types that are not intended to be known by survival players or are forbidden from typical use
+	 */
+	public static boolean entityCanBeObserved(EntityType<?> currentEntityType)
 	{
-		this.currentEntityType = currentEntityType;
+		return !MSTags.EntityTypes.REMOTE_OBSERVER_WHITELIST.contains(currentEntityType);
```
kirderf1, line 135: Based on usage, this tag appears to be a blacklist and not a whitelist.

kirderf1 (changes requested):

src/main/java/com/mraof/minestuck/tileentity/redstone/AreaEffectTileEntity.java
Dweblenod: Should be fixed now

src/main/java/com/mraof/minestuck/tileentity/redstone/StatStorerTileEntity.java
```diff
@@ -82,7 +83,7 @@ public void tick()
 		if(level == null || !level.isAreaLoaded(worldPosition, 1))
 			return;
 		
-		if(tickCycle % 6 == 1) //6 is wireless constant
+		if(tickCycle % WIRELESS_TICK_FREQUENCY == 1)
```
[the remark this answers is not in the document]
Dweblenod: I suppose "wireless" is not relevant to this tile entity but it uses the same refresh rate as wireless redstone components for consistency

src/main/java/com/mraof/minestuck/tileentity/redstone/WirelessRedstoneTransmitterTileEntity.java
Dweblenod: Do you think its an issue to store the facing?

src/main/java/com/mraof/minestuck/tileentity/redstone/WirelessRedstoneTransmitterTileEntity.java
Dweblenod: Not sure what you mean

src/main/java/com/mraof/minestuck/tileentity/redstone/WirelessRedstoneTransmitterTileEntity.java
Dweblenod: Moved it around so that it checks for the absolute coordinates first and then put the offset in the else statement

src/main/java/com/mraof/minestuck/tileentity/redstone/WirelessRedstoneTransmitterTileEntity.java
kirderf1: Instead of getting the block state from the world/level, get it as a function argument to `setOffsetFromDestinationBlockPos()`.

src/main/java/com/mraof/minestuck/tileentity/redstone/WirelessRedstoneTransmitterTileEntity.java
kirderf1: Not necessarily, and I do think what you've written works correctly. Rather, I think that the strategy with the cached facing makes the tile entity more complicated than it has to be for no clear benefit.
(I also see it as a bonus to change `getDestinationBlockPosFromOffset()` to not have [side effects](https://en.wikipedia.org/wiki/Side_effect_(computer_science)), as it's not the type of function that I'd expect to have side effects)

src/main/java/com/mraof/minestuck/tileentity/redstone/StatStorerTileEntity.java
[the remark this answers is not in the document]
kirderf1: Consistency in what sense?
That in order for them to work correctly together, they need the same refresh rate?
That a player or dungeon builder would expect the same refresh rate across all machine types for their builds?
Or simply that they behave similarly, and therefore may have the same refresh rate?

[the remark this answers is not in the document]
kirderf1: Alright, then it might make sense to declare it as a constant somewhere.
`public static final int PUZZLE_BLOCK_TICK_RATE = 6;`
or
`public static final int STANDARD_TICK_RATE = 6;`
in either `MSTileEntityTypes` or `MinestuckConfig`.
And then you could either refer to this constant directly in the puzzle blocks, or define and use a block-specific constant in each relevant puzzle block.
`private static final int TICK_RATE = MinestuckConfig.PUZZLE_BLOCK_TICK_RATE;`
How does that sound?

src/main/java/com/mraof/minestuck/client/gui/AreaEffectScreen.java
```diff
@@ -29,63 +34,112 @@ public class AreaEffectScreen extends Screen
...
+		addButton(effectTextField);
+		
+		this.effectAmplifierTextField = new TextFieldWidget(this.font, this.width / 2 + 45, yOffset + 79, 20, 18, new StringTextComponent("Current Effect Amplifier"));
+		this.effectAmplifierTextField.setValue(String.valueOf(te.getEffectAmplifier()));
+		addButton(effectAmplifierTextField);
+		
+		if(isAllMobs)
+			addButton(allMobsButton = new ExtendedButton(this.width / 2 - 65, yOffset + 105, 85, 20, new StringTextComponent("ALL MOBS"), button -> cycleIsAllMobs()));
+		else
+			addButton(allMobsButton = new ExtendedButton(this.width / 2 - 65, yOffset + 105, 85, 20, new StringTextComponent("JUST PLAYERS"), button -> cycleIsAllMobs()));
```
kirderf1, line 93: `isAllMobs` only determines the text on the button. With this in mind, you can simplify this part of the code.

kirderf1: With a helper function
```java
private void updateAllMobsButtonMessage()
{
	if(isAllMobs)
		allMobsButton.setMessage(new StringTextComponent("ALL MOBS"));
	else
		allMobsButton.setMessage(new StringTextComponent("JUST PLAYERS"));
}
```
you can simplify the initialization of the button to
 ```java
 addButton(allMobsButton = new ExtendedButton(this.width / 2 - 65, yOffset + 105, 85, 20, StringTextComponent.EMPTY, button -> cycleIsAllMobs()));
 updateAllMobsButtonMessage();
 ```
 And also use that helper function in `cycleIsAllMobs()` to make sure that the button message is only defined in one place.

kirderf1 (changes requested):
Another partial review, but I should be able to do them more frequently now.
How close is this PR to no longer be considered a draft?

src/main/java/com/mraof/minestuck/tileentity/redstone/StatStorerTileEntity.java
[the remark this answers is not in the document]
Dweblenod: Sounds good to me, I might put it in the config and allow a minimal level of control over the rate

Dweblenod:
> Another partial review, but I should be able to do them more frequently now. How close is this PR to no longer be considered a draft?

This can be considered no longer a draft when a block pressure plate item(that has been discussed amongst the dev team)

src/main/java/com/mraof/minestuck/block/redstone/TogglerBlock.java
```diff
@@ -0,0 +1,148 @@
...
+	}
+	
+	@Override
+	public void onPlace(BlockState state, World worldIn, BlockPos pos, BlockState oldState, boolean isMoving)
+	{
+		super.onPlace(state, worldIn, pos, oldState, isMoving);
+		updatePower(worldIn, pos);
+	}
+	
+	public void updatePower(World worldIn, BlockPos pos)
```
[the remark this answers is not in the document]
kirderf1: For example, some of the code blocks that you added an overview comment to could ideally be extracted ad functions.

src/main/java/com/mraof/minestuck/block/redstone/TogglerBlock.java
```diff
@@ -0,0 +1,148 @@
...
+			{
+				BlockPos facingPos = pos.relative(state.getValue(FACING));
+				BlockState facingState = worldIn.getBlockState(facingPos);
+				
+				if(state.getValue(DISCHARGE))
+				{
+					boolean makeSound = false;
+					
+					//removing redstone power from powered blocks
+					for(Property<?> property : facingState.getProperties())
```
kirderf1, line 94: Rewriting this section to use `facingState.hasProperty()` should make this section much simpler.

src/main/java/com/mraof/minestuck/block/redstone/TogglerBlock.java
```diff
@@ -0,0 +1,148 @@
...
+						}
+					}
+					
+					if(makeSound)
+					{
+						worldIn.playSound(null, facingPos, SoundEvents.UI_BUTTON_CLICK, SoundCategory.BLOCKS, 0.5F, 0.5F);
+					}
+				} else
+				{
+					for(Property<?> property : facingState.getProperties())
```
kirderf1, line 119: Rewriting this section to use `facingState.hasProperty()` should make this section simpler.

src/main/java/com/mraof/minestuck/client/gui/RemoteObserverScreen.java
```diff
@@ -8,50 +8,61 @@
...
-	public Button typeButton;
+	private Button typeButton;
+	private Button incrementButton;
+	private Button decrementButton;
+	private Button largeIncrementButton;
+	private Button largeDecrementButton;
 	
 	private TextFieldWidget entityTypeTextField;
-	
+	private boolean shouldFinish = true;
```
kirderf1, line 37: Not a very clean solution, as it's only there to give extra information to the return value of `getEntityType()`. There may be a better way of handling that entity type.

src/main/java/com/mraof/minestuck/tileentity/redstone/AreaEffectTileEntity.java
```diff
@@ -37,7 +40,7 @@ public void tick()
 		if(level == null || !level.isAreaLoaded(getBlockPos(), 0))
 			return;
 		
-		if(level.getGameTime() % 80 == 0 && !level.isClientSide && level.getBlockState(getBlockPos()).getBlock() instanceof AreaEffectBlock && getBlockState().getValue(AreaEffectBlock.POWERED))
+		if(level.getGameTime() % 80 == 0 && !level.isClientSide && level.getBlockState(getBlockPos()).getBlock() instanceof AreaEffectBlock && getBlockState().getValue(AreaEffectBlock.POWERED) && !getBlockState().getValue(AreaEffectBlock.SHUT_DOWN))
```
kirderf1, line 43: I think that you should be able to assume that `level.getBlockState(getBlockPos()).getBlock() instanceof AreaEffectBlock` cannot be false

src/main/java/com/mraof/minestuck/tileentity/redstone/RemoteComparatorTileEntity.java
```diff
@@ -0,0 +1,83 @@
...
+import net.minecraft.nbt.CompoundNBT;
+import net.minecraft.network.NetworkManager;
+import net.minecraft.network.play.server.SUpdateTileEntityPacket;
+import net.minecraft.tileentity.ITickableTileEntity;
+import net.minecraft.tileentity.TileEntity;
+
+public class RemoteComparatorTileEntity extends TileEntity implements ITickableTileEntity
+{
+	private int tickCycle;
+	private static final int TICK_FREQUENCY = MinestuckConfig.SERVER.puzzleBlockTickRate.get();
```
kirderf1, line 16: Since it's a config value, it might change during the course of the game. As such, you shouldn't really define a local constant like this. Use the config value directly instead.

Dweblenod:
Despite me having marked the PR as open, I will note that textures are still needed for the structure core

src/main/java/com/mraof/minestuck/tileentity/redstone/RemoteComparatorTileEntity.java
```diff
@@ -0,0 +1,82 @@
...
+	
+	private void sendUpdate()
+	{
+		if(level != null)
+		{
+			boolean shouldBePowered = RemoteComparatorBlock.isMatch(level, worldPosition);
+			
+			if(getBlockState().getValue(RemoteComparatorBlock.POWERED) != shouldBePowered)
+				level.setBlockAndUpdate(getBlockPos(), getBlockState().setValue(RemoteComparatorBlock.POWERED, shouldBePowered));
+			else level.sendBlockUpdated(getBlockPos(), getBlockState(), getBlockState(), 2);
```
kirderf1, line 44: If the this entity doesn't need any data synced to client-side, then the call to `sendBlockUpdated()` here doesn't seem to fill a purpose.

src/main/java/com/mraof/minestuck/tileentity/redstone/WirelessRedstoneTransmitterTileEntity.java
```diff
@@ -140,11 +161,11 @@ public CompoundNBT save(CompoundNBT compound)
 		
 		compound.putInt("tickCycle", tickCycle);
 		
-		getDestinationBlockPos();
+		getDestinationBlockPosFromOffset();
```
kirderf1, line 164: From what I can tell, this call here is only there to make sure that `this.offsetPos` is not null (or rather, set to the zero pos if it is null).
I'd suggest instantiating `this.offsetPos` so that it can never be null. Doing so would remove the need of this call and remove the need for a null check in `getDestinationBlockPosFromOffset()`.

src/main/java/com/mraof/minestuck/tileentity/redstone/WirelessRedstoneReceiverTileEntity.java
```diff
@@ -55,21 +55,30 @@ public void setLastTransmitterBlockPos(BlockPos lastTransmitterPosIn)
 	 */
 	public void renewFromLastTransmitter()
 	{
-		if(level != null && lastTransmitterBlockPos != null && !level.isClientSide && level.isAreaLoaded(lastTransmitterBlockPos, 1))
+		if(level != null && !level.isClientSide)
 		{
-			TileEntity tileEntity = level.getBlockEntity(lastTransmitterBlockPos);
-			if(tileEntity instanceof WirelessRedstoneTransmitterTileEntity)
+			BlockState state = getBlockState();
+			BlockState newState = WirelessRedstoneReceiverBlock.setPower(state, 0);
```
kirderf1, line 61: This is only the new state if it can't find the transmitter. Perhaps call it `unpoweredState`?

src/main/java/com/mraof/minestuck/tileentity/redstone/StructureCoreTileEntity.java
```diff
@@ -0,0 +1,256 @@
...
+				{
+					CoreCompatibleScatteredStructurePiece piece = getStructurePiece(structureStart);
+					
+					if(piece != null && piece.hasBeenCompleted())
+					{
+						if(!getBlockState().getValue(StructureCoreBlock.POWERED))
+							level.setBlockAndUpdate(getBlockPos(), getBlockState().setValue(StructureCoreBlock.POWERED, true));
+					} else
+					{
+						level.setBlockAndUpdate(getBlockPos(), getBlockState().setValue(StructureCoreBlock.POWERED, false));
```
kirderf1, line 108: Is there a reason that there's a blockstate check when powering the block, but not when unpowering it?

src/main/java/com/mraof/minestuck/tileentity/redstone/StructureCoreTileEntity.java
```diff
@@ -0,0 +1,256 @@
...
+				}
+			}
+		}
+	}
+	
+	public StructureStart<?> getStructureStart(StructureManager structureManager)
+	{
+		for(Structure<?> structureListIterate : SOLVABLE_STRUCTURES)
+		{
+			return structureManager.getStructureAt(getBlockPos(), true, structureListIterate);
```
kirderf1, line 119: This'll just return for the first structure in `SOLVABLE_STRUCTURES`.
Also, have you considered configuring each individual structure core for a specific structure, instead of having a global list of affected structures?

src/main/java/com/mraof/minestuck/tileentity/redstone/StructureCoreTileEntity.java
```diff
@@ -0,0 +1,256 @@
...
+	{
+		//compoundnbt.putString("id", Registry.STRUCTURE_FEATURE.getKey(this.getFeature()).toString());
+		
+		CoreCompatibleScatteredStructurePiece piece = getStructurePiece(structureStart);
+		if(piece != null)
+		{
+			if(!piece.hasBeenCompleted())
+				piece.nowCompleted();
+			else
+				Debug.debugf("has already been completed");
```
kirderf1, line 135: Usages of `Debug` are being replaced. Remove it if you're done with it, or replace it by a local logger.

src/main/java/com/mraof/minestuck/tileentity/redstone/StructureCoreTileEntity.java
```diff
@@ -0,0 +1,256 @@
...
+			if(pieceIterate instanceof CoreCompatibleScatteredStructurePiece)
+			{
+				return (CoreCompatibleScatteredStructurePiece) pieceIterate;
+			}
+		}
+		
+		return null;
+	}
+	
+	private void wipeSlate()
```
kirderf1, line 153: The behavior of this function doesn't seem to match the javadoc for `READ_AND_WIPE`, as it doesn't check if the structure has been completed.

src/main/java/com/mraof/minestuck/tileentity/redstone/StructureCoreTileEntity.java
```diff
@@ -0,0 +1,256 @@
...
+	{
+		if(level != null)
+		{
+			for(BlockPos iteratePos : BlockPos.betweenClosed(getBlockPos().offset(shutdownRange, shutdownRange, shutdownRange), getBlockPos().offset(-shutdownRange, -shutdownRange, -shutdownRange)))
+			{
+				BlockState iterateState = level.getBlockState(iteratePos);
+				
+				//removing functionality from blocks that should have limited survival access
+				for(Property<?> property : iterateState.getProperties())
```
kirderf1, line 162: Rewrite this with `blockState.hasProperty()`

kirderf1 (changes requested):
I recall bringing up a suggestion to replace use of `tickCycle` with a module of the world time (essentially `level.getGameTime() % tickRate == 0`). Did you have any thoughts on it?

src/main/java/com/mraof/minestuck/block/redstone/AreaEffectBlock.java
```diff
@@ -133,9 +119,15 @@ public void updatePower(World worldIn, BlockPos pos)
 			BlockState state = worldIn.getBlockState(pos);
 			boolean hasPower = worldIn.hasNeighborSignal(pos);
 			
-			if(state.getValue(POWERED) != hasPower)
-				worldIn.setBlockAndUpdate(pos, state.setValue(POWERED, hasPower));
-			else worldIn.sendBlockUpdated(pos, state, state, 2);
+			if(state.getValue(SHUT_DOWN))
+			{
+				worldIn.setBlockAndUpdate(pos, state.setValue(POWERED, false));
```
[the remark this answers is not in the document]
kirderf1: Idea for simplification: incorporate `SHUT_DOWN` into `hasPower`. `boolean hasPower = !state.getValue(SHUT_DOWN) && worldIn.hasNeighborSignal(pos);`

src/main/java/com/mraof/minestuck/client/gui/AreaEffectScreen.java
```diff
@@ -29,63 +35,112 @@ public class AreaEffectScreen extends Screen
...
 	
 	private void finish()
 	{
-		MSPacketHandler.sendToServer(new AreaEffectPacket(parseMinBlockPos(), parseMaxBlockPos(), te.getBlockPos()));
+		int minX = parseInt(minPosDestinationTextFieldX);
+		int minY = parseInt(minPosDestinationTextFieldY);
+		int minZ = parseInt(minPosDestinationTextFieldZ);
+		int maxX = parseInt(maxPosDestinationTextFieldX);
+		int maxY = parseInt(maxPosDestinationTextFieldY);
+		int maxZ = parseInt(maxPosDestinationTextFieldZ);
```
kirderf1, line 138: Do you think you could change the position parsing to something similar to what I made for the sendificator screen? Specifically with the goal of rejecting invalid input? (as opposed to silently changing invalid coords to 0 which happens right now) You can also consider whether or not you want to change the clickability of the finish-button.

src/main/java/com/mraof/minestuck/client/gui/RemoteObserverScreen.java
```diff
@@ -66,33 +76,41 @@ private void changeActiveType()
...
 		super.render(matrixStack, mouseX, mouseY, partialTicks);
 	}
 	
 	private void finish()
 	{
-		MSPacketHandler.sendToServer(new RemoteObserverPacket(activeType, te.getBlockPos(), getEntityType(entityTypeTextField.getValue())));
-		onClose();
+		Optional<EntityType<?>> attemptedEntityType = EntityType.byString(entityTypeTextField.getValue());
+		
+		if(attemptedEntityType.isPresent() && RemoteObserverTileEntity.entityCanBeObserved(attemptedEntityType.get()))
```
kirderf1, line 107: From my understanding of this screen, the entity type should only matter and be used for the `CURRENT_ENTITY_PRESENT` type, but here you require that there is a valid entity type regardless of which observer type is being set.

kirderf1 (changes requested):

src/main/java/com/mraof/minestuck/tileentity/redstone/StructureCoreTileEntity.java
Dweblenod: I have no memory of whether its intentional, will add a check before the unpowering line

src/main/java/com/mraof/minestuck/tileentity/redstone/StructureCoreTileEntity.java
Dweblenod: May take that consideration into effect at a later date

src/main/java/com/mraof/minestuck/tileentity/redstone/StructureCoreTileEntity.java
Dweblenod: Hope I used the right one

src/main/java/com/mraof/minestuck/tileentity/redstone/WirelessRedstoneTransmitterTileEntity.java
Dweblenod: Upon creation there is a default blockpos set to the variable now

src/main/java/com/mraof/minestuck/tileentity/redstone/StructureCoreTileEntity.java
Dweblenod: Actually ended up removing them, clogged things up

src/main/java/com/mraof/minestuck/tileentity/redstone/StructureCoreTileEntity.java
Dweblenod: Should be improved now, all functions related to a specific enum type have been reorganized and now include some of the same checks

src/main/java/com/mraof/minestuck/client/gui/AreaEffectScreen.java
Dweblenod: Probably not exactly what you were suggesting, but I added the red text coloration change incorporated into other puzzle block screens

src/main/java/com/mraof/minestuck/client/gui/RemoteObserverScreen.java
Dweblenod: It can now move forward with an invalid entity type present as long as the active type is not CURRENT_ENTITY_PRESENT. It will stick with the original entity type stored in the tile entity if its invalid but regardless of active type it will move forward with the widget entity type if it passes the original checks

src/main/java/com/mraof/minestuck/tileentity/redstone/StructureCoreTileEntity.java
```diff
@@ -0,0 +1,255 @@
...
+					{
+						if(piece.hasBeenCompleted())
+						{
+							if(!getBlockState().getValue(StructureCoreBlock.POWERED))
+								level.setBlockAndUpdate(getBlockPos(), getBlockState().setValue(StructureCoreBlock.POWERED, true));
+						} else
+						{
+							if(getBlockState().getValue(StructureCoreBlock.POWERED))
+								level.setBlockAndUpdate(getBlockPos(), getBlockState().setValue(StructureCoreBlock.POWERED, false));
+						}
```
kirderf1, line 112: These two code blocks are very similar and can be simplified. Two equivalent suggestions:
```java
if(piece.hasBeenCompleted() != getBlockState().getValue(StructureCoreBlock.POWERED))
	level.setBlockAndUpdate(getBlockPos(), getBlockState().setValue(StructureCoreBlock.POWERED, piece.hasBeenCompleted()));
```
```java
BlockState newState = getBlockState().setValue(StructureCoreBlock.POWERED, piece.hasBeenCompleted());
if(newState != getBlockState())
	level.setBlockAndUpdate(getBlockPos(), newState);
```

src/main/java/com/mraof/minestuck/client/gui/AreaEffectScreen.java
kirderf1: Yep, not exactly what I was suggesting, but it does serve the purpose of rejecting invalid input and giving some feedback to the player that something is invalid.

src/main/java/com/mraof/minestuck/tileentity/redstone/StructureCoreTileEntity.java
```diff
@@ -0,0 +1,255 @@
...
+		{
+			return structureManager.getStructureAt(getBlockPos(), true, structureListIterate);
+		}
+		
+		return null;
+	}
+	
+	private void writeToStructure(CoreCompatibleScatteredStructurePiece piece)
+	{
+		//compoundnbt.putString("id", Registry.STRUCTURE_FEATURE.getKey(this.getFeature()).toString());
```
kirderf1, line 131: Commented code that doesn't seem to belong.

kirderf1 (changes requested):
I haven't left any files unreviewed this time, so it shouldn't be long now until it's good to go.

src/main/java/com/mraof/minestuck/tileentity/redstone/StructureCoreTileEntity.java
```diff
@@ -0,0 +1,255 @@
...
+		WRITE;
+		
+		public static ActionType fromInt(int ordinal) //converts int back into enum
+		{
+			for(ActionType type : ActionType.values())
+			{
+				if(type.ordinal() == ordinal)
+					return type;
+			}
+			throw new IllegalArgumentException("Invalid ordinal of " + ordinal + " for structure core action type!");
```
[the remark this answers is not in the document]
kirderf1: Oops, meant to be the length of the values array. `ActionType.values().length`

src/main/java/com/mraof/minestuck/block/redstone/PlatformBlock.java
```diff
@@ -55,24 +61,77 @@ public BlockRenderType getRenderShape(BlockState state)
...
+						!(supportingState.getBlock() instanceof PlatformGeneratorBlock))
+					world.setBlockAndUpdate(pos, Blocks.AIR.defaultBlockState()); //removes platform block if there is no platform generator, or if a platform generator is there but it is either not powered or not facing the platform block
+				else if(supportingState.getBlock() instanceof PlatformGeneratorBlock && supportingState.getValue(PlatformGeneratorBlock.POWERED)) //switches which kind of platform block is present if the Invisibility mode does not match
+				{
+					if(supportingState.getValue(PlatformGeneratorBlock.INVISIBLE_MODE) != state.getValue(INVISIBLE) && supportingState.getValue(PlatformGeneratorBlock.FACING) == stateFacing)
+						world.setBlockAndUpdate(pos, state.setValue(INVISIBLE, supportingState.getValue(PlatformGeneratorBlock.INVISIBLE_MODE))); //TODO Visible platforms should override invisible ones
+					if(!supportingState.getValue(PlatformGeneratorBlock.INVISIBLE_MODE) && state.getValue(INVISIBLE) && supportingState.getValue(PlatformGeneratorBlock.FACING) != stateFacing)
+						world.setBlockAndUpdate(pos, state.setValue(INVISIBLE, false));
```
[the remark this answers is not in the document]
Dweblenod: Created several booleans to improve legibility, revealing multiple redundancies in code in the process which have been removed

src/main/java/com/mraof/minestuck/tileentity/redstone/StructureCoreTileEntity.java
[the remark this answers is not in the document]
Dweblenod: Done!

src/main/java/com/mraof/minestuck/tileentity/redstone/RemoteObserverTileEntity.java
```diff
@@ -135,6 +152,7 @@ public void load(BlockState state, CompoundNBT compound)
 		
 		this.tickCycle = compound.getInt("tickCycle");
 		this.activeType = ActiveType.fromInt(compound.getInt("activeTypeOrdinal"));
+		observingRange = compound.getInt("observingRange");
```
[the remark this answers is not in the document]
Dweblenod: Added an if-else that sets the range to 15 if the key does not exist in nbt

src/main/java/com/mraof/minestuck/block/redstone/PlatformBlock.java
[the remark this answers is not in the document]
Dweblenod: Turns out the second if statement here could never even be reached to begin with

src/main/java/com/mraof/minestuck/tileentity/redstone/RemoteObserverTileEntity.java
```diff
@@ -77,33 +81,28 @@ public void checkRelaventType()
...
+				else if(activeType == ActiveType.IS_CROUCHING && entity.isCrouching())
+					shouldBePowered = true;
+				else if(activeType == ActiveType.IS_ENTITY_BURNING && entity.isOnFire())
+					shouldBePowered = true;
+				else if(activeType == ActiveType.IS_ENTITY_INVISIBLE && entity.isInvisible())
+					shouldBePowered = true;
+				else if(activeType == ActiveType.IS_ELYTRA_FLYING && (entity instanceof LivingEntity && ((LivingEntity) entity).isFallFlying()))
+					shouldBePowered = true;
+				else if(activeType == ActiveType.IS_ENTITY_IN_WATER && entity.isInWater())
+					shouldBePowered = true;
```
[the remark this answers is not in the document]
Dweblenod: Works now, good suggestion!

src/main/java/com/mraof/minestuck/block/redstone/PlatformBlock.java
```diff
@@ -55,24 +61,86 @@ public BlockRenderType getRenderShape(BlockState state)
...
+					generatorHasPower = supportingState.getValue(PlatformGeneratorBlock.POWERED);
+				}
+				
+				if(!supportingStateIsGenerator || (!generatorHasPower || !generatorHasSameFacing))
+					world.setBlockAndUpdate(pos, Blocks.AIR.defaultBlockState()); //removes platform block if there is no platform generator, or if a platform generator is there but it is either not powered or not facing the platform block
+				else
+				{
+					boolean generatorSetToInvisible = supportingState.getValue(PlatformGeneratorBlock.INVISIBLE_MODE);
+					boolean thisSetToInvisible = state.getValue(INVISIBLE);
+					boolean differentInvisibilityStatuses = generatorSetToInvisible != thisSetToInvisible;
```
kirderf1, line 95: Variable introduction is slightly overdone with `differentInvisibilityStatuses`. I'd suggest inlining this variable into the if-statement.

src/main/java/com/mraof/minestuck/tileentity/redstone/RemoteObserverTileEntity.java
```diff
@@ -17,34 +20,48 @@
...
+		IS_UNDERLING_PRESENT((entity, observer) -> entity instanceof UnderlingEntity),
+		IS_ENTITY_BURNING((entity, observer) -> entity.isOnFire()),
+		IS_ENTITY_INVISIBLE((entity, observer) -> entity.isInvisible()),
+		IS_ELYTRA_FLYING((entity, observer) -> entity instanceof LivingEntity && ((LivingEntity) entity).isFallFlying()),
+		IS_ENTITY_UNDER_WATER((entity, observer) -> entity.isUnderWater()),
+		IS_ENTITY_WET((entity, observer) -> entity.isInWaterRainOrBubble()),
+		IS_ENTITY_ON_GROUND((entity, observer) -> entity.isOnGround()),
+		IS_SPRINTING((entity, observer) -> entity.isSprinting()); //TODO IS_BOSS_PRESENT
+		
+		BiPredicate<Entity, RemoteObserverTileEntity> typeConditions;
```
kirderf1, line 52: I suspect that this field should be private final.

src/main/java/com/mraof/minestuck/tileentity/redstone/StructureCoreTileEntity.java
kirderf1: Alright, but I'd still like some small fix for the immediate problem. Either:
- Use a single structure instead of a list of structures
- Fix the loop so that all entries in the list can get used (by returning the result of `getStructureAt()` only if it is not null)

kirderf1 (changes requested):

kirderf1 (approved):

PR #333, opened 2022-05-17
alchemiter animations + remodel

amiwyn:


Dweblenod:
Looking nice! Found a couple visual bugs so far that I will describe via text here. Will send pictures to you personally to better show what said messages mean.

- I believe the red bone on the head of the alchemiter arm represents a laser, based on the comic and on Riotmode's gif that shows the animation in action. That being said, even when in stationary mode the laser mode is sticking out

- Since the alchemiter arm is not a block, it actually stops being rendered if no other components of the alchemiter are visible which can occur most often when the player is standing near it. Related to this, it feels awkward(I believe players would find it acceptable at the end of the day if left as is) that there is no collision box where the arm is in stationary mode, since that means you can walk right through that part of the model. Unsure what would be the best remedy, but perhaps expanding the collision box of the closest block(while it is stationary at least) in the alchemiter would fix both provided the collision box factors into rendering(I have no idea if it does or not, but based on the following issue it does actually seem like it would work)

- On the 4 center edges of the alchemiter base, there is a raised section surrounded by two smaller ones. That raised section spans between the two center edge blocks but only 1 of the two blocks has it rendered and has a collision box to match it. This would mean that if the other center block was destroyed it overhangs

- There may have been a similar problem with the old alchemiter, but this one seems prone to inconsistent lighting resulting in neighboring blocks within the alchemiter having a contrast they shouldnt have given the circumstances. A similar issue occurs with the steep stairs of the frog temple so in general it might be good to get an idea of what conditions result in strange lighting(not your responsibility dont worry)

- The gears and machinery textures inside the alchemiter seem randomized which is fine overall, but at the moment it results in the insides looking discontinuous from block to block 

- The bone at the very base of the alchemiters arm intersects with the base, which results in Z-fighting when neighboring blocks are removed. Strictly a fix a modeler needs to fix, but warning for your future

SpoiledMysteryMeat:
confirming the red = laser thing, the animation should by default hide it so if it's not doing that I have to fix that I think.

Second point, the arm actually has a static model variant to account for that which should replace the entity when not animating.

the overhang is also intentional, since I want the entire machine to delete if you break part of it. There's no way to make the overhang not do that without inflating cubecounts.

The interiors shouldn't be randomized, I'm not sure why they're giving that impression.

Z-fighting will be fixed at a model-level.

Dweblenod:
Continuing on Riots follow up, for the second point do you mean to say that there is a physical block change that takes place when static or just that the visual appearance/model is changed when not doing something animated.

With the mention of the breaking of the whole machine, I would like to check with you Riot and some other coders to figure out if that proposed method is the best way to balance the coding and modeling side of things.

With the mention of interiors seeming randomized, I will send a picture to the dev team so that we can get on the same page about that(less work needed for you Ami!)

kirderf1 (approved):

Dweblenod:
Once the MSBlockShapes conflict is resolved, I would consider this ready to merge. 

There is still Z-fighting, issues with the model and collision boxes extending into another blockspace, and a suggestion that the arm of the alchemiter should be collide-able when at rest. However those first two are modeling considerations primarily and the last one is still debatable so I included them in the main geckolib PR notes so they are not forgotten and so this can move forward

SpoiledMysteryMeat:
the former, it replaces the entity with a block.

PR #334, opened 2022-05-18
cruxtruder remodel

amiwyn:
some texture fixes are needed
the weird render thing is still there


src/main/java/com/mraof/minestuck/block/MSBlockShapes.java
```diff
@@ -69,13 +69,11 @@
...
+	public static final CustomVoxelShape CRUXTRUDER_BASE = new CustomVoxelShape(new double[]{0,0,0,16,16,16});
+	public static final CustomVoxelShape CRUXTRUDER_TOP_CORNER = new CustomVoxelShape(new double[]{10, 0, 10, 16, 10, 16}, new double[]{12, 10, 12, 16, 12, 16});
+	public static final CustomVoxelShape CRUXTRUDER_TOP_SIDE = new CustomVoxelShape(new double[]{0, 0, 10, 16, 10, 16}, new double[]{0, 10, 12, 16, 12, 16}, new double[]{2, -0.45672, 4.2961, 14, 5.54328, 10.2961});
+	public static final CustomVoxelShape CRUXTRUDER_TOP_CENTER = new CustomVoxelShape(new double[]{0, -3, 0, 16, 9, 16}, new double[]{3, 9, 3, 13, 10, 13}, new double[]{2, 10, 2, 14, 13, 14});
+	public static final CustomVoxelShape CRUXTRUDER_TUBE = new CustomVoxelShape(new double[]{2, 0, 2, 14, 16, 14}, new double[]{7.5, 8.5, 0, 8.5, 9.5, 2}, new double[]{6, 7, -0.005, 10, 11, -0.005}, new double[]{6, 11, -1, 10, 12, 0}, new double[]{6, 6, -1, 10, 7, 0}, new double[]{5, 7, -1, 6, 11, 0}, new double[]{10, 7, -1, 11, 11, 0});
```
kirderf1, line 76: These `-0.005` coordinates are slightly unexpected. Is there a reason why these are simply not `0`?

kirderf1 (commented):
The code looks fine except for one question. It would be nice cleaning up the model representation by giving proper texture names (instead of "0"), and also if the texture could possibly be split up into several textures in some sensible manner.

src/main/java/com/mraof/minestuck/block/MSBlockShapes.java
amiwyn: I simply copied the values that were in the model file

amiwyn: do you think it rounds up to 0 in that case? i never tested it

kirderf1: In the model, it might have been offset slightly to avoid some case of z-fighting (I haven't looked at the model enough to say if that is the case)
For the wireframe highlight and collision box that these shapes are used for, I don't think this offset serves the same purpose.

amiwyn: does it matter? i can round it up if you want

kirderf1: Not really, I suppose. It struck me as a code cleanup thing, but it's not really important in any sense.

PR #341, opened 2022-05-27
Totemlathe model + animations

amiwyn:


src/main/java/com/mraof/minestuck/tileentity/machine/TotemLatheTileEntity.java
```diff
@@ -20,333 +20,319 @@
...
+    }
+
+    private <E extends TileEntity & IAnimatable> PlayState carveAnimation(AnimationEvent<E> event) {
+        if (this.isProcessing) {
+            event.getController().setAnimation(new AnimationBuilder().addAnimation("carvetotem", false));
+            return PlayState.CONTINUE;
+        }
+        event.getController().markNeedsReload();
+        return PlayState.STOP;
+    }
```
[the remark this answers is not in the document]
Dweblenod: Having a timer could also give us the chance to add a sound effect

amiwyn:
i have to change the main tile entity position so it lines up with the center block instead

kirderf1:
My first thought is backwards-compability. What impact will it have on previous totem lathes? Can it be done in a way where any items in the original tile entity isn't lost?

amiwyn:
youll have to place a new one, the old one is 1x3x4 and the new one is 1x3x3

kirderf1:
Yeah. Having to replace the old one is fine by me. What about items in the old one?

amiwyn:
youll lose the dowel but you keep the card

kirderf1:
Can you make sure that both items can drop? Also, since some blocks will dissapear when shrinking the totem lathe, make sure that the blocks still exist as dummy blocks, or to handle the removed blocks in the `MissingMappings` event.

amiwyn:
Is it really necessary? it works fine without handling the removed blocks plus the player would only lose one cruxite worth of materials. the player would still need to remove the old machine and put down the new one in either cases

kirderf1:
It is preferable. You should at the very least consider what it would take to preserve the dowel.

Dweblenod:
Has the discussion above been resolved? If so, I would be fine merging this PR after the conflicting files(MSBlockShapes and ClientProxy) are also resolved

amiwyn:
the items now drop on block break so the issue is resolved

kirderf1 (commented):
Became too tired for the day before getting everything reviewed. Gonna try and review the rest at some point.

src/main/java/com/mraof/minestuck/tileentity/machine/TotemLatheTileEntity.java
```diff
@@ -193,14 +165,19 @@ public void onRightClick(PlayerEntity player, BlockState clickedState)
 			handleSlotClick(player, working);
 		
 		//if they have clicked the dowel block
-		if(clickedState.getBlock() == MSBlocks.TOTEM_LATHE.ROD.get() || clickedState.getBlock() == MSBlocks.TOTEM_LATHE.DOWEL_ROD.get())
+		if(clickedState.getBlock() == MSBlocks.TOTEM_LATHE.DOWEL_ROD.get() || clickedState.getBlock() == MSBlocks.TOTEM_LATHE.DOWEL_ROD.get())
```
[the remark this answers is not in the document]
amiwyn: how did i pulled that off 😂

src/main/java/com/mraof/minestuck/client/renderer/tileentity/TotemLatheRenderer.java
```diff
@@ -0,0 +1,92 @@
...
+			RenderUtils.translate(bone, stack);
+			RenderUtils.moveToPivot(bone, stack);
+			stack.translate(0.25, -0.375, 0.03125);
+			stack.mulPose(Vector3f.ZP.rotationDegrees(90));
+			
+			// render the dowel with the blockRenderer
+			ClientWorld level = Minecraft.getInstance().level;
+			BlockRendererDispatcher blockRenderer = Minecraft.getInstance().getBlockRenderer();
+			BlockPos pos = lathe.getBlockPos();
+			IModelData modelData = new ModelDataMap.Builder().build();
```
kirderf1, line 83: Looks like you could use `EmptyModelData.INSTANCE` instead

src/main/java/com/mraof/minestuck/client/renderer/tileentity/TotemLatheRenderer.java
```diff
@@ -0,0 +1,92 @@
...
+			blockRenderer.renderModel(cruxiteDowel, pos, level, stack, renderTypeBuffer.getBuffer(RenderTypeLookup.getRenderType(cruxiteDowel, false)), false, level.random, modelData);
+			
+			stack.popPose();
+			renderTypeBuffer.getBuffer(RenderType.entityCutoutNoCull(this.getTextureLocation(lathe)));
```
kirderf1, line 87: Looks like you're getting a render buffer without using it. Why?

src/main/java/com/mraof/minestuck/tileentity/machine/TotemLatheDowelTileEntity.java
```diff
@@ -0,0 +1,96 @@
...
+public class TotemLatheDowelTileEntity extends ItemStackTileEntity implements IAnimatable
+{
+	private final AnimationFactory factory = new AnimationFactory(this);
+	
+	public TotemLatheDowelTileEntity()
+	{
+		super(MSTileEntityTypes.TOTEM_LATHE_DOWEL.get());
+	}
+	
+	private TotemLatheTileEntity getTotemLatheEntity()
```
kirderf1, line 33: I'd recommend a `@Nullable` annotation (though it's not strictly necessary)

src/main/java/com/mraof/minestuck/tileentity/machine/TotemLatheTileEntity.java
```diff
@@ -119,49 +110,30 @@ public boolean setDowel(ItemStack stack)
...
+				te = new ItemStackTileEntity();
+				level.setBlockEntity(pos, te);
 			}
+			ItemStackTileEntity teItem = (ItemStackTileEntity) te;
+			teItem.setStack(stack);
+			if(!state.equals(newState))
+				level.setBlockAndUpdate(pos, newState);
+			else level.sendBlockUpdated(pos, state, state, 2);
+			isProcessing = false;
+			level.sendBlockUpdated(getBlockPos(), getBlockState(), getBlockState(), Constants.BlockFlags.BLOCK_UPDATE);
```
kirderf1, line 128:
```java
level.sendBlockUpdated(getBlockPos(), getBlockState(), getBlockState(), Constants.BlockFlags.BLOCK_UPDATE);
```
appear to be a duplication of what happens in
```java
if(!state.equals(newState))
	level.setBlockAndUpdate(pos, newState);
else level.sendBlockUpdated(pos, state, state, 2);
```

src/main/java/com/mraof/minestuck/tileentity/machine/TotemLatheTileEntity.java
```diff
@@ -119,49 +110,30 @@ public boolean setDowel(ItemStack stack)
...
-				if(!(te instanceof ItemStackTileEntity))
-				{
-					te = new ItemStackTileEntity();
-					level.setBlockEntity(pos, te);
-				}
-				ItemStackTileEntity teItem = (ItemStackTileEntity) te;
-				teItem.setStack(stack);
-				
-				return true;
+				te = new ItemStackTileEntity();
```
kirderf1, line 119: I suspect that some of the `ItemStackTileEntity` references in this function should be `TotemLatheDowelTileEntity` instead

src/main/java/com/mraof/minestuck/tileentity/machine/TotemLatheTileEntity.java
```diff
@@ -119,49 +110,30 @@ public boolean setDowel(ItemStack stack)
...
+			level.sendBlockUpdated(getBlockPos(), getBlockState(), getBlockState(), Constants.BlockFlags.BLOCK_UPDATE);
+			return true;
 		}
 		return false;
 	}
 	
 	public ItemStack getDowel()
 	{
-		BlockPos pos = getBlockPos().above().relative(getFacing().getCounterClockWise(), 2);
+		BlockPos pos = getBlockPos().above().relative(getFacing().getCounterClockWise(), 1);
```
kirderf1, line 136: Should probably be done with `MSBlocks.TOTEM_LATHE.getDowelPos(getBlockPos(), getBlockState())` instead

src/main/java/com/mraof/minestuck/tileentity/machine/TotemLatheTileEntity.java
kirderf1: Nevermind. I realised that the first one called is for the dowel block, and the second, new line, is for the slot block. Perhaps add a comment or two to make this clearer?
Could also be nice if the first `sendBlockUpdated()` also used `Constants.BlockFlags.BLOCK_UPDATE`.

kirderf1 (commented):
The PR should be ready to be merged once the review comments have been resolved.
A little bit sad that there is no sound effect after these changes, but Dweblenod should address that if something needs to be done about it.

src/main/java/com/mraof/minestuck/client/renderer/tileentity/TotemLatheRenderer.java
amiwyn: its to reset the buffer back to its state after using it to render the dowel. maybe theres a better way to achieve that

amiwyn: i added a comment

src/main/java/com/mraof/minestuck/block/machine/TotemLatheBlock.java
```diff
@@ -38,6 +40,7 @@ public TotemLatheBlock(MachineMultiblock machine, CustomVoxelShape shape, BlockP
 		super(machine, properties);
 		this.shape = shape.createRotatedShapes();
 		this.mainPos = mainPos;
+		this.defaultBlockState().getRenderShape();
```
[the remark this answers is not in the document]
amiwyn: 😨 youre right

PR #342, opened 2022-05-27
Extra knockback when attacking

amiwyn:


Dweblenod (approved):

Dweblenod:
Will wait for Kirderf to get a chance to glance over this but I think it looks ready to go. We may extend this knockback resistance to additional underlings depending on testing feedback

src/main/java/com/mraof/minestuck/entity/AnimatedCreatureEntity.java
```diff
@@ -205,7 +216,12 @@ protected void checkAndPerformAttack(LivingEntity enemy, double distToEnemySqr)
 					this.mob.getNavigation().stop();
 				}
 				entity.startAttack();
-				// TODO temporary knockback resistance while attacking
+				
+				if (this.entity.knockbackResistWhileAttacking >= 0.01) 
+				{
+					attributes.put(Attributes.KNOCKBACK_RESISTANCE, new AttributeModifier("attack.knockback", 1.0, AttributeModifier.Operation.ADDITION));
+					this.entity.getAttributes().addTransientAttributeModifiers(attributes);
+				}
```
kirderf1, line 224: There are some things that are different from how I first imagined that this would be done:
- That the attribute modifier would have a static uuid so that the modifier could afterwards be identified and removed or replaced. You're constructing the modifier with a random uuid.
- That the attribute modifier would be removed when the attack has finished. I'm not seeing a code change that does this.
- That you would add just the modifier directly. You appear to add the modifier by first adding it to some private (and static!) map, and then sending along the map, but you never do anything else with the map? (Including clearing it out at any point)

I'm pretty sure something needs to change, but I'd like to know more about the intent behind the code before suggesting any specific change. How is this feature meant to work? How is the attribute modifier meant to be managed? How is the lifecycle of the modifier meant to look like? What is the purpose of the attributes multimap?

amiwyn: i accidentally broke it while merging (tab vs spaces makes it really difficult for git)

amiwyn: >That the attribute modifier would have a static uuid so that the modifier could afterwards be identified and removed or replaced.

I couldnt find anything that can do that without passing on a map

kirderf1: `new AttributeModifier(UUID, String, double, AttributeModifier.Operation)`
`entity.getAttributes().getInstance(Attributes.KNOCKBACK_RESISTANCE).removeModifier(uuid)`
`entity.getAttributes().getInstance(Attributes.KNOCKBACK_RESISTANCE).removeModifier(modifier)`
`entity.getAttributes().getInstance(Attributes.KNOCKBACK_RESISTANCE).addTransientModifier(modifier)`

kirderf1 (approved):
The attribute modifier could still have a fixed uuid (like for example `EndermanEntity.SPEED_MODIFIER_ATTACKING`, `WitchEntity.SPEED_MODIFIER_DRINKING`, `ZombieEntity.SPEED_MODIFIER_BABY`).
But because the modifier is transient, it shouldn't last long enough to matter in practice.

PR #343, opened 2022-05-27
Entity head tracking

amiwyn:


src/main/java/com/mraof/minestuck/client/model/AnimatedModel.java
```diff
@@ -0,0 +1,20 @@
...
+public abstract class AnimatedModel<T extends IAnimatable> extends AnimatedGeoModel<T> {
+	@Override
+	public void setLivingAnimations(T entity, Integer uniqueID, @Nullable AnimationEvent predicate) {
+		super.setLivingAnimations(entity, uniqueID, predicate);
+		IBone head = this.getAnimationProcessor().getBone("head");
+		EntityModelData extraData = (EntityModelData) predicate.getExtraDataOfType(EntityModelData.class).get(0);
+		head.setRotationX(extraData.headPitch * ((float) Math.PI / 180F));
+		head.setRotationY(extraData.netHeadYaw * ((float) Math.PI / 180F));
+	}
+}
```
Dweblenod, line 20: While its clear with context, or by looking through the few lines of code here, what this does I believe some javadoc commenting on lines 10/11 would be appropriate so that its clear just from hovering over the class

src/main/java/com/mraof/minestuck/client/model/UnderlingModel.java
```diff
@@ -3,11 +3,10 @@
 import com.mraof.minestuck.Minestuck;
 import com.mraof.minestuck.entity.underling.UnderlingEntity;
 import net.minecraft.util.ResourceLocation;
-import software.bernie.geckolib3.model.AnimatedGeoModel;
 
 import java.util.Objects;
 
-public class UnderlingModel<T extends UnderlingEntity> extends AnimatedGeoModel<T>
+public class UnderlingModel<T extends UnderlingEntity> extends AnimatedModel<T>
```
Dweblenod, line 9: While it would be most applicable elsewhere, I think it would be easy to forget if not preserved in code. The ogres attack animation involves its head bone, which causes it to awkwardly look directly forward or rotate when attacking

Dweblenod: This is an animation fix however, so having seen its implementation elsewhere it looks all good to move forward with

Dweblenod: Riotmode has been made aware of the issue and I will put a note in Geckolib that it still needs fixed to avoid requiring commenting here

PR #346, opened 2022-06-02
Animated entity cleanup

kirderf1:
Does various refactors and other changes related to features of `AnimatedCreatureEntity`, splitting part of it off into a new abstract class `AttackingAnimatedEntity`.
The `SimpleTexturedEntityRenderer` has also been changed to remove the need for the `SimpleTexturedEntity` class.


kirderf1:
I also thought about moving functionality from `AttackingAnimatedEntity` into the attack goal to the point that the delayed attack can be added to an entity not through inheritance, but through implementing an interface and registering the `DelayedAttackGoal`.

I also thought about splitting up the `Actions` enum into one type used by consorts with `NONE`, `TALK`, `PANIC`, and one type used by attacking mobs with `NONE`, `ATTACK`, `ATTACK_RECOVERY`.

But I didn't find it clear enough how best to do these things, so I ended up skipping them. I might try these things again in a future PR if I feel like it.

kirderf1:
That's not possible. Enums are final and cannot be extended, as they represent a fixed-size of variants.
Instead, I was considering splitting the enum class into two separate enum classes, and let either one be used in `AnimatedCreatureEntity` though generics. (`class AnimatedCreatureEntity<T extends Enum<T>>`, `class ConsortEntity extends AnimatedCreatureEntity<ConsortAction>`)

src/main/java/com/mraof/minestuck/entity/AnimatedCreatureEntity.java
```diff
@@ -76,74 +36,14 @@ public void tick()
...
-				this.entityData.set(CURRENT_ACTION, Actions.NONE.ordinal());
-			}
+			if(animationTicks == 0)
+				this.endTimedAction(this.getCurrentAction());
 		}
 	}
 	
-	private void performAttack()
+	protected void endTimedAction(Actions action)
 	{
```
Dweblenod, line 45: title is still open to different interpretations from my view, can you add javadoc?

src/main/java/com/mraof/minestuck/entity/underling/UnderlingEntity.java
```diff
@@ -38,15 +38,18 @@
...
 import java.util.*;
 
-public abstract class UnderlingEntity extends AnimatedCreatureEntity implements IMob, IAnimatable
+public abstract class UnderlingEntity extends AttackingAnimatedEntity implements IMob, IAnimatable
 {
 	public static final UUID GRIST_MODIFIER_ID = UUID.fromString("08B6DEFC-E3F4-11EA-87D0-0242AC130003");
 	private static final DataParameter<String> GRIST_TYPE = EntityDataManager.defineId(UnderlingEntity.class, DataSerializers.STRING);
+	
+	private final AnimationFactory factory = new AnimationFactory(this);
 	protected final EntityListFilter attackEntitySelector = new EntityListFilter(new ArrayList<>());    //TODO this filter isn't being saved. F1X PLZ
```
Dweblenod, line 53: I know you did not create this, but do you have any suggestions on how to fix?

src/main/java/com/mraof/minestuck/entity/AttackingAnimatedEntity.java
```diff
@@ -0,0 +1,116 @@
...
+	protected void onAttackStart()
+	{
+	}
+	
+	/**
+	 * Is called when an attack starts. Can be extended to remove effects once an attack has ended.
+	 */
+	protected void onAttackEnd()
+	{
+	}
```
Dweblenod, line 89: Can you foresee any scenario in which modifiers set during these functions could accidentally get applied multiple times without having been removed? For example if the game is quit during an attack

Dweblenod, line 89: would this be an appropriate location to apply effects or additional modifications to the targetted mob if they are hit? As a random example, if an underling attacks and successfully hits a player and causes the player to catch fire

Dweblenod (commented):
I like the increased flexibility this PR offers! I will submit a secondary review set if I notice any issues while testing it out in game

src/main/java/com/mraof/minestuck/entity/underling/BasiliskEntity.java
```diff
@@ -98,6 +99,13 @@ public void die(DamageSource cause)
 		}
 	}
 	
+	@Override
+	protected void onAttackStart()
+	{
+		// Will stop the entity while performing its attack animation
+		this.getNavigation().stop();
+	}
```
Dweblenod, line 107: This is not enough currently to prevent any movement before the end of the animation (which also then can mess with animations), modifying the movement speed via attributes may be required

Dweblenod: This issue extends to all underlings and consorts, although not all have issues with their animations because of it

src/main/java/com/mraof/minestuck/entity/underling/UnderlingEntity.java
kirderf1: To fix it would involve creating some serialisation functions on the filter and use them in the entitys serialisation functions.
However, I'd like to have a closer look at the filter and determine if that functionality should simply be removed instead, and remove the serialisation need that way instead.

src/main/java/com/mraof/minestuck/entity/AttackingAnimatedEntity.java
kirderf1: For your example, it depends on the type of modifier. If it's a "permanent" modifier, it will be serialized, and thus will be remain when the entity is unloaded and reloaded. If it's a transient modifier, it will not be serialized', and thus be gone when the entity is unloaded and reloaded.

However, a possible scenario is that a delayed attack starts, and something else calls `setAction()` and changes the action before it has completed. There is no such call anywhere right now, but this shows a problem with the current action system, where actions can technically be interrupted, but the interruption is not handled.

One option is to strictly disallow action interruption in `setAction()`, or handle interruption such that the modifier gets properly removed. I'm leaning towards a third alternative that prevents outside interruptions by moving the attack timer into the ai goal. Doing so would also mean that the action state for attacking entites could be simplified to a "isAttacking" property.

kirderf1: No. This function call happens for an attack end, which does not neccesarily correspond to an attack hit. There eould need to be a different overridable function for that.

src/main/java/com/mraof/minestuck/entity/underling/BasiliskEntity.java
kirderf1: That would've been a problem before as well, but I could look at it and see if I come up with something that works better.

Dweblenod: It has been a problem with the changes during geckolib, not necessarily something that needs to be completed during this PR but if you have any suggestions we can resolve this conversation

src/main/java/com/mraof/minestuck/entity/AttackingAnimatedEntity.java
Dweblenod: Do you believe it is likely to be a problem or something we could hold off on?

src/main/java/com/mraof/minestuck/entity/underling/UnderlingEntity.java
Dweblenod: If you do not think it is a major issue I am also fine with leaving it

src/main/java/com/mraof/minestuck/entity/AttackingAnimatedEntity.java
kirderf1: As said, it's not a problem right now, but it might become a problem if we start getting more types of actions introduced. I think we can hold off on this together with my ideas from https://github.com/mraof/Minestuck/pull/346#issuecomment-1146206387

src/main/java/com/mraof/minestuck/entity/underling/UnderlingEntity.java
kirderf1: The filter tracks which entities that the underling considers to be enemies. If a non-underling entity attacks the underling, it gets added to the filter, so that the specific underling would keep attacking entities of that type. Because the filter is not serialized, the list of enemies gets cleared on unload-reload. (The player entity type is always added back, so it'll still consider players to be enemies)

This is a minor issue, and not a major issue.

src/main/java/com/mraof/minestuck/entity/underling/BasiliskEntity.java
Dweblenod: If you add a //TODO I would consider this resolved for now

PR #351, opened 2022-06-21
Frog temple improvements

Dweblenod:
New content/changes:
- Blank disks: can now be crafted from 4 raw cruxite and a gold nugget
- Completed sburb code: version of sburb code that contains all necessary info
- Sburb Code: now is reworked to store hieroglpyhs it has been right clicked to, can be right clicked to a computer in order to simultaneously upload any yet stored code from the item to the computer and download any yet stored code from the computer to the item. An actual nucleotide sequence is shown when opening up the item(with blank sections for yet recorded code)
- Disk Burner computer program: new computer program installed by default to the computer, allows server and client disks to be created once the computer has all necessary genetic code from the sburb code items and has been fed at least 1 blank disk(1:1 blank disk to program disk)
- Server/Client disks: no longer can be crafted from a blank disk and sburb code, requires DiskBurner program to make
- Cartographers now sell maps to yet to be generated frog temples, wandering traders now sell blank disks(approximately 25% of the time without other mods/datapacks installed), leatherworkers now sell crumpled hats, masons now sell carving tools/stone slabs and can duplicate frog hieroglyphs for a price, weaponsmiths now sell enchanted lucerne hammers, fishermen now purchase cicadas and grasshoppers, librarians can now purchase various Minestuck added books(which change whether the villager is in the Medium or not)
- Lotus flower now uses a loot table instead of hard-coded drops(drops themselves have not changed), allowing for easier modification by addons or datapack creators
- Increased blank disk drops in frog temple and vanilla chests, added blank disk to more structure loot chests in the overworld
- Ladders in the frog temple have been replaced with a new block called green stone brick embedded ladder which work the same as normal ladders

New form of collecting code and server/client disks:
1. Player activates the lotus flower in order to receive the computer parts(as before) and sburb code(now has more complexity, and when spawned in this way has a special piece of code needed to continue).
2. Player right clicks the sburb code to all hieroglyph blocks present in the hieroglyph block tag(GREEN_HIEROGLYPHS) in order to record their aesthetically corresponding nucleotide sequences. Book and Quills can still be used to create sburb code but when created lack the paradox code supplied by the lotus flower copy
3. Player right clicks the sburb code with recorded hieroglyphs to an active computer. 2 blank disks need to be right clicked to the computer as well to be burned. Once the computer has had all blocks in the hieroglyph block tag uploaded from the sburb code items and the paradox code, the Disk Burner program can be used to create a copy of the server disk and client disk each from a blank disk using the genetic code
4. From here the server/client disks are used as usual. The various pieces of genetic code do not have to be supplied to the computer all at once as on right click the computer and sburb code item share missing pieces they possess with each other


Dweblenod:
Still requires bug fixing but the actual content is complete and functional. Seems smart to wait until 1.18 is live before merging this, so it still is in the back log despite no longer being a draft

src/main/java/com/mraof/minestuck/util/MSTags.java
```diff
@@ -158,4 +166,39 @@ private static TagKey<Biome> tag(String name)
...
+				return true;
+		
+		return false;
+	}
+	
+	public static List<Block> getBlocksFromTag(TagKey<Block> itemTag)
+	{
+		List<Block> blockListFromTag = new ArrayList<>();
+		Registry.BLOCK.getTagOrEmpty(itemTag).forEach(blockHolder -> blockListFromTag.add(blockHolder.value()));
+		return blockListFromTag;
```
kirderf1, line 195:
```suggestion
		return ForgeRegistries.BLOCKS.tags().getTag(blockTag).stream().toList();
```

src/main/java/com/mraof/minestuck/util/MSTags.java
```diff
@@ -158,4 +166,39 @@ private static TagKey<Biome> tag(String name)
...
+	 * Used when getBlocksFromTag has already been called to avoid redundancy
+	 */
+	public static boolean blockExistsInList(List<Block> blockList, Block block)
+	{
+		for(Block blockIterate : blockList)
+			if(blockIterate == block)
+				return true;
+		
+		return false;
+	}
```
kirderf1, line 189: Replace usage of this with `List.contains()`.

src/main/java/com/mraof/minestuck/util/MSTags.java
```diff
@@ -158,4 +166,39 @@ private static TagKey<Biome> tag(String name)
...
 	}
+	
+	public static boolean blockExistsInTag(TagKey<Block> blockTag, Block block)
+	{
+		for(Block blockIterate : getBlocksFromTag(blockTag))
+			if(blockIterate == block)
+				return true;
+		
+		return false;
+	}
```
kirderf1, line 177: Replace usage of this with `BlockState.is()`.

src/main/java/com/mraof/minestuck/util/MSTags.java
```diff
@@ -158,4 +166,39 @@ private static TagKey<Biome> tag(String name)
...
+		List<Block> blockListFromTag = new ArrayList<>();
+		Registry.BLOCK.getTagOrEmpty(itemTag).forEach(blockHolder -> blockListFromTag.add(blockHolder.value()));
+		return blockListFromTag;
+	}
+	
+	public static List<ItemStack> getItemStacksFromTag(TagKey<Item> itemTag)
+	{
+		List<ItemStack> stackListFromTag = new ArrayList<>();
+		Registry.ITEM.getTagOrEmpty(itemTag).forEach(itemHolder -> stackListFromTag.add(new ItemStack(itemHolder)));
+		return stackListFromTag;
```
kirderf1, line 202:
```suggestion
		return ForgeRegistries.ITEMS.tags().getTag(itemTag).stream().map(ItemStack::new).toList();
```

src/main/java/com/mraof/minestuck/network/computer/BurnDiskPacket.java
```diff
@@ -0,0 +1,76 @@
...
+		BlockPos bePos = buffer.readBlockPos();
+		int programId = buffer.readInt();
+		return new BurnDiskPacket(bePos, programId);
+	}
+	
+	@Override
+	public void execute(ServerPlayer player)
+	{
+		Level level = player.level;
+		BlockEntity blockEntity = level.getBlockEntity(bePos);
```
kirderf1, line 51: There needs to be some safety against triggering chunk loading, and against interacting with computers far away. (See use of `ComputerBlockEntity.forNetworkIfPresent()` in other computer-related packages)

src/main/java/com/mraof/minestuck/world/gen/structure/FrogTemplePiece.java
```diff
@@ -264,38 +263,20 @@ private Selector()
...
+					MSBlocks.GREEN_STONE_BRICK_IGUANA_LEFT.get(), MSBlocks.GREEN_STONE_BRICK_IGUANA_LEFT.get(), MSBlocks.GREEN_STONE_BRICK_IGUANA_LEFT.get(), MSBlocks.GREEN_STONE_BRICK_IGUANA_LEFT.get(),
+					MSBlocks.GREEN_STONE_BRICK_IGUANA_RIGHT.get(), MSBlocks.GREEN_STONE_BRICK_IGUANA_RIGHT.get(), MSBlocks.GREEN_STONE_BRICK_IGUANA_RIGHT.get(), MSBlocks.GREEN_STONE_BRICK_IGUANA_RIGHT.get(),
+					MSBlocks.GREEN_STONE_BRICK_NAK_LEFT.get(), MSBlocks.GREEN_STONE_BRICK_NAK_LEFT.get(), MSBlocks.GREEN_STONE_BRICK_NAK_LEFT.get(), MSBlocks.GREEN_STONE_BRICK_NAK_LEFT.get(),
+					MSBlocks.GREEN_STONE_BRICK_NAK_RIGHT.get(), MSBlocks.GREEN_STONE_BRICK_NAK_RIGHT.get(), MSBlocks.GREEN_STONE_BRICK_NAK_RIGHT.get(), MSBlocks.GREEN_STONE_BRICK_NAK_RIGHT.get(),
+					MSBlocks.GREEN_STONE_BRICK_SALAMANDER_LEFT.get(), MSBlocks.GREEN_STONE_BRICK_SALAMANDER_LEFT.get(), MSBlocks.GREEN_STONE_BRICK_SALAMANDER_LEFT.get(), MSBlocks.GREEN_STONE_BRICK_SALAMANDER_LEFT.get(),
+					MSBlocks.GREEN_STONE_BRICK_SALAMANDER_RIGHT.get(), MSBlocks.GREEN_STONE_BRICK_SALAMANDER_RIGHT.get(), MSBlocks.GREEN_STONE_BRICK_SALAMANDER_RIGHT.get(), MSBlocks.GREEN_STONE_BRICK_SALAMANDER_RIGHT.get()
+			};
```
kirderf1, line 277: Could you turn this into a weighted list instead? See `WeightedRandom` and `WeightedEntry`.

kirderf1 (changes requested):

src/main/java/com/mraof/minestuck/world/gen/structure/FrogTemplePiece.java
Dweblenod: It works in game but please check to make sure I used a rational approach

src/main/java/com/mraof/minestuck/util/MSTags.java
Dweblenod: I may be confused, but I do not see that method in the BlockState class

src/main/java/com/mraof/minestuck/network/computer/BurnDiskPacket.java
Dweblenod: I believe I fixed it

src/main/java/com/mraof/minestuck/computer/ProgramData.java
```diff
@@ -64,11 +64,12 @@ public static int getProgramID(ItemStack item)
 		return -2;
 	}
 	
-	@Nonnull
 	public static ItemStack getItem(int id)
 	{
 		if(id == -1)
 			return new ItemStack(Items.MUSIC_DISC_11);
+		if(id == 2)
+			return null;
```
kirderf1, line 72: Instead of making this itemstack nullable, can you use `ItemStack.EMPTY`?

src/main/java/com/mraof/minestuck/data/loot_table/MSChestLootTables.java
```diff
@@ -64,7 +69,7 @@ public void accept(BiConsumer<ResourceLocation, LootTable.Builder> lootProcessor
 						.add(LootItem.lootTableItem(Items.DIAMOND).setWeight(1).setQuality(3).apply(SetItemCountFunction.setCount(UniformGenerator.between(1, 3))))
 						.add(LootItem.lootTableItem(Items.PUMPKIN).setWeight(3).setQuality(-1).apply(SetItemCountFunction.setCount(UniformGenerator.between(1, 5))))
 						.add(LootItem.lootTableItem(Items.WRITABLE_BOOK).setWeight(2).setQuality(1).apply(SetItemCountFunction.setCount(UniformGenerator.between(0, 1))))
-						.add(LootItem.lootTableItem(MSItems.BLANK_DISK.get()).setWeight(5).setQuality(3).apply(SetItemCountFunction.setCount(UniformGenerator.between(1, 2))))
+						.add(LootItem.lootTableItem(MSItems.BLANK_DISK.get()).setWeight(5).setQuality(3).apply(SetItemCountFunction.setCount(UniformGenerator.between(0, 2))))
```
kirderf1, line 72: Potentially setting the item count to 0 seems a bit odd. Would you consider tweaking the loot entry weight instead?

src/main/java/com/mraof/minestuck/entity/LotusFlowerEntity.java
```diff
@@ -182,16 +192,27 @@ protected void updateAndSendAnimation(Animation animation)
...
-		Vec3 posVec = this.position();
-		
-		ItemEntity unpoweredComputerItemEntity = new ItemEntity(level, posVec.x(), posVec.y() + 1D, posVec.z(), new ItemStack(MSItems.COMPUTER_PARTS.get(), 1));
-		level.addFreshEntity(unpoweredComputerItemEntity);
-		
-		ItemEntity sburbCodeItemEntity = new ItemEntity(level, posVec.x(), posVec.y() + 1D, posVec.z(), new ItemStack(MSItems.SBURB_CODE.get(), 1));
-		level.addFreshEntity(sburbCodeItemEntity);
+		if(!level.isClientSide)
+		{
+			ServerLevel serverWorld = (ServerLevel) level;
```
kirderf1, line 202: Rename `world` -> `level`

src/main/java/com/mraof/minestuck/entity/LotusFlowerEntity.java
```diff
@@ -182,16 +192,27 @@ protected void updateAndSendAnimation(Animation animation)
...
+			
+			LootTable lootTable = serverWorld.getServer().getLootTables().get(MSLootTables.LOTUS_FLOWER_DEFAULT);
+			List<ItemStack> loot = lootTable.getRandomItems(new LootContext.Builder(serverWorld).create(LootContextParamSets.EMPTY));
+			if(loot.isEmpty())
+				LOGGER.warn("Tried to generate loot for Lotus Flower, but no items were generated!");
+			
+			for(ItemStack itemStack : loot)
+			{
+				if(itemStack.getItem() == MSItems.SBURB_CODE.get())
+					SburbCodeItem.setParadoxInfo(itemStack, true);
```
kirderf1, line 212: This seems better suited for a loot function

src/main/java/com/mraof/minestuck/item/MSItems.java
```diff
@@ -349,7 +349,8 @@
 	public static final RegistryObject<Item> CRUXITE_APPLE = REGISTER.register("cruxite_apple", () -> new CruxiteAppleItem(new Item.Properties().stacksTo(1).tab(MSItemGroup.MAIN).rarity(Rarity.UNCOMMON))); //TODO have to fix Cruxite artifact classes
 	public static final RegistryObject<Item> CRUXITE_POTION = REGISTER.register("cruxite_potion", () -> new CruxitePotionItem(new Item.Properties().stacksTo(1).tab(MSItemGroup.MAIN).rarity(Rarity.UNCOMMON)));
 	
-	public static final RegistryObject<Item> SBURB_CODE = REGISTER.register("sburb_code", () -> new Item(new Item.Properties().tab(MSItemGroup.MAIN)));
+	public static final RegistryObject<Item> SBURB_CODE = REGISTER.register("sburb_code", () -> new SburbCodeItem(new Item.Properties().stacksTo(1).tab(MSItemGroup.MAIN)));
+	public static final RegistryObject<Item> COMPLETED_SBURB_CODE = REGISTER.register("completed_sburb_code", () -> new ReadableSburbCodeItem(new Item.Properties().tab(MSItemGroup.MAIN).rarity(Rarity.UNCOMMON)));
```
kirderf1, line 353: I suggest making the completed sburb code stack to 1 as well

src/main/java/com/mraof/minestuck/item/ReadableSburbCodeItem.java
```diff
@@ -0,0 +1,161 @@
...
+		
+		return super.use(levelIn, playerIn, handIn);
+	}
+	
+	/**
+	 * Uses the nbt of the itemstack and attempts to return a list of blocks based on the string of their registry name
+	 */
+	public static List<Block> getRecordedBlocks(ItemStack stack)
+	{
+		if(stack.getItem() == MSItems.COMPLETED_SBURB_CODE.get())
```
kirderf1, line 61: Use `stack.is(item)`

src/main/java/com/mraof/minestuck/world/gen/structure/FrogTemplePiece.java
```diff
@@ -264,38 +265,21 @@ private Selector()
...
+			weightedBlockList.add(WeightedEntry.wrap(MSBlocks.GREEN_STONE_BRICK_FROG.get(), 1)); //only one frog hieroglyph in order to have a component more difficult to find
+			weightedBlockList.add(WeightedEntry.wrap(MSBlocks.GREEN_STONE_BRICK_TURTLE.get(), 36));
+			weightedBlockList.add(WeightedEntry.wrap(MSBlocks.GREEN_STONE_BRICK_SKAIA.get(), 36));
+			weightedBlockList.add(WeightedEntry.wrap(MSBlocks.GREEN_STONE_BRICK_LOTUS.get(), 36));
+			weightedBlockList.add(WeightedEntry.wrap(MSBlocks.GREEN_STONE_BRICK_IGUANA_LEFT.get(), 16));
+			weightedBlockList.add(WeightedEntry.wrap(MSBlocks.GREEN_STONE_BRICK_IGUANA_RIGHT.get(), 16));
+			weightedBlockList.add(WeightedEntry.wrap(MSBlocks.GREEN_STONE_BRICK_NAK_LEFT.get(), 16));
+			weightedBlockList.add(WeightedEntry.wrap(MSBlocks.GREEN_STONE_BRICK_NAK_RIGHT.get(), 16));
+			weightedBlockList.add(WeightedEntry.wrap(MSBlocks.GREEN_STONE_BRICK_SALAMANDER_LEFT.get(), 16));
+			weightedBlockList.add(WeightedEntry.wrap(MSBlocks.GREEN_STONE_BRICK_SALAMANDER_RIGHT.get(), 16));
```
kirderf1, line 279: Since this function is being called plenty of times, I'd suggest building the list in the constructor instead. That way it only happens once per structure, instead of once per block.

src/main/java/com/mraof/minestuck/world/gen/structure/FrogTemplePiece.java
```diff
@@ -264,38 +265,21 @@ private Selector()
...
+			
+			Optional<WeightedEntry.Wrapper<Block>> optional = WeightedRandom.getRandomItem(rand, weightedBlockList);
+			optional.ifPresent(blockWrapper -> this.next = blockWrapper.getData().defaultBlockState()); //sets the next blockstate to an element of the weighted list as long as the optional is present
```
kirderf1, line 282: The optional should only be empty if the list is empty. And since the list should very much not be empty, I'd suggest just using `optional.orElseThrow()` instead.

src/main/java/com/mraof/minestuck/world/gen/structure/FrogTemplePiece.java
kirderf1: If you do so, you could also store the total weight from `WeightedRandom.getTotalWeight()` and pass it into `WeightedRandom.getRandomItem()` as well, since that'd also be the same for every block.

src/main/java/com/mraof/minestuck/util/BlockHitResultUtil.java
```diff
@@ -0,0 +1,43 @@
...
+import net.minecraft.world.phys.BlockHitResult;
+import net.minecraft.world.phys.Vec3;
+import net.minecraftforge.common.ForgeMod;
+
+public class BlockHitResultUtil
+{
+	/**
+	 * Based on the Item class function of the same name, without the FluidMode context
+	 */
+	public static BlockHitResult getPlayerPOVHitResult(Level level, Player playerEntity)
```
kirderf1, line 16: Why is this needed over the version in `Item`?

src/main/java/com/mraof/minestuck/network/computer/BurnDiskPacket.java
```diff
@@ -0,0 +1,78 @@
...
+		BlockPos bePos = buffer.readBlockPos();
+		int programId = buffer.readInt();
+		return new BurnDiskPacket(bePos, programId);
+	}
+	
+	@Override
+	public void execute(ServerPlayer player)
+	{
+		Level level = player.level;
+		if(level.isAreaLoaded(bePos, 0))    //TODO also check distance to the computer pos (together with a continual check clientside)
```
kirderf1, line 51: I'd suggest using `ComputerBlockEntity.forNetworkIfPresent()` here

src/main/java/com/mraof/minestuck/item/SburbCodeItem.java
```diff
@@ -0,0 +1,175 @@
...
+import net.minecraft.world.phys.BlockHitResult;
+import net.minecraft.world.phys.HitResult;
+import org.jetbrains.annotations.Nullable;
+
+import java.util.List;
+
+/**
+ * Used for the Sburb Code item, extends ReadableSburbCodeItem which is used by Completed Sburb Code. Allows players to store a list of blocks from the GREEN_HIEROGLPYH block tag that in game reflects the collection of genetic code for the creation of sburb
+ */
+public class SburbCodeItem extends ReadableSburbCodeItem
```
kirderf1, line 32: Odd naming convention with `SburbCodeItem` being a more specific version of `ReadableSburbCodeItem` instead of the other way around. I suggest renaming this class to `IncompleteSburbCodeItem` or `FillableSburbCodeItem`.

kirderf1 (changes requested):
Reviewing is incomplete, but should still be a good bit towards finished.

src/main/java/com/mraof/minestuck/block/machine/ComputerBlock.java
```diff
@@ -108,8 +115,20 @@ private void turnOn(BlockState state, Level level, BlockPos pos, Player player,
 	
 	private boolean insertDisk(ComputerBlockEntity blockEntity, BlockState state, Level level, BlockPos pos, Player player, InteractionHand handIn)
 	{
-		int id = ProgramData.getProgramID(player.getItemInHand(handIn));
-		if(id != -2 && !blockEntity.hasProgram(id) && blockEntity.installedPrograms.size() < 2 && !blockEntity.hasProgram(-1))
+		ItemStack stackInHand = player.getItemInHand(handIn);
+		int id = ProgramData.getProgramID(stackInHand);
+		
+		if(stackInHand.getItem() == MSItems.BLANK_DISK.get())
```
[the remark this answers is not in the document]
Dweblenod: Done now

kirderf1:
It seems like a reasonable goal to get a 1.18 release with this one before we move on to 1.19.

src/main/java/com/mraof/minestuck/util/BlockHitResultUtil.java
Dweblenod: Its not entirely necessary, all it does is skip the fluid check. It seemed useful to have it, along with the other two functions, be all kept in one place

src/main/java/com/mraof/minestuck/network/computer/BurnDiskPacket.java
Dweblenod: Im not sure how to handle the creation of the disk item and other context specific to the disk burning using that function

src/main/java/com/mraof/minestuck/world/gen/structure/FrogTemplePiece.java
Dweblenod: I got it working to what I believe is what you recommended but as before please check to make sure it was an efficient approach

src/main/java/com/mraof/minestuck/entity/LotusFlowerEntity.java
Dweblenod: Now using a function

src/main/java/com/mraof/minestuck/item/SburbCodeItem.java
Dweblenod: Changed it to IncompleteSburbCodeItem, am considering simplifying the parent class to just SburbCodeItem for simplicity

src/main/java/com/mraof/minestuck/util/MSTags.java
Dweblenod: Considering how short that version is, and the fact that blockExistsInTag was unused, I just went and deleted it

src/main/java/com/mraof/minestuck/client/gui/ReadableSburbCodeScreen.java
```diff
@@ -0,0 +1,293 @@
...
+import java.nio.charset.StandardCharsets;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.stream.Collectors;
+
+/**
+ * Code is modified components of, or inspired by, WinGameScreen/ReadBookScreen/StoneTabletScreen.
+ * It reads the nucleotide sequence stored in rana_temporaria_sec22b.txt and associates a number of lines of the text to each hieroglyph from the block tag with one extra pseudo-hieroglyph in the form of the paradox code
+ */
+public class ReadableSburbCodeScreen extends Screen
```
kirderf1, line 32: I suspect that more fields here should be private. What is your opinion?

src/main/java/com/mraof/minestuck/computer/DiskBurner.java
```diff
@@ -0,0 +1,50 @@
...
+import com.mraof.minestuck.util.MSTags;
+
+import java.util.ArrayList;
+
+public class DiskBurner extends ButtonListProgram
+{
+	public static final String NAME = "minestuck.disk_burner_program";
+	public static final String BURN_SERVER_DISK = "minestuck.burn_server_disk";
+	public static final String BURN_CLIENT_DISK = "minestuck.burn_client_disk";
+	public static final String CHOOSE = "minestuck.choose";
```
kirderf1, line 15: Perhaps add to the unlocalized name slightly to make it specific to the disk burner program

src/main/java/com/mraof/minestuck/computer/DiskBurner.java
```diff
@@ -0,0 +1,50 @@
...
+	
+	@Override
+	public ArrayList<UnlocalizedString> getStringList(ComputerBlockEntity be)
+	{
+		ArrayList<UnlocalizedString> list = new ArrayList<>();
+		list.add(new UnlocalizedString(CHOOSE));
+		
+		//TODO remove the buttons during an earlier check, it takes two result-less clicks to remove them
```
kirderf1, line 23: Could you explain what this todo is about?

src/main/java/com/mraof/minestuck/block/machine/ComputerBlock.java
```diff
@@ -123,14 +142,72 @@ private boolean insertDisk(ComputerBlockEntity blockEntity, BlockState state, Le
...
-		} else return false;
+		}
+		
+		return false;
+	}
+	
+	private boolean inputCode(ComputerBlockEntity blockEntity, BlockState state, Level levelIn, BlockPos pos, Player player, InteractionHand handIn)
+	{
+		ItemStack heldStack = player.getItemInHand(handIn);
+		if(heldStack.getItem() instanceof ReadableSburbCodeItem)
```
kirderf1, line 153: Can you replace these instanceof checks with appropriate `heldStack.is(item)` checks instead?

src/main/java/com/mraof/minestuck/block/machine/ComputerBlock.java
```diff
@@ -123,14 +142,72 @@ private boolean insertDisk(ComputerBlockEntity blockEntity, BlockState state, Le
...
+		return false;
 	}
 	
 	@Nullable
 	@Override
 	public BlockEntity newBlockEntity(BlockPos pos, BlockState state)
 	{
-		return state.getValue(STATE) != State.OFF ? new ComputerBlockEntity(pos, state) : null;
+		ComputerBlockEntity be = new ComputerBlockEntity(pos, state);
+		be.installedPrograms.put(2, true); //the program disk burner has no associated item and should always exist on the computer
```
kirderf1, line 209: Move this to the block entity constructor

src/main/java/com/mraof/minestuck/data/loot_table/MSMiscLootTables.java
```diff
@@ -32,5 +35,22 @@ public void accept(BiConsumer<ResourceLocation, LootTable.Builder> lootProcessor
...
+				.withPool(LootPool.lootPool().name("computer").setRolls(ConstantValue.exactly(1))
+						.add(LootItem.lootTableItem(MSItems.COMPUTER_PARTS.get()).setWeight(10).setQuality(3)))
+				.withPool(LootPool.lootPool().name("code").setRolls(ConstantValue.exactly(1))
+						.add(LootItem.lootTableItem(MSItems.SBURB_CODE.get()).setWeight(10).setQuality(3)).apply(SetNbtFunction.setTag(giveParadoxFragment())))
+		);
+	}
+	
+	private static CompoundTag giveParadoxFragment()
+	{
+		ItemStack stack = new ItemStack(MSItems.SBURB_CODE.get());
```
kirderf1, line 49: This item stack doesn't seem to serve a purpose.

src/main/java/com/mraof/minestuck/block/machine/ComputerBlock.java
```diff
@@ -161,18 +238,37 @@ private void dropItems(Level level, int x, int y, int z, BlockState state)
...
 		if(state.getValue(STATE) == State.BROKEN)
 		{
 			float rx = rand.nextFloat() * 0.8F + 0.1F;
 			float ry = rand.nextFloat() * 0.8F + 0.1F;
 			float rz = rand.nextFloat() * 0.8F + 0.1F;
-			ItemEntity entityItem = new ItemEntity(level, x + rx, y + ry, z + rz, ProgramData.getItem(-1));
-			entityItem.setDeltaMovement(rand.nextGaussian() * factor, rand.nextGaussian() * factor + 0.2F, rand.nextGaussian() * factor);
-			level.addFreshEntity(entityItem);
+			ItemStack diskStack = ProgramData.getItem(-1);
+			if(diskStack != null)
```
kirderf1, line 266: `ProgramData.getItem()` shouldn't be returning null, so this check should not be needed.

src/main/java/com/mraof/minestuck/block/machine/ComputerBlock.java
```diff
@@ -161,18 +238,37 @@ private void dropItems(Level level, int x, int y, int z, BlockState state)
 			float rx = rand.nextFloat() * 0.8F + 0.1F;
 			float ry = rand.nextFloat() * 0.8F + 0.1F;
 			float rz = rand.nextFloat() * 0.8F + 0.1F;
-			ItemEntity entityItem = new ItemEntity(level, x + rx, y + ry, z + rz, ProgramData.getItem(program));
+			ItemStack diskStack = ProgramData.getItem(program);
+			if(diskStack != null && !diskStack.isEmpty())
```
kirderf1, line 242: `ProgramData.getItem()` shouldn't be returning null, so the null check here should not be needed.

src/main/java/com/mraof/minestuck/item/IncompleteSburbCodeItem.java
```diff
@@ -0,0 +1,175 @@
...
+import net.minecraft.world.level.block.Block;
+import net.minecraft.world.level.block.state.BlockState;
+import net.minecraft.world.phys.BlockHitResult;
+import net.minecraft.world.phys.HitResult;
+import org.jetbrains.annotations.Nullable;
+
+import java.util.List;
+
+/**
+ * Used for the Sburb Code item, extends ReadableSburbCodeItem which is used by Completed Sburb Code. Allows players to store a list of blocks from the GREEN_HIEROGLPYH block tag that in game reflects the collection of genetic code for the creation of sburb
```
kirderf1, line 30: Add line break to make this comment more readable

src/main/java/com/mraof/minestuck/item/ReadableSburbCodeItem.java
```diff
@@ -0,0 +1,161 @@
...
+ */
+public class ReadableSburbCodeItem extends Item
+{
+	public ReadableSburbCodeItem(Item.Properties properties)
+	{
+		super(properties);
+	}
+	
+	@Override
+	public InteractionResultHolder<ItemStack> use(Level levelIn, Player playerIn, InteractionHand handIn)
```
kirderf1, line 39: The alternative function `item.useOn()` seems to be designed for items right-clicked on blocks. Is there a reason that you're using `use()` instead of `useOn()`? If not, I'd suggest using that one instead.

src/main/java/com/mraof/minestuck/network/computer/BurnDiskPacket.java
kirderf1: The function takes a block entity consumer as an argument. You'd use the computer block entity inside this consumer as you're right now using the block entity after the instanceof check.

src/main/java/com/mraof/minestuck/util/BlockHitResultUtil.java
kirderf1: With the fluid check, is it the `ClipContext.Fluid` argument you're referring to? Because it is not hardcoded in the item version of the function.

src/main/java/com/mraof/minestuck/util/MSTags.java
```diff
@@ -158,4 +166,22 @@ private static TagKey<Biome> tag(String name)
...
 		}
 	}
+	
+	/**
+	 * Used when getBlocksFromTag has already been called to avoid redundancy
+	 */
+	public static boolean blockExistsInList(List<Block> blockList, Block block)
+	{
+		return blockList.contains(block);
+	}
```
kirderf1, line 176: In its current state, I suggest that you inline this function

src/main/java/com/mraof/minestuck/world/gen/structure/FrogTemplePiece.java
```diff
@@ -27,12 +26,16 @@
...
+import java.util.List;
+import java.util.Optional;
 import java.util.Random;
 
 public class FrogTemplePiece extends CoreCompatibleScatteredStructurePiece
 {
 	private boolean createRan = false; //boolean check to prevent certain objects(the lotus flower entity) from spawning several times over
 	private static final FrogTemplePiece.Selector HIEROGLYPHS = new FrogTemplePiece.Selector();
+	private static final List<WeightedEntry.Wrapper<Block>> weightedBlockList = buildWeightedList();
+	private static final int totalWeight = WeightedRandom.getTotalWeight(weightedBlockList);
```
kirderf1, line 38: Please move these inside the selector class.

src/main/java/com/mraof/minestuck/world/gen/structure/FrogTemplePiece.java
```diff
@@ -27,12 +26,16 @@
...
 
+import java.util.List;
+import java.util.Optional;
 import java.util.Random;
 
 public class FrogTemplePiece extends CoreCompatibleScatteredStructurePiece
 {
 	private boolean createRan = false; //boolean check to prevent certain objects(the lotus flower entity) from spawning several times over
 	private static final FrogTemplePiece.Selector HIEROGLYPHS = new FrogTemplePiece.Selector();
+	private static final List<WeightedEntry.Wrapper<Block>> weightedBlockList = buildWeightedList();
```
kirderf1, line 37: Problem: You're calling `get()` on registry objects during static initialization, this is not good! You could either make the list non-static (so that it is initialized when the structure piece is created, which will not happen until after things have been loaded), or you could store them as suppliers instead of blocks.

src/main/java/com/mraof/minestuck/item/loot/functions/SetSburbCodeFragments.java
```diff
@@ -0,0 +1,72 @@
...
+	{
+		List<Block> completeList = MSTags.getBlocksFromTag(MSTags.Blocks.GREEN_HIEROGLYPHS);
+		List<Block> pickedList = new ArrayList<>();
+		if(!completeList.isEmpty())
+		{
+			int numberOfIterations = Math.max(1, random.nextInt(Math.min(3, completeList.size())));
+			
+			for(int iterate = 0; iterate < numberOfIterations; iterate++) //up to two runs at filling the book with another fragment of code
+			{
+				pickedList.add(completeList.get(random.nextInt(completeList.size()))); //random element from the full list of blocks
```
kirderf1, line 49: Seems like this could end up adding the same block twice (or more) to the list. This doesn't seem intended.

kirderf1 (changes requested):

src/main/java/com/mraof/minestuck/client/gui/ReadableSburbCodeScreen.java
Dweblenod: That seems reasonable

src/main/java/com/mraof/minestuck/computer/DiskBurner.java
Dweblenod: While the code correctly prevents players from burning more disks than there are blank disks stored in the computer, the display of available programs to burn is not immediately updated. What should happen is that as soon as there are no more blank disks available to burn, the buttons for burning disks should dissapear. However it takes multiple clicks for the button display to update

src/main/java/com/mraof/minestuck/block/machine/ComputerBlock.java
Dweblenod: I have no plans of my own, but I was wondering if it would be appropriate to keep it as is in case addons wanted to add their own book versions? There may be other sections of code I cant think of right now that would prevent that but it seemed useful to avoid hardcoding

src/main/java/com/mraof/minestuck/item/ReadableSburbCodeItem.java
Dweblenod: I dont remember if this was the reason originally, but I cant seem to get the screen to be openable when using useOn

src/main/java/com/mraof/minestuck/network/computer/BurnDiskPacket.java
Dweblenod: Should be good now but will not resolve until you have approved how its been changed

src/main/java/com/mraof/minestuck/world/gen/structure/FrogTemplePiece.java
Dweblenod: Made non static

src/main/java/com/mraof/minestuck/block/machine/ComputerBlock.java
kirderf1: Oh, here's a related idea: instead of handling sburb codes as part of `computerBlock.use()`, how about instead using `sburbCodeItem.onItemUseFirst()`?

src/main/java/com/mraof/minestuck/client/gui/ReadableSburbCodeScreen.java
```diff
@@ -0,0 +1,285 @@
...
+			this.addRenderableWidget(new Button(this.width / 2 - 100, 196, 200, 20, new TranslatableComponent("gui.done"), (p_214161_1_) -> {
+				this.minecraft.setScreen(null);
+			}));
+		}
+	}
+	
+	protected void createPageControlButtons()
+	{
+		int i = (this.width - 192) / 2;
+		this.forwardButton = this.addRenderableWidget(new PageButton(i + 116, 159, true, (p_214159_1_) -> {
```
kirderf1, line 224: Don't use SRG-names in our own code. You should also use the lambda form without braces here.

src/main/java/com/mraof/minestuck/client/gui/ReadableSburbCodeScreen.java
```diff
@@ -0,0 +1,285 @@
...
+		{
+			return true;
+		} else
+		{
+			switch(pKeyCode)
+			{
+				case 266:
+					this.backButton.onPress();
+					return true;
+				case 267:
```
kirderf1, line 277: Use the appropriate constants from `GLFW` instead.

src/main/java/com/mraof/minestuck/item/IncompleteSburbCodeItem.java
```diff
@@ -0,0 +1,176 @@
...
+	public InteractionResult useOn(UseOnContext context)
+	{
+		ItemStack stackInUse = context.getItemInHand();
+		Level worldIn = context.getLevel();
+		Player player = context.getPlayer();
+		InteractionHand handIn = context.getHand();
+		
+		if(!worldIn.isClientSide && player != null)
+		{
+			BlockHitResult blockRayTraceResult = BlockHitResultUtil.getPlayerPOVHitResult(worldIn, player);
```
kirderf1, line 50: A hit result should already be available from the context.

src/main/java/com/mraof/minestuck/item/IncompleteSburbCodeItem.java
```diff
@@ -0,0 +1,176 @@
...
+ */
+public class IncompleteSburbCodeItem extends ReadableSburbCodeItem
+{
+	public IncompleteSburbCodeItem(Properties properties)
+	{
+		super(properties);
+	}
+	
+	@Override
+	public InteractionResult useOn(UseOnContext context)
```
kirderf1, line 41: The returned interaction result should not be "pass" when this interaction actually does something.

src/main/java/com/mraof/minestuck/item/IncompleteSburbCodeItem.java
```diff
@@ -0,0 +1,176 @@
...
+		{
+			for(Block block : hieroglpyhsList)
+			{
+				if(block == hieroglyphBlock)
+				{
+					addRecordedInfo(stackInUse, hieroglyphBlock);
+					break;
+				}
+			}
+		}
```
kirderf1, line 88: Looks like you're checking that `hieroglyphsList` contains `hieroglyphBlock` twice; once with a function, and then once manually.

src/main/java/com/mraof/minestuck/blockentity/ComputerBlockEntity.java
```diff
@@ -47,32 +50,42 @@ public ComputerBlockEntity(BlockPos pos, BlockState state)
...
 		else this.owner = IdentifierHandler.load(nbt, "owner");
 		if(gui != null)
 			gui.updateGui();
+		
+		if(nbt.contains("hieroglyphsStored"))
+			hieroglyphsStored = ReadableSburbCodeItem.getRecordedBlocks(nbt.getList("hieroglyphsStored", Tag.TAG_STRING));
+		if(nbt.contains("hasParadoxInfoStored"))
+			hasParadoxInfoStored = nbt.getBoolean("hasParadoxInfoStored");
+		if(nbt.contains("blankDisksStored"))
+			blankDisksStored = nbt.getInt("blankDisksStored");
```
kirderf1, line 88: Move this above the `updateGui()` call. You want to load hieroglyph and bland disk data before updating the gui. This is the cause of the problem you described in the disk burner class.

src/main/java/com/mraof/minestuck/block/machine/ComputerBlock.java
```diff
@@ -161,16 +236,32 @@ private void dropItems(Level level, int x, int y, int z, BlockState state)
...
+		}
+		
+		for(int iterate = 0; iterate < be.blankDisksStored; iterate++)
+		{
+			float rx = rand.nextFloat() * 0.8F + 0.1F;
+			float ry = rand.nextFloat() * 0.8F + 0.1F;
+			float rz = rand.nextFloat() * 0.8F + 0.1F;
+			ItemEntity entityItem = new ItemEntity(level, x + rx, y + ry, z + rz, MSItems.BLANK_DISK.get().getDefaultInstance());
 			entityItem.setDeltaMovement(rand.nextGaussian() * factor, rand.nextGaussian() * factor + 0.2F, rand.nextGaussian() * factor);
 			level.addFreshEntity(entityItem);
```
kirderf1, line 255: After seeing this repeated now in three places, I went and looked for a helper function and found `Containers.dropItemStack()`. I suggest using that one! I would then also recommend using it once with the stack size set to the number of disks to drop, rather than looping over the amount to drop them one at a time.

src/main/java/com/mraof/minestuck/client/gui/ReadableSburbCodeScreen.java
```diff
@@ -0,0 +1,285 @@
...
+			Resource resource = this.minecraft.getResourceManager().getResource(new ResourceLocation(Minestuck.MOD_ID, "texts/rana_temporaria_sec22b.txt")); //The text is broken into lines 48 characters long to neatly fit within a block of text that avoids empty spaces
+			InputStream inputStream = resource.getInputStream();
+			BufferedReader bufferedReader = new BufferedReader(new InputStreamReader(inputStream, StandardCharsets.UTF_8));
+			textList = bufferedReader.lines().collect(Collectors.toList());
+			
+			linesPerBlock = textList.size() / MAX_HIEROGLYPH_COUNT;
+			totalPages = Math.round((float) (textList.size() + 1) / (float) LINES_PER_PAGE);
+		} catch(Exception ignored)
+		{
+		}
```
kirderf1, line 88: You should ad a finally clause that tries to close the resource in the case of an error (see `WinScreen.wrapCreditsIO()`. You may as well also log the exception if any occur)

kirderf1 (changes requested):

src/main/java/com/mraof/minestuck/util/BlockHitResultUtil.java
Dweblenod: Ah wait I forgot a reason why I did this upon going to delete. The original function has protected access

src/main/java/com/mraof/minestuck/computer/DiskBurner.java
Dweblenod: Your code advice resolved this issue, thanks!

src/main/java/com/mraof/minestuck/block/machine/ComputerBlock.java
Dweblenod: Changed, and seems to work fine in the new setting

src/main/java/com/mraof/minestuck/util/MSTags.java
```diff
@@ -9,15 +10,22 @@
...
+import net.minecraftforge.registries.ForgeRegistries;
+
+import java.util.ArrayList;
+import java.util.List;
 
 public class MSTags
 {
 	public static class Blocks
 	{
+		//make an alternative block tag for hieroglyph block sets that accomplish the same task, every block included in the green hieroglyph tag is required to get completed code
```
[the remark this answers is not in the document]
Dweblenod: Let me know if the new version makes sense

src/main/java/com/mraof/minestuck/item/IncompleteSburbCodeItem.java
Dweblenod: Removed the if statement

src/main/java/com/mraof/minestuck/item/IncompleteSburbCodeItem.java
Dweblenod: Let me know if you consider the changes appropriate

src/main/java/com/mraof/minestuck/item/loot/MSLootEvents.java
```diff
@@ -2,28 +2,137 @@
...
+		return new MerchantOffer(emeraldStack, weaponStack, maxUses, villagerXp, priceMultiplier);
+	}
+	
+	@SubscribeEvent
+	public static void addCustomWanderingVillagerTrade(WandererTradesEvent event)
+	{
+		List<VillagerTrades.ItemListing> trades = event.getGenericTrades();
+		//between the two new trades, there is an approximately 25% chance for a blank disk to appear in a wandering trader slot(as of 1.16 with no further modifications to the trades)
+		trades.add(1, (villager, random) -> new MerchantOffer(new ItemStack(Items.EMERALD, 3), new ItemStack(MSItems.BLANK_DISK.get()), 3, 12, 0.05F));
+		trades.add(1, (villager, random) -> new MerchantOffer(new ItemStack(Items.EMERALD, 3), new ItemStack(MSItems.BLANK_DISK.get()), 3, 12, 0.05F));
```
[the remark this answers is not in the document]
Dweblenod: I believe the choice of index is arbitrary, however they occupy the same spot because I did not want more than 1 blank disk trade per trader

src/main/java/com/mraof/minestuck/item/IncompleteSburbCodeItem.java
kirderf1: Rather than removing `contains()`, how about keeping specifically that part:
```java
		if(hieroglpyhsList.contains(hieroglyphBlock)
		{
			addRecordedInfo(stackInUse, hieroglyphBlock);
		}
```

src/main/java/com/mraof/minestuck/skaianet/SkaianetHandler.java
```diff
@@ -4,19 +4,24 @@
...
 import com.mraof.minestuck.blockentity.ComputerBlockEntity;
+import net.minecraft.core.BlockPos;
 import net.minecraft.nbt.CompoundTag;
 import net.minecraft.nbt.Tag;
 import net.minecraft.resources.ResourceKey;
 import net.minecraft.server.MinecraftServer;
 import net.minecraft.server.level.ServerLevel;
 import net.minecraft.server.level.ServerPlayer;
+import net.minecraft.world.entity.item.ItemEntity;
+import net.minecraft.world.item.ItemStack;
```
kirderf1, line 22: Unused imports

src/main/java/com/mraof/minestuck/blockentity/ComputerBlockEntity.java
```diff
@@ -221,6 +242,34 @@ public void putServerMessage(String message)
...
+			ItemEntity entityItem = new ItemEntity(level, bePos.getX() + rx, bePos.getY() + ry, bePos.getZ() + rz, diskStack);
+			entityItem.setDeltaMovement(random.nextGaussian() * 0.05F, random.nextGaussian() * 0.05F + 0.2F, random.nextGaussian() * 0.05F);
+			level.addFreshEntity(entityItem);
+			
+			blankDisksStored--;
+			
+			//Imitates the structure block to ensure that changes are sent client-side
+			setChanged();
+			BlockState state = getBlockState();
+			level.sendBlockUpdated(bePos, state, state, 3);
```
kirderf1, line 269:
```suggestion
			markBlockForUpdate();
```

src/main/java/com/mraof/minestuck/blockentity/ComputerBlockEntity.java
```diff
@@ -221,6 +242,34 @@ public void putServerMessage(String message)
...
 	}
 	
+	public void burnDisk(int programId)
+	{
+		if(level == null)
+			return;
+		
+		BlockPos bePos = getBlockPos();
+		ItemStack diskStack = ProgramData.getItem(programId);
+		if(!diskStack.isEmpty() && blankDisksStored > 0)
```
kirderf1, line 252: Realized that you never actually verify here that this computer has the info to burn these disks. You need to do this.

src/main/java/com/mraof/minestuck/client/gui/ReadableSburbCodeScreen.java
```diff
@@ -0,0 +1,298 @@
...
+		else if(pageModulo == 1)
+			RenderSystem.setShaderTexture(0, BOOK_TEXTURES_02);
+		else if(pageModulo == 2)
+			RenderSystem.setShaderTexture(0, BOOK_TEXTURES_03);
+		
+		
+		int topX = (this.width - GUI_WIDTH) / 2;
+		int topY = 2;
+		this.blit(poseStack, topX, topY, 0, 0, GUI_WIDTH, GUI_HEIGHT);
+		{
```
kirderf1, line 171: There are some purposeless braces here that I suggest that you remove

src/main/java/com/mraof/minestuck/client/gui/ReadableSburbCodeScreen.java
```diff
@@ -0,0 +1,298 @@
...
+							Component line = new TextComponent(text.substring(start, end)).setStyle(style);
+							font.draw(poseStack, line, ((this.width - GUI_WIDTH) / 2F + TEXT_OFFSET_X) * scale, (lineY.intValue() + TEXT_OFFSET_Y) * scale, 0x00A300); //hex is green
+							lineY.add(CUSTOM_LINE_HEIGHT);
+						}
+					});
+				}
+			}
+		}
+		
+		poseStack.popPose();
```
kirderf1, line 198: Move this inside the same block as the push. A pop must only ever be called when it is certain that its corresponding push has been called.

src/main/java/com/mraof/minestuck/client/gui/ReadableSburbCodeScreen.java
```diff
@@ -0,0 +1,298 @@
...
+		
+		for(int iterate = 0; iterate < MAX_HIEROGLYPH_COUNT; iterate++)
+		{
+			if(iterate == 0) //first iteration, checks for paradox code
+				booleans[iterate] = paradoxCode;
+			else if(!recordedBlockList.isEmpty() && iterate < FULL_HIEROGLYPH_LIST.size() + 1) //all but first iteration, checks for blocks in FULL_HIEROGLYPH_LIST, +1 to offset paradox code
+				booleans[iterate] = recordedBlockList.contains(FULL_HIEROGLYPH_LIST.get(iterate - 1));
+			else //unrecorded blocks
+				booleans[iterate] = false;
+		}
```
kirderf1, line 218:
```suggestion
		booleans[0] = paradoxCode;
		if(!recordedBlockList.isEmpty())
		{
			for(int blockIndex = 0; blockIndex < FULL_HIEROGLYPH_LIST.size(); blockIndex++)
				booleans[blockIndex + 1] = recordedBlockList.contains(FULL_HIEROGLYPH_LIST.get(blockIndex));
		}
```

src/main/java/com/mraof/minestuck/blockentity/ComputerBlockEntity.java
Dweblenod: The button is not displayed on the screen if the computer does not have the necessary components which keeps it in check, or is there another angle to the issue?

kirderf1: Yeah, there's network packet security. A player with a modified mod (or perhaps just a custom addon) could in theory send whatever packet they like. For that reason, when a player says that it performs some action by sending a packet, basic level of anti-cheat is to check that the player actually can perform that action.

src/main/java/com/mraof/minestuck/client/gui/ReadableSburbCodeScreen.java
```diff
@@ -0,0 +1,294 @@
...
+		return lineY < LINES_PER_PAGE * CUSTOM_LINE_HEIGHT;
+	}
+	
+	protected void createMenuControls()
+	{
+		if(this.minecraft != null)
+		{
+			this.addRenderableWidget(new Button(this.width / 2 - 100, 196, 200, 20, new TranslatableComponent("gui.done"), (p_214161_1_) -> {
+				this.minecraft.setScreen(null);
+			}));
```
kirderf1, line 230:
```suggestion
			this.addRenderableWidget(new Button(this.width / 2 - 100, 196, 200, 20, new TranslatableComponent("gui.done"), button -> this.minecraft.setScreen(null)));
```

src/main/java/com/mraof/minestuck/data/loot_table/MSMiscLootTables.java
```diff
@@ -32,5 +35,20 @@ public void accept(BiConsumer<ResourceLocation, LootTable.Builder> lootProcessor
...
+						.add(LootItem.lootTableItem(MSItems.COMPUTER_PARTS.get()).setWeight(10).setQuality(3)))
+				.withPool(LootPool.lootPool().name("code").setRolls(ConstantValue.exactly(1))
+						.add(LootItem.lootTableItem(MSItems.SBURB_CODE.get()).setWeight(10).setQuality(3)).apply(SetNbtFunction.setTag(giveParadoxFragment())))
+		);
+	}
+	
+	private static CompoundTag giveParadoxFragment()
+	{
+		CompoundTag nbt = new CompoundTag();
+		nbt.putBoolean("hasParadoxInfo", true);
```
kirderf1, line 50: Probably better to handle this in `IncompleteSburbCodeItem`, so that *how* the paradox fragment is stored in nbt is determined all in one place.

src/main/java/com/mraof/minestuck/item/loot/MSLootEvents.java
```diff
@@ -2,28 +2,137 @@
...
+		{
+			trades.get(2).add((villager, random) -> createEnchantedItemOffer(random, new ItemStack(MSItems.LUCERNE_HAMMER.get()), 4, 3, 1, 0.05F));
+		}
+		
+		if(event.getType() == VillagerProfession.LIBRARIAN)
+		{
+			ItemStack[] mediumBookTypes = new ItemStack[]{new ItemStack(MSItems.NONBINARY_CODE.get()), new ItemStack(MSItems.BINARY_CODE.get()), new ItemStack(MSItems.TILLDEATH_HANDBOOK.get()), new ItemStack(MSItems.TABLESTUCK_MANUAL.get()), new ItemStack(MSItems.WISEGUY.get()), new ItemStack(MSItems.SASSACRE_TEXT.get()), new ItemStack(MSItems.FLARP_MANUAL.get())};
+			ItemStack[] overworldBookTypes = new ItemStack[]{new ItemStack(MSItems.COMPLETED_SBURB_CODE.get())};
+			trades.get(2).add((villager, random) -> new MerchantOffer(MSDimensions.isInMedium(villager.getServer(), villager.level.dimension()) ? mediumBookTypes[random.nextInt(mediumBookTypes.length)] : overworldBookTypes[random.nextInt(overworldBookTypes.length)], new ItemStack(Items.EMERALD, 3), 4, 2, 0.05F));
+		}
```
Dweblenod, line 91: @kirderf1 do you have any recommendations on how to reduce the frequency of a given trade? The only method I can see to reduce the frequency is to set these as a trade for an index which has a lot of other options. However in the case of the cartographer, the rate at which frog temple maps are a trade option is over 50%. Im going to change it so it shares the same index with the ocean monument and woodland mansion to avoid the issue but if you have advice it may be useful for the future

src/main/java/com/mraof/minestuck/item/loot/MSLootEvents.java
```diff
@@ -2,28 +2,137 @@
...
+	
+	@SubscribeEvent
+	public static void addCustomVillagerTrade(VillagerTradesEvent event)
+	{
+		Int2ObjectMap<List<VillagerTrades.ItemListing>> trades = event.getTrades();
+		
+		if(event.getType() == VillagerProfession.CARTOGRAPHER)
+		{
+			trades.get(1).add((villager, random) -> new MerchantOffer(new ItemStack(Items.EMERALD, 8), new ItemStack(Items.COMPASS), createFrogTempleMap(villager), 12, 7, 0.2F));
+		}
```
Dweblenod, line 68: the world freezes up(lag) when a cartographer who has the frog temple map is first clicked, assumedly due to the trade loading. I use code I have assumed is pretty much identical to how vanilla handles the creation of the vanilla maps, but obviously they do not have any performance issues, its a one time thing per villager but it would be ideal to find a solution

src/main/java/com/mraof/minestuck/data/loot_table/MSMiscLootTables.java
Dweblenod: Moved it into IncompleteSburbCodeItem. Didnt seem appropriate to combine it with the existing methods if that was part of the recommendation so let me know if you want further changes

src/main/java/com/mraof/minestuck/item/loot/MSLootEvents.java
kirderf1: So from the code and minecraft wiki, it looks like trading works something like this:
Villagers go through 5 different levels, and get up to two trade offers for each level, with each level for each profession having different possible trade offers.
When choosing offers, if there are two or less possible offers, it chooses that one or two offers. Otherwise it chooses two offers randomly.
Cartographer level 2 has a glass pane trade and ocean explorer map, which would normally choose those two every time. With the frog temple map, it'll choose two among the three, giving each item a probability of 2/3 to appear.

I think there are three different approaches that can be taken:
- Accept that offers added by us may contend with few other offers and thus have a high probability of appearing (what offers would we add that actually should have a lower probability of appearing?)
- Try to add offers to levels which have a lot of offers already, such as cartographer 4 or mason 4.
- Add copies of the vanilla trade offers to increase the probability of those. While this is an option, I don't think it's a good option as it goes against the weight-free design of trade offers.

src/main/generated/resources/assets/minestuck/lang/en_us.json
```diff
@@ -991,6 +992,7 @@
   "entity.minestuck.turtle.sound": "...",
   "entity.minestuck.turtle.sound.2": "..!",
   "entity.minestuck.vitality_gel": "Vitality Gel",
+  "filled_map.tagkey[minecraft:worldgen/configured_structure_feature / minestuck:scanner_located]": "Frog Temple",
```
kirderf1, line 995: This is not a good translation key. Since `MSLootEvents.createFrogTempleMap()` isn't trying to be a general function, how about just hardcoding the translation akin to vanilla?

src/main/java/com/mraof/minestuck/item/loot/MSLootEvents.java
```diff
@@ -2,28 +2,144 @@
...
+				ItemStack itemstack = MapItem.create(serverLevel, templePos.getX(), templePos.getZ(), (byte) 2, true, true);
+				MapItem.renderBiomePreviewMap(serverLevel, itemstack);
+				MapItemSavedData.addTargetDecoration(itemstack, templePos, "+", MapDecoration.Type.RED_X);
+				itemstack.setHoverName(new TranslatableComponent("filled_map." + MSTags.Structures.SCANNER_LOCATED.toString().toLowerCase(Locale.ROOT)));
+				
+				return itemstack;
+			}
+		}
+		
+		return new ItemStack(MSItems.DICE.get());
```
kirderf1, line 122: Doesn't seem right. Handle failures like vanilla does instead (ie return a null trade offer if it can't generate one)

src/main/java/com/mraof/minestuck/item/loot/MSLootEvents.java
kirderf1: After pausing the game in debug mode when the world freezes up, it was [in the middle of finding the minimum block for the frog temple](https://github.com/Dweblenod/Minestuck/blob/7b470ae0b9b7a326e6a2a4e565879883e68e8168/src/main/java/com/mraof/minestuck/world/gen/structure/FrogTemplePiece.java#L46). There's a good reason to think that we need to rewrite the code for height displacement of that frog temple piece into something that performs better.

Dweblenod: is that something you think is appropriate to do in this PR?

kirderf1: Not necessarily part of this PR as it doesn't already touch that part of the code, but it is a good idea to fix it before the next major release.

src/main/java/com/mraof/minestuck/item/loot/MSLootEvents.java
kirderf1: A null item stack result is not acceptable. You need to return a null merchant offer instead. It may be easier to do if you change the function to return the merchant offer instead. To give a second alternative, you could have it return an `Optional<ItemStack>`, and then you could used it with `createFrogTempleMap().map(mapItem -> ...).orElse(null)`.

Dweblenod: should be good now

src/main/java/com/mraof/minestuck/item/loot/MSLootEvents.java
Dweblenod: moved it to the second level trade

src/main/java/com/mraof/minestuck/item/loot/MSLootEvents.java
Dweblenod: Pardon, wrong comment, still need to fix

Dweblenod: Now returns a merchant offer instead of item stack

src/main/java/com/mraof/minestuck/item/loot/MSLootEvents.java
Dweblenod: Then I will resolve this particular conversation

src/main/java/com/mraof/minestuck/data/MinestuckEnUsLanguageProvider.java
```diff
@@ -877,11 +880,18 @@ protected void addTranslations()
 		addItem(MSItems.CRUXITE_APPLE, "Cruxite Artifact");
 		addItem(MSItems.CRUXITE_POTION, "Cruxite Artifact");
 		addItem(MSItems.SBURB_CODE, "SBURB Code");
-		addItemTooltip(MSItems.SBURB_CODE, "It seems temple hieroglyphs can be translated into computer game code using a Book and Quill.");
+		addItemTooltip(MSItems.SBURB_CODE, "It seems hieroglyphs can be translated into genetic code. Perhaps this genetic code could in turn be made into computer code?");
+		addItemExtra(MSItems.SBURB_CODE, "completion", "§dPercent completion: %s%%§r"); //section signs result in the percent number being light purple, double percent symbol required to get it to show up in game
+		addItemExtra(MSItems.SBURB_CODE, "paradox_hint", "All of the hieroglyphs that can be found have been recorded here, what could be missing?");
+		addItemExtra(MSItems.SBURB_CODE, "additional_info", "Right click on different types of hieroglyph blocks to record them, the air to read the code, or a computer to exchange code. Create new code books using a Book and Quill on hieroglyphs.");		addItemExtra(MSItems.SBURB_CODE, "shift_for_more_info", "Press §eSHIFT§r for more info");
```
kirderf1, line 886:
```suggestion
		addItemExtra(MSItems.SBURB_CODE, "additional_info", "Right click on different types of hieroglyph blocks to record them, the air to read the code, or a computer to exchange code. Create new code books using a Book and Quill on hieroglyphs.");
		addItemExtra(MSItems.SBURB_CODE, "shift_for_more_info", "Press §eSHIFT§r for more info");
```

kirderf1 (approved):
Looks good! Just make sure to resolve the merge conflict as well.

PR #352, opened 2022-07-02
Ground Slam and tweaks to animated attack framework

Dweblenod:
Gives Ogres a ground slam attack that is used when the ogre cannot reach a player to attack them normally. The ground slam propels players that are on the ground and who are not crouching into the air with additional semi-random horizontal momentum that may knock them off of whatever they are standing on. 

Adds a cooldown to prevent certain Goals from activating to prevent entities spamming special attacks such as ground slam and reworks AnimatedAttackWhenInRangeGoal to allow for more variation in the range at which the Goal will initiate.

Animated goals will also now not be interrupted by other goals by default.

There is now a slam sound effect and a swoosh sound effect. The first is used with the ground slam, and the latter is planned to be used for multiple animated attacks across multiple entity types.

Animation is by Riotmode, slam sound effect is by triageGremlin, swoosh sound effect is by Dweblenod


Dweblenod:
Should be changed to focus less on if the target is a player entity

Dweblenod:
Is ready for review but needs to have sounds effects added for the slam

kirderf1 (changes requested):

src/main/java/com/mraof/minestuck/entity/underling/OgreEntity.java
```diff
@@ -38,12 +42,20 @@ public static AttributeModifierMap.MutableAttribute ogreAttributes()
 				.add(Attributes.ATTACK_DAMAGE, 6);
 	}
 	
+	@Override
+	protected void defineSynchedData()
+	{
+		super.defineSynchedData();
+		entityData.define(CURRENT_ATTACK_TYPE, AnimatedCreatureEntity.Actions.values().length + 1);
```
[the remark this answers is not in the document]
Dweblenod: I was under the impression that this data had to exist in a new ordinal, so this seemed like the way to ensure it was an unused location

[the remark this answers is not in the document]
kirderf1: Hang on, could you explain what you think that the second argument is for?

src/main/java/com/mraof/minestuck/entity/MSEntityTypes.java
```diff
@@ -23,6 +23,7 @@
 @Mod.EventBusSubscriber(modid = Minestuck.MOD_ID, bus = Mod.EventBusSubscriber.Bus.MOD)
 public final class MSEntityTypes
 {
+	//TODO reduce underling rate
```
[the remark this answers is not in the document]
Dweblenod: I assume this was a suggestion to reduce the spawn rate of the underlings, however there are more appropriate PRs to include this and there is even one that changes this

Dweblenod:
Chose to force push and rework some of the changes

src/main/java/com/mraof/minestuck/entity/ai/attack/GroundSlamGoal.java
```diff
@@ -0,0 +1,110 @@
...
+		if(doHurt)
+		{
+			//after the clamp is applied: 10 damage at 1-2 blocks away, 3.4 damage at 5 blocks away, 1 damage at 10 blocks away, 0.2 damage at 15 blocks away
+			double distanceModified = (1.2 / (Math.sqrt(attacker.distanceToSqr(target)) * 0.05)) - 1.4;
+			float damage = Math.min(Math.max((float) distanceModified, 0.2F), 10F);
+			target.hurt(DamageSource.mobAttack(attacker), damage);
+		}
+		
+		//non-player characters don't need to be damaged
+		target.push(target.getRandom().nextFloat() - 0.5, 0.75, target.getRandom().nextFloat() - 0.5);
```
Dweblenod, line 108: Im happy with how this is implemented now, but it would be nice to have a simple way to move a player without injuring them

src/main/java/com/mraof/minestuck/entity/ai/attack/GroundSlamGoal.java
```diff
@@ -0,0 +1,110 @@
...
+			{
+				if(iteratedEntity != this.entity && !iteratedEntity.existingCooldownIsLonger(GROUP_SLAM_COOLDOWN))
+				{
+					//TODO slamGoal bool does not seem to trigger
+					for(Goal iteratedGoal : iteratedEntity.goalSelector.getAvailableGoals())
+					{
+						boolean slamGoal = iteratedGoal.equals(this);
+						if(slamGoal)
+							iteratedEntity.setCooldown(GROUP_SLAM_COOLDOWN + phasedAnimation.getTotalAnimationLength()); //plays at beginning to prevent overlapping
+					}
```
Dweblenod, line 55: Trying to make it so that it checks for all nearby entities, and if those nearby entities possess the GroundSlamGoal then a cooldown is also applied to them. Not sure if this is a bad approach or if I messed up the conditions of it since it doesn't seem to work

Dweblenod: No longer trying to give a cooldown via Goal. Now it just gives a group cooldown to entities of the same type

kirderf1: Yeah, that sounds sensible for now.

src/main/java/com/mraof/minestuck/entity/ai/attack/GroundSlamGoal.java
kirderf1: Are there any examples in vanilla that you've looked at or could look at?

Dweblenod: I believe I looked into it when I first started working on this, and I hadnt found what I was looking for, Might investigate the .push() method

kirderf1 (changes requested):
Gonna look into moving the player as well when I next have the time.

src/main/java/com/mraof/minestuck/entity/ai/attack/GroundSlamGoal.java
```diff
@@ -0,0 +1,104 @@
...
+		if(target != null && targetCanBeHit(target) && target instanceof Player playerTarget && playerTarget.isOnGround())
+		{
+			boolean inMeleeRange = getStandardAttackReachSqr(target) >= entity.distanceToSqr(target);
+			
+			//slam will connect if the player is close or if the player is farther away but not crouching, always requires player be on the ground
+			if(inMeleeRange || !playerTarget.isCrouching())
+			{
+				fling(attacker, playerTarget, true);
+			}
+		}
```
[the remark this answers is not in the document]
Dweblenod: Ah yes you are correct, I dont want it to be specific to a player. That used to be my priority

src/main/java/com/mraof/minestuck/entity/ai/attack/GroundSlamGoal.java
```diff
@@ -0,0 +1,104 @@
...
+		//flinging any nearby smaller mobs on the ground indiscriminately
+		AABB aabb = new AABB(attacker.blockPosition()).inflate(4);
+		List<LivingEntity> entityList = level.getEntitiesOfClass(LivingEntity.class, aabb);
+		if(!entityList.isEmpty())
+		{
+			for(LivingEntity iteratedEntity : entityList)
+			{
+				boolean smallerThanAttacker = iteratedEntity.getBoundingBox().getSize() < attacker.getBoundingBox().getSize();
+				
+				if(iteratedEntity != attacker && iteratedEntity != target && iteratedEntity.isOnGround() && !iteratedEntity.noPhysics && smallerThanAttacker)
```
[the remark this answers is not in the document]
Dweblenod: Not for this, just because entities that are close by get hit even when they dont crouch

src/main/java/com/mraof/minestuck/entity/ai/attack/GroundSlamGoal.java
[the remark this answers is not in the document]
Dweblenod: changed

src/main/java/com/mraof/minestuck/entity/ai/attack/GroundSlamGoal.java
```diff
@@ -0,0 +1,104 @@
...
```
kirderf1, line 69: Should the distance also be checked here, perhaps?

kirderf1 (commented):
Looks pretty good, and appears to work well in-game.

Some thing that I wonder about is that applying some (or small) damage seems appropriate for any entity that is flinged rather than just the target. You do cite underlings targeting each other as a reason, though I think that it'd be a good idea to make underlings not target each other in the first place. I believe that can easily be done in `UnderlingEntity` by moving the underling tag check from `addEnemy()` to `setTarget()`.
Does that seem sound?

Also with the regular attack possibly cancelling out the ground slam, I'm not sure if you can get within melee range and still get the ground attack.  Perhaps the ground slam goal should override `Goal.isInterruptable()` and return false?

src/main/java/com/mraof/minestuck/entity/ai/attack/GroundSlamGoal.java
Dweblenod: Im not sure I am seeing why, since the AABB already defines a similar area of distance. Or am I misunderstanding?

kirderf1: Well, it's not entirely consistent with how distances work with the target. The AABB is also currently using some arbitrary and possibly different size.

Dweblenod:
> Looks pretty good, and appears to work well in-game.
> 
> Some thing that I wonder about is that applying some (or small) damage seems appropriate for any entity that is flinged rather than just the target. You do cite underlings targeting each other as a reason, though I think that it'd be a good idea to make underlings not target each other in the first place. I believe that can easily be done in `UnderlingEntity` by moving the underling tag check from `addEnemy()` to `setTarget()`. Does that seem sound?
> 
> Also with the regular attack possibly cancelling out the ground slam, I'm not sure if you can get within melee range and still get the ground attack. Perhaps the ground slam goal should override `Goal.isInterruptable()` and return false?

I personally feel that even if the underlings had their tag changed to prevent aggression against each other, that having everything around an ogre get injured would not improve gameplay. 

For the punch canceling the ground slam, setting it to uninterruptable sounds smart!

kirderf1:
Alright then, go for that.

src/main/java/com/mraof/minestuck/entity/ai/attack/GroundSlamGoal.java
Dweblenod: That is true, do you feel like their arbitrary relationship is too much?

kirderf1: It's a problem that I would want fixed. But I think it would be fine with a todo for now since it's not worth the time before 4/13.

Dweblenod: It hath been done

src/main/java/com/mraof/minestuck/entity/animation/ActionCooldown.java
```diff
@@ -0,0 +1,23 @@
+package com.mraof.minestuck.entity.animation;
+
+/**
+ * Used to help standardize the method in which action cooldown(how long between uses of an animated action) values are maintained
+ */
+public interface ActionCooldown
+{
+	//TODO Figure out if it can be made so that group cooldowns can be applied to any animated entities with the same Goal, as opposed to giving a cooldown to just entities of the same type
```
kirderf1, line 8: You could create a new interface that is implemented by all entites that should have their cooldown reset

kirderf1 (approved):

PR #353, opened 2022-07-04
Fireball goal and basilisk tweaks

Dweblenod:
Created a goal that allows basilisks to shoot fireballs at their target if that target is out of range of normal attacks. There now also exists a tail whip attack for when the target is too close for the bite attack to make sense. The basilisk is now also immune to fire damage to further the association and functionality of fire breath. The bite and whip attacks now have a sound effect to accompany them

Continues development from ground slam PR

Animations already existed, but were provided by Riotmode. Bite sound made by Dweblenod


src/main/java/com/mraof/minestuck/entity/ai/attack/FireballShootGoal.java
```diff
@@ -0,0 +1,123 @@
...
+import java.util.Objects;
+
+/**
+ * Alternative attack for when the attack target is out of standard range, a small fireball is shot at the targets initial position.
+ */
+public class FireballShootGoal<T extends AttackingAnimatedEntity> extends AnimatedAttackWhenInRangeGoal<T>
+{
+	public static final int GROUP_SHOOT_COOLDOWN = 20; //how long it should take before another nearby mob can shoot fireballs
+	
+	private BlockPos initialTargetPos;
```
kirderf1, line 25: Doesn't look like this pos is used properly. Is it meant to be used in place of the target position in `attemptToLandAttack()`?

kirderf1: Alright, you'll need to fix the goal so that's the case then

Dweblenod: oh I see now hehe its not actually used

src/main/java/com/mraof/minestuck/entity/ai/attack/FireballShootGoal.java
```diff
@@ -0,0 +1,129 @@
...
+	}
+	
+	@Override
+	public void start()
+	{
+		super.start();
+		
+		initialTargetPos = Objects.requireNonNull(this.entity.getTarget()).blockPosition();
+		
+		//code block handles group cooldown
```
kirderf1, line 48: I recommend instead extracting the code block as a function named like that

Dweblenod: it hath been done

Dweblenod:
Any other concerns with this PR?

kirderf1 (approved):

PR #354, opened 2022-07-04
Consort animation tweaks + Idle animation

Dweblenod:
Adding changes + additions made by Riotmode to the consorts, including an animation for when they are doing no other activities


Dweblenod:
It has been checked in game for functionality, so factoring in that and the simplicity of the idle animation code, I am merging this without waiting on other peoples review

PR #359, opened 2022-07-24
datapack generated starting modus

amiwyn:
I dont know if thats how you wanted it @kirderf1


src/main/java/com/mraof/minestuck/util/StartingModusManager.java
```diff
@@ -0,0 +1,56 @@
...
+	public StartingModusManager()
+	{
+		super(GSON, "minestuck/config");
+	}
+	
+	@Override
+	protected void apply(Map<ResourceLocation, JsonElement> jsonEntries, ResourceManager resourceManager, ProfilerFiller profiler)
+	{
+		JsonElement element = jsonEntries.get(new ResourceLocation(Minestuck.MOD_ID, "starting_modus"));
+		this.startingModusTypes = GSON.fromJson(element, new TypeToken<ArrayList<String>>(){}.getType());
```
kirderf1, line 38: It may be appropriate to parse this into a list of `ModusType`s, to take that responsibility away from `PlayerData`, and also to log an error earlier when some list entry cannot be parsed into a modus type.

src/main/java/com/mraof/minestuck/util/StartingModusManager.java
```diff
@@ -0,0 +1,56 @@
...
+import java.util.Map;
+import java.util.Objects;
+
+@Mod.EventBusSubscriber(bus = Mod.EventBusSubscriber.Bus.FORGE)
+public class StartingModusManager extends SimpleJsonResourceReloadListener
+{
+	private static final Gson GSON = new GsonBuilder().registerTypeAdapter(BoondollarPricing.class, new BoondollarPricing.Serializer()).create();
+	
+	private static StartingModusManager INSTANCE;
+	private List<String> startingModusTypes;
```
kirderf1, line 27: One improvement that should now be easy to make is to introduce weights. So letting each modus type to be paired with a weight number that determines how ofter that modus type is chosen.

Consider this a soft suggestion, as I don't think "weights" is a particularly important feature in this specific case.

kirderf1 (changes requested):
Yep, this is indeed what I had in mind. Just need a few changes to clean it up.

src/main/java/com/mraof/minestuck/data/MinestuckStartingModusProvider.java
```diff
@@ -0,0 +1,66 @@
...
+		pCache.putNew(path, hash);
+	}
+	
+	@Override
+	public String getName()
+	{
+		return "Minestuck Starting Modus";
+	}
+	
+	public List<String> getModusTypes()
```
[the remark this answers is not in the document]
amiwyn: oopsie

src/main/java/com/mraof/minestuck/util/StartingModusManager.java
```diff
@@ -0,0 +1,96 @@
...
+		return ModusTypes.REGISTRY.getValue(name);
+	}
+	
+	@SubscribeEvent
+	public static void onResourceReload(AddReloadListenerEvent event)
+	{
+		event.addListener(INSTANCE = new StartingModusManager());
+	}
+	
+	public static StartingModusManager getInstance()
```
kirderf1, line 87: This function appears unused in favor of `getStartingModusTypes()`. Could be removed if there is no specific point in keeping it.

src/main/java/com/mraof/minestuck/util/StartingModusManager.java
```diff
@@ -0,0 +1,96 @@
...
+	}
+	
+	private ModusType<?> createModus(String key)
+	{
+		ResourceLocation name = ResourceLocation.tryParse(key);
+		
+		if(name == null)
+			LOGGER.error("Unable to parse starting modus type {} as a resource location!", key);
+		
+		return ModusTypes.REGISTRY.getValue(name);
```
kirderf1, line 78: Could be made even better by logging an error when some name doesn't match any modus type (and the name string isn't empty)

kirderf1 (changes requested):
I'd also like an answer on my weight suggestion regarding if you want to implement it or leave it be.

kirderf1:
@amiwyn How is it going? Do you think that you'll be able to do the last cleanup to finish up this PR?

src/main/java/com/mraof/minestuck/util/StartingModusManager.java
amiwyn: I dont see the immediate need for it right now

src/main/java/com/mraof/minestuck/data/MinestuckStartingModusProvider.java
```diff
@@ -0,0 +1,65 @@
...
+		LinkedList<ModusType<?>> modusTypes = new LinkedList<>();
+		modusTypes.add(ModusTypes.STACK);
+		modusTypes.add(ModusTypes.QUEUE);
+		return modusTypes;
+	}
+	
+	@Override
+	public void run(HashCache pCache) throws IOException
+	{
+		Path path = this.generator.getOutputFolder().resolve(PATH);
```
kirderf1, line 43: Tested it out, and discovered that the path is missing parts, causing it to generate the file in "generated/resources/minestuck/config" folder instead of  "generated/resources/data/minestuck/minestuck/config"

src/main/java/com/mraof/minestuck/data/GeneratedGristCostConfigProvider.java
```diff
@@ -41,6 +41,7 @@ protected void addEntries()
 		serializer(RecipeSerializer.SHAPED_RECIPE);
 		serializer(RecipeSerializer.SHAPELESS_RECIPE);
 		serializer(MSRecipeTypes.NON_MIRRORED);
+		type(RecipeType.SMITHING, SmithingInterpreter.INSTANCE);
```
kirderf1, line 44: Now that you've merged changes, this line has become a duplicate. Please remove it and rerun "runData" to also remove changes to "grist_cost_geneeration_recipes".

src/main/java/com/mraof/minestuck/util/StartingModusManager.java
```diff
@@ -0,0 +1,96 @@
...
+			LOGGER.warn("Invalid json in data pack: '{}'", location.toString(), runtimeexception);
+		}
+		
+		return null;
+	}
+	
+	@Override
+	protected void apply(List<String> modusTypes, ResourceManager resourceManager, ProfilerFiller profiler)
+	{
+		startingModusTypes = modusTypes.stream().map(this::createModus).collect(Collectors.toList());
```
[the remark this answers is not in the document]
kirderf1: Not yet resolved as you're still returning null in one place. Also, I'd suggest using `Collections.emptyList()` for a purposefully empty list.

src/main/java/com/mraof/minestuck/util/StartingModusManager.java
```diff
@@ -0,0 +1,96 @@
...
+	private List<ModusType<?>> startingModusTypes;
+	
+	@Override
+	protected List<String> prepare(ResourceManager resourceManager, ProfilerFiller profiler)
+	{
+		ResourceLocation location = new ResourceLocation(Minestuck.MOD_ID, PATH);
+		if(!resourceManager.hasResource(location))
+			return null;
+		
+		try(Resource resource = resourceManager.getResource(location);
```
kirderf1, line 48: It may be nice to load lists across multiple namespaces and then merge them together into one. Doing so would better support potential addons adding new modus types by letting them add their types through their own data file without having to override minestucks data file. `RecipeGeneratedCostHandler` does exactly that for its data. Consider this a softer request that I'll likely do myself later if it doesn't get done in this PR.

kirderf1 (changes requested):

src/main/java/com/mraof/minestuck/data/MinestuckStartingModusProvider.java
amiwyn: I couldnt reproduce it

src/main/java/com/mraof/minestuck/data/GeneratedGristCostConfigProvider.java
kirderf1: The changes to data is still there. You have not yet run "runData" and committed the data file changes. If you've run "runData", but the generated files aren't updating as they should, it might be that the datagen cache doesn't match the actual generated files. To clear the cache, you can just delete the cache folder in the "generated" folder.

src/main/java/com/mraof/minestuck/data/MinestuckStartingModusProvider.java
kirderf1: Looks like you fixed it in the most recent commit. (Gonna test it tonight and see if it does what it should)
If you couldn't reproduce it before, it might be due to the cache problem described in the other review comment.

kirderf1 (approved):

PR #360, opened 2022-07-25
Use DeferredRegister in MSBlocks

Dweblenod:

PR #362, opened 2022-08-07
Use DeferredRegister in MSItems

Dweblenod:
A continuation of #360 that results in a switch over to using the DeferredRegister method of getting items


src/main/java/com/mraof/minestuck/data/BoondollarPricingProvider.java
```diff
@@ -44,112 +45,112 @@ public BoondollarPricingProvider(DataGenerator dataGenerator, String modid)
...
-		add(ZILLIUM_SKITTLES, 250, 300);
+		add(ONION.get(), 9, 15);
+		add(JAR_OF_BUGS.get(), 12, 18);
+		add(BUG_ON_A_STICK.get(), 4, 6);
+		add(CONE_OF_FLIES.get(), 4, 6);
+		add(GRASSHOPPER.get(), 90, 110);
+		add(SALAD.get(), 10, 14);
+		add(CHOCOLATE_BEETLE.get(), 30, 35);
+		add(DESERT_FRUIT.get(), 2, 6);
+		add(MSBlocks.GLOWING_MUSHROOM.get(), 10, 15);
```
kirderf1, line 56: Since MSItems now also hold references to block items, I'd suggest removing the static import of `MSBlocks`, and instead usie the already imported references from `MSItems` instead

src/main/java/com/mraof/minestuck/data/MinestuckEnUsLanguageProvider.java
```diff
@@ -494,649 +494,649 @@ protected void addTranslations()
...
-		addExtra(MSItems.CASSETTE_FAR, "desc", "C418 - far");
-		add(MSItems.CASSETTE_MALL, "Cassette");
-		addExtra(MSItems.CASSETTE_MALL, "desc", "C418 - mall");
-		add(MSItems.CASSETTE_EMISSARY, "Cassette");
-		addExtra(MSItems.CASSETTE_EMISSARY, "desc", "Catboss - Emissary of Dance");
-		add(MSItems.CASSETTE_DANCE_STAB, "Cassette");
-		addExtra(MSItems.CASSETTE_DANCE_STAB, "desc", "Catboss - Dance-Stab-Dance");
-		add(MSItems.CASSETTE_RETRO_BATTLE, "Cassette");
-		addExtra(MSItems.CASSETTE_RETRO_BATTLE, "desc", "SinFrog - Retro Battle Theme");
+		add(MSItems.CLAW_HAMMER.get(), "Claw Hammer");
```
kirderf1, line 497: Use the supplier version of these functions instead.

src/main/java/com/mraof/minestuck/effects/MSEffects.java
```diff
@@ -11,7 +11,7 @@
  */
 public class MSEffects
 {
-	public static final DeferredRegister<MobEffect> REGISTER = DeferredRegister.create(ForgeRegistries.MOB_EFFECTS, Minestuck.MOD_ID);
+	public static final DeferredRegister<MobEffect> EFFECTS = DeferredRegister.create(ForgeRegistries.MOB_EFFECTS, Minestuck.MOD_ID);
```
kirderf1, line 14: Why was this rename made?

src/main/java/com/mraof/minestuck/inventory/MiniTotemLatheContainer.java
```diff
@@ -43,9 +43,9 @@ public MiniTotemLatheContainer(MenuType<? extends MiniTotemLatheContainer> type,
 		
 		assertItemHandlerSize(inventory, 4);
 		
-		addSlot(new InputSlot(inventory, 0, CARD_1_X, CARD_1_Y, MSItems.CAPTCHA_CARD));
-		addSlot(new InputSlot(inventory, 1, CARD_2_X, CARD_2_Y, MSItems.CAPTCHA_CARD));
-		addSlot(new InputSlot(inventory, 2, DOWEL_X, DOWEL_Y, MSBlocks.CRUXITE_DOWEL.get().asItem()));
+		addSlot(new InputSlot(inventory, 0, CARD_1_X, CARD_1_Y, MSItems.CAPTCHA_CARD.get()));
+		addSlot(new InputSlot(inventory, 1, CARD_2_X, CARD_2_Y, MSItems.CAPTCHA_CARD.get()));
+		addSlot(new InputSlot(inventory, 2, DOWEL_X, DOWEL_Y, MSItems.CRUXITE_DOWEL.get())); //was MSBlocks.CRUXITE_DOWEL.get().asItem()
```
kirderf1, line 48: The change is fine. You can remove the comment.

src/main/java/com/mraof/minestuck/item/crafting/alchemy/GristTypes.java
```diff
@@ -19,27 +19,27 @@
...
+	public static final RegistryObject<GristType> DIAMOND = GRIST_TYPES.register("diamond", () -> new GristType(new GristType.Properties(0.2F, 5).spawnsFor(UNCOMMON, ANY).candy(() -> MSItems.DIAMOND_MINT.get()).secondary(GristTypes.GOLD)));
+	public static final RegistryObject<GristType> GOLD = GRIST_TYPES.register("gold", () -> new GristType(new GristType.Properties(0.2F, 5).spawnsFor(UNCOMMON, ANY).candy(() -> MSItems.GOLD_CANDY_RIBBON.get()).secondary(GristTypes.URANIUM)));
+	public static final RegistryObject<GristType> URANIUM = GRIST_TYPES.register("uranium", () -> new GristType(new GristType.Properties(0.2F, 5).spawnsFor(UNCOMMON, ANY).candy(() -> MSItems.URANIUM_GUMMY_BEAR.get()).secondary(GristTypes.DIAMOND)));
+	public static final RegistryObject<GristType> ARTIFACT = GRIST_TYPES.register("artifact", () -> new GristType(new GristType.Properties(0.1F, 1).candy(() -> MSItems.ARTIFACT_WARHEAD.get())));
+	public static final RegistryObject<GristType> ZILLIUM = GRIST_TYPES.register("zillium", () -> new GristType(new GristType.Properties(0.0F, 10).candy(() -> MSItems.ZILLIUM_SKITTLES.get()).notUnderlingType()));
```
kirderf1, line 42: `candy()` accepts suppliers, so you can pass in the registry objects directly.

src/main/java/com/mraof/minestuck/tileentity/machine/MiniTotemLatheTileEntity.java
```diff
@@ -44,7 +44,7 @@ public MiniTotemLatheTileEntity(BlockPos pos, BlockState state)
 	@Override
 	protected ItemStackHandler createItemHandler()
 	{
-		return new CustomHandler(4, (index, stack) -> (index == 0 || index == 1) && stack.getItem() == MSItems.CAPTCHA_CARD || index == 2 && stack.getItem() == MSBlocks.CRUXITE_DOWEL.get().asItem());
+		return new CustomHandler(4, (index, stack) -> (index == 0 || index == 1) && stack.getItem() == MSItems.CAPTCHA_CARD.get() || index == 2 && stack.getItem() == MSItems.CRUXITE_DOWEL.get()); //was MSBlocks.CRUXITE_DOWEL.get().asItem()
```
kirderf1, line 47: The change is fine. You can remove the comment.

src/main/java/com/mraof/minestuck/world/gen/feature/tree/EndTreeFeature.java
```diff
@@ -33,7 +33,7 @@ public boolean place(FeaturePlaceContext<NoneFeatureConfiguration> context)
 		WorldGenLevel level = context.level();
 		BlockPos position = context.origin();
 		BlockPos soilPos = position.below();
-		if(level.getBlockState(soilPos).canSustainPlant(level, soilPos, Direction.UP, MSBlocks.END_SAPLING.get()))
+		if(level.getBlockState(soilPos).canSustainPlant(level, soilPos, Direction.UP, (IPlantable) MSBlocks.END_SAPLING.get()))
```
kirderf1, line 36: This cast should not be needed.

src/main/java/com/mraof/minestuck/item/MSItems.java
```diff
@@ -30,1327 +31,990 @@
...
+	public static final RegistryObject<Item> FROG = ITEMS.register("frog", () -> new FrogItem(new Item.Properties().stacksTo(1).tab(MSItemGroup.LANDS)));
+	public static final RegistryObject<Item> CARVING_TOOL = ITEMS.register("carving_tool", () -> new Item(new Item.Properties().stacksTo(1).tab(MSItemGroup.LANDS)));
+	public static final RegistryObject<MSArmorItem> CRUMPLY_HAT = ITEMS.register("crumply_hat", () -> new MSArmorItem(MSItemTypes.CLOTH_ARMOR, EquipmentSlot.HEAD, new Item.Properties().stacksTo(1).tab(MSItemGroup.LANDS)));
+	public static final RegistryObject<Item> STONE_EYEBALLS = ITEMS.register("stone_eyeballs", () -> new Item(new Item.Properties().tab(MSItemGroup.LANDS)));
+	public static final RegistryObject<Item> STONE_SLAB = ITEMS.register("stone_slab", () -> new StoneTabletItem(MSBlocks.STONE_SLAB.get(), new Item.Properties().tab(MSItemGroup.LANDS)));
+	public static /*final*/ RegistryObject<Item> SHOP_POSTER = ITEMS.register("shop_poster", () -> new HangingItem(ShopPosterEntity::new, new Item.Properties().stacksTo(1).tab(MSItemGroup.LANDS)));
```
kirderf1, line 445: This should be final.

src/main/java/com/mraof/minestuck/item/MSItems.java
```diff
@@ -30,1327 +31,990 @@
...
 /**
- * This class contains all non-ItemBlock items that minestuck adds,
- * and is responsible for initializing and registering these.
+ * This class initializes and registers all items and blockitems in the mod. Utilizes a DeferredRegister, so utilizations of items from this class will often require .get()
  */
-@ObjectHolder(Minestuck.MOD_ID)
 @Mod.EventBusSubscriber(modid = Minestuck.MOD_ID, bus = Mod.EventBusSubscriber.Bus.MOD)
 public class MSItems
 {
+	//TODO removed @ObjectHolder(Minestuck.MOD_ID), ensure that was okay to do
```
kirderf1, line 47: It should be fine, as all fields are now initialized with registry objects from the deferred register.

src/main/java/com/mraof/minestuck/item/MSItems.java
```diff
@@ -30,1327 +31,990 @@
...
+	public static final RegistryObject<BlockItem> RAINBOW_WOOD = registerBlockItem(MSBlocks.RAINBOW_WOOD, MSItemGroup.LANDS);
+	public static final RegistryObject<BlockItem> END_WOOD = registerBlockItem(MSBlocks.END_WOOD, MSItemGroup.LANDS);
+	public static final RegistryObject<BlockItem> VINE_WOOD = registerBlockItem(MSBlocks.VINE_WOOD, MSItemGroup.LANDS);
+	public static final RegistryObject<BlockItem> FLOWERY_VINE_WOOD = registerBlockItem(MSBlocks.FLOWERY_VINE_WOOD, MSItemGroup.LANDS);
+	public static final RegistryObject<BlockItem> DEAD_WOOD = registerBlockItem(MSBlocks.DEAD_WOOD, MSItemGroup.LANDS);
+	public static final RegistryObject<BlockItem> PETRIFIED_WOOD = registerBlockItem(MSBlocks.PETRIFIED_WOOD, MSItemGroup.LANDS);
+	public static final RegistryObject<BlockItem> GLOWING_PLANKS = registerBlockItem(MSBlocks.GLOWING_PLANKS, MSItemGroup.LANDS);
+	public static final RegistryObject<BlockItem> FROST_PLANKS = registerBlockItem(MSBlocks.FROST_PLANKS, MSItemGroup.LANDS);
+	public static final RegistryObject<BlockItem> RAINBOW_PLANKS = registerBlockItem(MSBlocks.RAINBOW_PLANKS, MSItemGr
```
kirderf1, line 994: Use the variable from the lambda when creating the block item (`block1`)

kirderf1, line 1002: Use the variable from the lambda when creating the block item (`block1`)

kirderf1, line 1010: Use the variable from the lambda when creating the block item (`block1`)

kirderf1, line 1018: `getRegistryName()` is not called here, so the comment does not make much sense.

kirderf1 (changes requested):

src/main/java/com/mraof/minestuck/effects/MSEffects.java
Dweblenod: I know its not directly applicable to the PR but it seemed close enough related to the PR(since the Items register was named to ITEMS) that it would be fine to do so without created a PR just for the task

src/main/java/com/mraof/minestuck/item/MSItems.java
Dweblenod: Deleted now. However there was a near identical TODO i made for MSBlocks. Should that one be removed?

src/main/java/com/mraof/minestuck/item/MSItems.java
Dweblenod: Would you mind sharing how the line should look with the lambda? Im confused why its being applied to these wood block items

Dweblenod: Ah at least from my view of github the comment was applied to the wrong location. I assume this refers to the helper functions at the bottom so I will change those lambda parameter names to look like the following
`return registerBlockItem(block, block1 -> new BlockItem(block1, properties));`

Dweblenod: Understood, I wanted to add a comment to explain why we were letting it go without being checked. Would the following make sense to you?
`//assumed location() will occur without fail as it works through DeferredRegistry`

src/main/java/com/mraof/minestuck/item/MSItems.java
kirderf1: Yeah

src/main/java/com/mraof/minestuck/item/MSItems.java
kirderf1: Not really, since `location()` should give a non-null result regardless of if it came from the deferred registry or not.
However, `getKey()` has a nullable annotation, while we assume that it is not null.

src/main/java/com/mraof/minestuck/effects/MSEffects.java
kirderf1: I suppose I should've phrased it about the wider naming scheme instead of a specific naming instance.

At this time, I personally think that deferred register fields should all be called `REGISTER`. (As long as it is in situations such as these where only one specific deferred register is used within one appropriately named class)

Dweblenod: Will make the change then! Fine going with that approach. I saw them get named individually in javadoc for a relevant class so I took that approach

kirderf1: It does make sense to name them like that in the javadoc to distinguish between different deferred registers in the same class, or if the registry name is not already present in the class name.
So I think that naming scheme is more relevant for a small mods without dedicated classes for each registry.

src/main/java/com/mraof/minestuck/item/MSItems.java
Dweblenod: Is what I put now acceptable?

KingBalDoro: idk i can't read i'm too drunk

src/main/java/com/mraof/minestuck/fluid/MSFluids.java
```diff
@@ -27,25 +27,25 @@
...
+	public static final ForgeFlowingFluid.Properties BRAIN_JUICE_PROPERTIES = new ForgeFlowingFluid.Properties(BRAIN_JUICE, FLOWING_BRAIN_JUICE, FluidAttributes.builder(STILL_BRAIN_JUICE_TEXTURE, FLOWING_BRAIN_JUICE_TEXTURE)).bucket(() -> MSItems.BRAIN_JUICE_BUCKET.get()).block(() -> MSBlocks.BRAIN_JUICE.get()).explosionResistance(100F);
+	public static final ForgeFlowingFluid.Properties WATER_COLORS_PROPERTIES = new ForgeFlowingFluid.Properties(WATER_COLORS, FLOWING_WATER_COLORS, FluidAttributes.builder(STILL_WATER_COLORS_TEXTURE, FLOWING_WATER_COLORS_TEXTURE)).canMultiply().bucket(() -> MSItems.WATER_COLORS_BUCKET.get()).block(() -> MSBlocks.WATER_COLORS.get()).explosionResistance(100F);
+	public static final ForgeFlowingFluid.Properties ENDER_PROPERTIES = new ForgeFlowingFluid.Properties(ENDER, FLOWING_ENDER, FluidAttributes.builder(STILL_ENDER_TEXTURE, FLOWING_ENDER_TEXTURE)).bucket(() -> MSItems.ENDER_BUCKET.get()).block(() -> MSBlocks.ENDER.get()).slopeFindDistance(2).levelDecreasePerBlock(2).explosionResistance(100F);
```
kirderf1, line 50: Here are some other functions that take suppliers. Can you do the same thing with `block()` and `bucket()` that you did with grist candy?

kirderf1 (commented):
Looks good and ready. Just one minor change that could be done.

src/main/java/com/mraof/minestuck/fluid/MSFluids.java
Dweblenod: finished!

PR #363, opened 2022-08-21
Adding Music Weapons

LunyaticCat:


src/main/java/com/mraof/minestuck/item/weapon/MusicPlayerItem.java
```diff
@@ -0,0 +1,146 @@
...
+import net.minecraftforge.network.NetworkHooks;
+import org.jetbrains.annotations.Nullable;
+
+import java.util.Random;
+
+public class MusicPlayerItem extends WeaponItem
+{
+	private MobEffect effect = null;
+	private float applyingChance = 0.0F;
+	int duration = 0;
```
kirderf1, line 41: You should not be using mutable fields on items. Use some different means for passing these around.

src/main/java/com/mraof/minestuck/item/weapon/MusicPlayerItem.java
```diff
@@ -0,0 +1,146 @@
...
+	
+	private static IItemHandler getItemStackHandlerMusicPlayer(ItemStack itemStack)
+	{
+		return itemStack.getCapability(CapabilityItemHandler.ITEM_HANDLER_CAPABILITY).orElseThrow(() ->
+				new IllegalArgumentException("ItemStack " + itemStack + " is not an item handler"));
```
kirderf1, line 46: The error message is misleading. Something like
`"Expected an item handler for the music player item, but " + itemStack + " does not expose an item handler."`
should be both more detailed and more accurate.

src/main/java/com/mraof/minestuck/item/weapon/MusicPlayerItem.java
```diff
@@ -0,0 +1,146 @@
...
+			attackerLuckValue = 0.0D;
+		
+		if(targetLuck != null)
+			targetLuckValue = targetLuck.getValue();
+		else
+			targetLuckValue = 0.0D;
+		
+		chanceToHit = chanceToHit - (attackerLuckValue / 10) + (targetLuckValue / 10);
+		Debug.debug("chanceToHit: " + chanceToHit + " applyingChance: " + applyingChance);
+		Debug.debug("attackerLuckValue: " + attackerLuckValue + " targetLuckValue: " + targetLuckValue);
```
kirderf1, line 124: Don't use `Debug`. Use `System.out.println()` for debugging purposes, or a class-specific logger for long-term purposes.

src/main/java/com/mraof/minestuck/inventory/musicplayer/MusicPlayerContainer.java
```diff
@@ -0,0 +1,67 @@
...
+	
+	public MusicPlayerContainer(int windowId, Inventory playerInventory)
+	{
+		this(windowId, playerInventory, new ItemStackHandlerMusicPlayer(1));
+	}
+	
+	@Override
+	public boolean stillValid(Player playerIn)
+	{
+		return true;
```
kirderf1, line 37: I realize that you may need to hold on to the lazy optional to some degree in order to check if the capability still exists, instead of potentially holding on to an item handler that may become invalidated without the container knowing it. Probably worthwhile to check on some example elsewhere in the modding community on what might be needed.

src/main/java/com/mraof/minestuck/inventory/musicplayer/CapabilityProviderMusicPlayer.java
```diff
@@ -0,0 +1,58 @@
...
+import net.minecraftforge.common.capabilities.Capability;
+import net.minecraftforge.common.capabilities.ICapabilityProvider;
+import net.minecraftforge.common.util.INBTSerializable;
+import net.minecraftforge.common.util.LazyOptional;
+import net.minecraftforge.items.CapabilityItemHandler;
+import net.minecraftforge.items.IItemHandler;
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+
+public class CapabilityProviderMusicPlayer implements ICapabilityProvider, INBTSerializable<CompoundTag>
```
kirderf1, line 14: Shouldn't `MusicPlayer` be a prefix here?

src/main/java/com/mraof/minestuck/inventory/captchalogue/CaptchaDeckContainer.java
```diff
@@ -34,12 +36,8 @@ public void setContainerItem(ItemStack stack)
 		inventory.setItem(0, stack);
 	}
 	
-	private void addSlots(Inventory playerInventory)
+	private void addSlots()
```
kirderf1, line 39: It doesn't seem accurate to move the player inventory slots from this function directly into the constructor.
I'd suggest moving it back by calling `ContainerHelper.addPlayerInventorySlots()` in `addSlots()`, or alternatively to inline the `addSlots()` helper function.

src/main/java/com/mraof/minestuck/data/recipe/MinestuckCombinationsProvider.java
```diff
@@ -565,18 +566,25 @@ protected void buildCraftingRecipes(Consumer<FinishedRecipe> consumer)
...
+		CombinationRecipeBuilder.of(MSItems.CASSETTE_PIGSTEP).input(Items.MUSIC_DISC_PIGSTEP).or().namedInput(MSBlocks.CASSETTE_PLAYER.get()).build(consumer);
 		CombinationRecipeBuilder.of(MSItems.CASSETTE_RETRO_BATTLE).input(MSItems.MUSIC_DISC_RETRO_BATTLE).or().namedInput(MSBlocks.CASSETTE_PLAYER.get()).build(consumer);
+		CombinationRecipeBuilder.of(MSItems.CASSETTE_STAL).input(Items.MUSIC_DISC_STAL).or().namedInput(MSBlocks.CASSETTE_PLAYER.get()).build(consumer);
+		CombinationRecipeBuilder.of(MSItems.CASSETTE_STRAD).input(Items.MUSIC_DISC_STRAD).or().namedInput(MSBlocks.CASSETTE_PLAYER.get()).build(consumer);
+		CombinationRecipeBuilder.of(MSItems.CASSETTE_WAIT).input(Items.MUSIC_DISC_WAIT).or().namedInput(MSBlocks.CASSETTE_PLAYER.get()).build(consumer);
+		CombinationRecipeBuilder.of(MSItems.CASSETTE_WARD).input(Items.MUSIC_DISC_WARD).or().namedInput(MSBlocks.CASSETTE_PLAYER.get()).build(consumer);
 		
-		CombinationRecipeBuilder.of(MSBlocks.BLACK_CHESS_DIRT.get()).input(Items.DIRT).and().input(Items.BLACK_DYE).build(consumer);
+    CombinationRecipeBuilder.of(MSBlocks.BLACK_CHESS_DIRT.get()).input(Items.DIRT).and().input(Items.BLACK_DYE).build(consumer);
+
```
kirderf1, line 587: Looks like some unintended changes here that you can revert.

src/main/java/com/mraof/minestuck/client/sounds/MusicPlayerOnPlayerSoundInstance.java
```diff
@@ -0,0 +1,37 @@
...
+import com.mraof.minestuck.inventory.musicplayer.IMusicPlaying;
+import net.minecraft.client.resources.sounds.AbstractTickableSoundInstance;
+import net.minecraft.sounds.SoundSource;
+import net.minecraft.world.entity.player.Player;
+
+
+public class MusicPlayerOnPlayerSoundInstance extends AbstractTickableSoundInstance
+{
+	private final Player player;
+	private final IMusicPlaying soundEvent;
```
kirderf1, line 12: You should not be holding on to an `IMusicPlaying` instance like this. You should hold on to the lazy optional of it instead.

src/main/java/com/mraof/minestuck/block/EnumCassetteType.java
```diff
@@ -1,24 +1,57 @@
...
 	
 	@Override
 	public String getSerializedName()
 	{
 		return name().toLowerCase();
 	}
+	
+	public static SoundEvent getSoundEvent(EnumCassetteType type)
+	{
+		return switch(type)
```
kirderf1, line 36: It should be appropriate to define a field on the enum instead of this switch block.

kirderf1 (changes requested):
As different parts are completed down the line, I'd like to see some javadocs that describe the function and purpose behind the different new classes.

src/main/java/com/mraof/minestuck/inventory/musicplayer/CapabilityProviderMusicPlayer.java
```diff
@@ -0,0 +1,58 @@
...
+	private final LazyOptional<IMusicPlaying> lazyInitialisionSupplierMusicPlaying = LazyOptional.of(this::getCachedMusicPlaying);
+	private ItemStackHandlerMusicPlayer itemStackHandlerMusicPlayer;
+	private MusicPlaying musicPlaying;
+	
+	@NotNull
+	@Override
+	public <T> LazyOptional<T> getCapability(@NotNull Capability<T> cap, @Nullable Direction side)
+	{
+		if(CapabilityItemHandler.ITEM_HANDLER_CAPABILITY == cap)
+			return (LazyOptional<T>) (lazyInitialisionSupplierItemHandler);
```
kirderf1, line 26: A slightly cleaner way of doing the cast is to call `lazyOptional.cast()` instead of doing the cast directly.

src/main/java/com/mraof/minestuck/item/weapon/MusicPlayerItem.java
LunyaticCat: Ok, I changed both the item handler and the music playing capabilities.

src/main/java/com/mraof/minestuck/item/weapon/MusicPlayerItem.java
LunyaticCat: Ok

src/main/java/com/mraof/minestuck/inventory/musicplayer/CapabilityProviderMusicPlayer.java
LunyaticCat: Yes, im changing it

src/main/java/com/mraof/minestuck/inventory/musicplayer/CapabilityProviderMusicPlayer.java
LunyaticCat: oh ok !

src/main/java/com/mraof/minestuck/client/sounds/MusicPlayerOnPlayerSoundInstance.java
LunyaticCat: How can I access get`CurrentMusic()` from something like `LazyOptional<IMusicPlaying> soundEvent` then?

src/main/java/com/mraof/minestuck/data/recipe/MinestuckCombinationsProvider.java
LunyaticCat: that should be fixed

src/main/java/com/mraof/minestuck/inventory/musicplayer/CapabilityProviderMusicPlayer.java
```diff
@@ -0,0 +1,58 @@
...
+import net.minecraftforge.common.util.LazyOptional;
+import net.minecraftforge.items.CapabilityItemHandler;
+import net.minecraftforge.items.IItemHandler;
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+
+public class CapabilityProviderMusicPlayer implements ICapabilityProvider, INBTSerializable<CompoundTag>
+{
+	private final LazyOptional<IItemHandler> lazyInitialisionSupplierItemHandler = LazyOptional.of(this::getCachedInventory);
+	private final LazyOptional<IMusicPlaying> lazyInitialisionSupplierMusicPlaying = LazyOptional.of(this::getCachedMusicPlaying);
```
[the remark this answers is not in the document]
LunyaticCat: I changed it to `lazyInitSupplierItemHandler` if that's fine

src/main/java/com/mraof/minestuck/inventory/captchalogue/CaptchaDeckContainer.java
LunyaticCat: I don't remember touching the captchalogue class

LunyaticCat: oh right the helper function !

src/main/java/com/mraof/minestuck/block/EnumCassetteType.java
LunyaticCat: Ok, it should be right now

src/main/java/com/mraof/minestuck/inventory/musicplayer/ItemStackHandlerMusicPlayer.java
```diff
@@ -0,0 +1,20 @@
+package com.mraof.minestuck.inventory.musicplayer;
+
+import com.mraof.minestuck.util.MSTags;
+import net.minecraft.world.item.ItemStack;
+import net.minecraftforge.items.ItemStackHandler;
+
+import javax.annotation.Nonnull;
+
+public class ItemStackHandlerMusicPlayer extends ItemStackHandler
```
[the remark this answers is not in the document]
LunyaticCat: it's not only used in the capability Provider but also in the container
I don't know if I would be able to access it if I make it anonymous.
Plus, making it generic as it is right now will allow other devs to understand better how to make their own.

src/main/java/com/mraof/minestuck/inventory/captchalogue/CaptchaDeckContainer.java
LunyaticCat:
```java
private void addSlots(Inventory playerInventory)
	{
		ContainerHelper.addPlayerInventorySlots(this::addSlot, 9, 63, playerInventory);
		addSlot(new Slot(this.inventory, 0, 81, 32)
		{
			@Override
			public boolean mayPlace(ItemStack stack)
			{
				return ModusTypes.getTypeFromItem(stack.getItem()) != null || stack.getItem().equals(MSItems.CAPTCHA_CARD) && !AlchemyHelper.isPunchedCard(stack);
			}
		});
	}
	```
	Would this work for you?

File: src/main/java/com/mraof/minestuck/inventory/musicplayer/ItemStackHandlerMusicPlayer.java (hunk 12 above)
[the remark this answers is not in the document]
Comment (kirderf1) [reply src/main/java/com/mraof/minestuck/inventory/musicplayer/ItemStackHandlerMusicPlayer.java:9]: > it's not only used in the capability Provider but also in the container

Ah, good point. I'll pull back this suggestion for now then.

File: src/main/java/com/mraof/minestuck/inventory/captchalogue/CaptchaDeckContainer.java (hunk 6 above)
Comment (kirderf1) [reply src/main/java/com/mraof/minestuck/inventory/captchalogue/CaptchaDeckContainer.java:39]: Yep, that's what I mean by
> moving it back by calling `ContainerHelper.addPlayerInventorySlots()` in `addSlots()`

File: src/main/java/com/mraof/minestuck/data/recipe/MinestuckCombinationsProvider.java (hunk 7 above)
Comment (kirderf1) [reply src/main/java/com/mraof/minestuck/data/recipe/MinestuckCombinationsProvider.java:587]: The change is still there.

File: src/main/java/com/mraof/minestuck/client/sounds/MusicPlayerOnPlayerSoundInstance.java (hunk 8 above)
Comment (kirderf1) [reply src/main/java/com/mraof/minestuck/client/sounds/MusicPlayerOnPlayerSoundInstance.java:12]: The simplest way of doing it is calling `lazyOptional.getValue()`, and it'll return the `IMusicPlaying` instance if it is present, or null if it's not present (and I assume that the sound instance should be stopped when the capability instance is no longer present)

File: src/main/java/com/mraof/minestuck/inventory/musicplayer/ItemHandlerCapabilityProvider.java [hunk 13]
```diff
@@ -0,0 +1,47 @@
...
+	@NotNull
+	@Override
+	public <T> LazyOptional<T> getCapability(@NotNull Capability<T> cap, @Nullable Direction side)
+	{
+		if(CapabilityItemHandler.ITEM_HANDLER_CAPABILITY == cap)
+			return (lazyInitSupplierItemHandler).cast();
+		return LazyOptional.empty();
+	}
+	
+	private @NotNull ItemStackHandlerMusicPlayer getCachedInventory() {
```
kirderf1, line 28: I'm wondering if it's worthwhile to go the extra step to lazily initialize the item handler (which as far as I can tell is done for performance only).
I'll have to get back to this

src/main/java/com/mraof/minestuck/inventory/musicplayer/ItemStackHandlerMusicPlayer.java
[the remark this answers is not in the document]
kirderf1: Acutally new suggestion: `CassetteSlotItemHandler`, `CassetteHolderItemHandler` or just `CassetteItemHandler`

src/main/java/com/mraof/minestuck/inventory/musicplayer/MusicPlayerContainer.java
```diff
@@ -0,0 +1,78 @@
...
+import net.minecraft.world.entity.player.Player;
+import net.minecraft.world.inventory.AbstractContainerMenu;
+import net.minecraft.world.inventory.Slot;
+import net.minecraft.world.item.ItemStack;
+import net.minecraftforge.items.IItemHandler;
+import net.minecraftforge.items.SlotItemHandler;
+
+import javax.annotation.Nonnull;
+
+public class MusicPlayerContainer extends AbstractContainerMenu
```
kirderf1, line 17: Rename to `CassetteContainerMenu` or `CassetteHolderMenu`

src/main/java/com/mraof/minestuck/item/MSItems.java
```diff
@@ -64,6 +64,7 @@
...
 	public static final RegistryObject<Item> SCARLET_ZILLYHOO = REGISTER.register("scarlet_zillyhoo", () -> new WeaponItem(new WeaponItem.Builder(MSItemTypes.CORUNDUM_TIER, 9, -3.2F).efficiency(4.0F).set(MSItemTypes.HAMMER_TOOL).add(OnHitEffect.setOnFire(50)), new Item.Properties().tab(MSItemGroup.WEAPONS).rarity(Rarity.RARE)));
 	public static final RegistryObject<Item> MWRTHWL = REGISTER.register("mwrthwl", () -> new WeaponItem(new WeaponItem.Builder(MSItemTypes.WELSH_TIER, 8, -3.2F).efficiency(4.0F).set(MSItemTypes.HAMMER_TOOL), new Item.Properties().tab(MSItemGroup.WEAPONS).rarity(Rarity.EPIC)));
+	public static final RegistryObject<Item> MUSIC_HAMMER = REGISTER.register("music_hammer", () -> new MusicPlayerItem(new WeaponItem.Builder(Tiers.DIAMOND, 7, -3.0F).efficiency(5.0F).set(MSItemTypes.HAMMER_TOOL), new Item.Properties().tab(MSItemGroup.WEAPONS).rarity(Rarity.EPIC)));
```
kirderf1, line 67: I did not see a way to obtain this item in survival. What is the plan for the item on that front?

src/main/java/com/mraof/minestuck/inventory/musicplayer/IMusicPlaying.java
```diff
@@ -0,0 +1,15 @@
+package com.mraof.minestuck.inventory.musicplayer;
+
+import com.mraof.minestuck.block.EnumCassetteType;
+import net.minecraft.sounds.SoundEvent;
+import net.minecraft.world.item.ItemStack;
+
+public interface IMusicPlaying {
+	void setCassetteType(EnumCassetteType cassetteType);
+	
+	void setCurrentMusicPlayer(ItemStack itemStack);
```
kirderf1, line 10: The two setters always seem to be used together. Would it make sense to combine them?

src/main/java/com/mraof/minestuck/network/MusicPlayerPacket.java
```diff
@@ -0,0 +1,61 @@
...
+		EnumCassetteType cassetteType = EnumCassetteType.values()[buffer.readInt()];
+		
+		return new MusicPlayerPacket(entityID, cassetteType);
+	}
+	
+	@Override
+	public void execute()
+	{
+		Entity entity = Minecraft.getInstance().level.getEntity(entityID);
+		if(entity instanceof Player){
```
kirderf1, line 46: Make sure to format your code based on the style of the project. In Intellij, this can be done with the "Reformat Code" action.

src/main/java/com/mraof/minestuck/network/MusicPlayerPacket.java
```diff
@@ -0,0 +1,61 @@
...
+		int entityID = buffer.readInt(); //readInt spits out the values you gave to the PacketBuffer in encode in that order
+		EnumCassetteType cassetteType = EnumCassetteType.values()[buffer.readInt()];
+		
+		return new MusicPlayerPacket(entityID, cassetteType);
+	}
+	
+	@Override
+	public void execute()
+	{
+		Entity entity = Minecraft.getInstance().level.getEntity(entityID);
```
kirderf1, line 45: Referring to client-only classes from a common class might be a bad idea for dedicated servers. Can you move everything you're currently doing in the `execute()` function to a class in the client package? (a common class in this context meaning that the class is used both server and client side)

kirderf1 (changes requested):

src/main/java/com/mraof/minestuck/block/EnumCassetteType.java
```diff
@@ -1,24 +1,84 @@
...
+			false, false, false))),
+	STRAD(() -> SoundEvents.MUSIC_DISC_STRAD, () -> onHitEffect(new MobEffectInstance(MobEffects.UNLUCK, 200, 0),
+			0.10F)),
+	WAIT(() -> SoundEvents.MUSIC_DISC_WAIT, () -> onHitEffect(new MobEffectInstance(MobEffects.MOVEMENT_SLOWDOWN, 200, 0),
+			0.30F)),
+	WARD(() -> SoundEvents.MUSIC_DISC_WARD, () -> onHitEffect(new MobEffectInstance(MobEffects.GLOWING, 750, 0), 1F));
+	
+	private final Supplier<SoundEvent> soundEvent;
+	@Nullable
+	private final Supplier<EffectContainer> effectContainer;
```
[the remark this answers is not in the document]
LunyaticCat: done !

src/main/java/com/mraof/minestuck/client/sounds/MusicPlayerOnPlayerSoundInstance.java
```diff
@@ -0,0 +1,25 @@
...
+import net.minecraft.client.resources.sounds.EntityBoundSoundInstance;
+import net.minecraft.sounds.SoundEvent;
+import net.minecraft.sounds.SoundSource;
+import net.minecraft.world.entity.player.Player;
+
+import java.util.HashMap;
+import java.util.Map;
+
+
+public class MusicPlayerOnPlayerSoundInstance extends EntityBoundSoundInstance
```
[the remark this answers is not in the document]
LunyaticCat: it has the entitiesMap in it, I can't use it directly if im not moving the map elsewhere

src/main/java/com/mraof/minestuck/item/MSItems.java
LunyaticCat: this item is only here to test the MusicPlayerItem class. The designer teams having created sprites recently, I will replace it by their sword right now.

src/main/java/com/mraof/minestuck/client/sounds/PlayerMusicClientHandler.java
```diff
@@ -0,0 +1,43 @@
...
+import net.minecraft.sounds.SoundSource;
+import net.minecraft.world.entity.Entity;
+import net.minecraft.world.entity.player.Player;
+
+import java.util.HashMap;
+import java.util.Map;
+
+public class PlayerMusicClientHandler
+{
+	private static final Map<Integer, EntityBoundSoundInstance> entitiesMap = new HashMap<>();
```
kirderf1, line 16: I wonder if there needs to be a concern about data buildup in this map.

src/main/java/com/mraof/minestuck/inventory/MSMenuTypes.java
```diff
@@ -20,4 +21,5 @@
 	public static final RegistryObject<MenuType<CaptchaDeckMenu>> CAPTCHA_DECK = REGISTER.register("captcha_deck", () -> new MenuType<>(CaptchaDeckMenu::new));
 	public static final RegistryObject<MenuType<EditmodeMenu>> EDIT_MODE = REGISTER.register("edit_mode", () -> new MenuType<>(EditmodeMenu::new));
 	public static final RegistryObject<MenuType<ConsortMerchantMenu>> CONSORT_MERCHANT = REGISTER.register("consort_merchant", () -> new MenuType<>((IContainerFactory<ConsortMerchantMenu>) ConsortMerchantMenu::load));
+	public static final RegistryObject<MenuType<CassetteContainerMenu>> MUSIC_PLAYER = REGISTER.register("music_player", () ->new MenuType<>(CassetteContainerMenu::new));
```
kirderf1, line 24: It may be appropriate to rename the menu type according to the menu class name

src/main/java/com/mraof/minestuck/item/weapon/MusicPlayerItem.java
```diff
@@ -0,0 +1,177 @@
...
+	
+	@SubscribeEvent
+	public static void entityAttachCapabilitiesEvent(AttachCapabilitiesEvent<Entity> attachCapabilitiesEvent)
+	{
+		if(attachCapabilitiesEvent.getObject() instanceof Player)
+		{
+			attachCapabilitiesEvent.addCapability(new ResourceLocation(Minestuck.MOD_ID, "musicplaying"),
+					new MusicPlayingCapabilityProvider());
+		}
+	}
```
kirderf1, line 59: I'd suggest putting this in `CapabilityMusicPlaying` instead.

src/main/java/com/mraof/minestuck/client/gui/MusicPlayerScreen.java
```diff
@@ -0,0 +1,41 @@
...
+import com.mojang.blaze3d.systems.RenderSystem;
+import com.mojang.blaze3d.vertex.PoseStack;
+import com.mraof.minestuck.inventory.musicplayer.CassetteContainerMenu;
+import net.minecraft.client.gui.screens.inventory.AbstractContainerScreen;
+import net.minecraft.client.renderer.GameRenderer;
+import net.minecraft.network.chat.Component;
+import net.minecraft.resources.ResourceLocation;
+import net.minecraft.world.entity.player.Inventory;
+
+public class MusicPlayerScreen extends AbstractContainerScreen<CassetteContainerMenu>
```
kirderf1, line 12: Rename this class and the texture image based on the menu name.

kirderf1 (changes requested):

src/main/java/com/mraof/minestuck/block/EnumCassetteType.java
```diff
@@ -1,24 +1,84 @@
...
+	}
+	@Nonnull
+	public EffectContainer getEffectContainer()
+	{
+		return Objects.requireNonNull(effectContainer);
+	}
+	public record EffectContainer(Supplier<MobEffectInstance> effect, float applyingChance, boolean onHit)
+	{
+	}
+	private static EffectContainer userEffect(Supplier<MobEffectInstance> effect){
```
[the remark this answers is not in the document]
LunyaticCat: idea automatically format it that way with the file you gave
is there a way to fix it?

src/main/java/com/mraof/minestuck/item/weapon/MusicPlayerItem.java
```diff
@@ -0,0 +1,194 @@
...
+	@Override
+	public void inventoryTick(ItemStack stack, Level level, Entity entityIn, int itemSlot, boolean isSelected)
+	{
+		super.inventoryTick(stack, level, entityIn, itemSlot, isSelected);
+		if(stack.getItem() instanceof MusicPlayerItem)
+		{
+			ItemStack itemInMusicPlayer = getItemStackHandlerMusicPlayer(stack).getStackInSlot(0);
+			stack.getOrCreateTag().putBoolean("HasCassette", itemInMusicPlayer != ItemStack.EMPTY);
+		}
+	}
```
kirderf1, line 128: a) The instanceof check should be unnecessary since it should be assumed that the item of that stack is this item.
b) Don't use `stack != ItemStack.EMPTY`. Use `!stack.isEmpty()`.
c) Please try overriding `getShareTag()` instead.

kirderf1 (changes requested):
I'm not a bit fan of the higher-res texture, though I'm not sure if I should be the one to make a final say on that.

Dweblenod:
@kirderf1 certainly!

Dweblenod:
> I'm not a bit fan of the higher-res texture, though I'm not sure if I should be the one to make a final say on that.

As a matter of project hierarchy, the system is intended so that Riotmode has final say on art. If there is an issue with performance, implementation, or anything related to coding, then it could be different. We can discuss art direction with him if its a matter of concern

kirderf1:
> @kirderf1 certainly!

Alright, thanks!

Dweblenod (commented):
If you are looking for further ideas with this PR, what if the item tooltip lists the currently stored cassette?

Dweblenod:
This may have already been considered, but the cassette in the item is deleted if the item is destroyed(such as by losing all durability)

src/main/java/com/mraof/minestuck/block/EnumCassetteType.java
```diff
@@ -1,24 +1,88 @@
...
+			false, false, false))),
+	RETRO_BATTLE_THEME(MSSoundEvents.MUSIC_DISC_RETRO_BATTLE_THEME, userEffect(() -> new MobEffectInstance(MobEffects.DAMAGE_RESISTANCE, 100, 0,
+			false, false, false))),
+	STAL(() -> SoundEvents.MUSIC_DISC_STAL, userEffect(() -> new MobEffectInstance(MobEffects.JUMP, 100, 1,
+			false, false, false))),
+	STRAD(() -> SoundEvents.MUSIC_DISC_STRAD, onHitEffect(() -> new MobEffectInstance(MobEffects.UNLUCK, 200, 0),
+			0.10F)),
+	WAIT(() -> SoundEvents.MUSIC_DISC_WAIT, onHitEffect(() -> new MobEffectInstance(MobEffects.MOVEMENT_SLOWDOWN, 200, 0),
+			0.30F)),
+	WARD(() -> SoundEvents.MUSIC_DISC_WARD, onHitEffect(() -> new MobEffectInstance(MobEffects.GLOWING, 750, 0), 1F));
```
Dweblenod, line 48: Still trying to gauge the impact of these effects, but I have this to say for now. I think the following effects need some changes: 
-Blocks: Suggestion to change effect. The confusion/nausea effect can be actually disorienting to the human playing the game. There is a video settings slider that allows people to control the strength of the distortion, however its also not something people may be expecting to come up against, and if they do reduce the distortion than having it on the weapon doesnt accomplish much
-Dance Stab Dance: Suggestion to change effect. While the song is being played, the damage boost effects means the weapon essentially deals 13 damage which is more than even the top tier swords
-Emissary of Dance: Consider reducing the length of the poison damage. Another weapon which deals poison damage(Red Eye) has a duration of 140 which is less than half the 400 the cassette is currently set to
-Far: essentially permanent speed buff may be too strong
-Retro Battle Theme: essentially permanent damage resistance may be too strong

src/main/java/com/mraof/minestuck/item/MSItems.java
```diff
@@ -99,6 +98,7 @@
...
 	public static final RegistryObject<Item> SHATTER_BACON = REGISTER.register("shatter_bacon", () -> new WeaponItem(new WeaponItem.Builder(MSItemTypes.MEAT_TIER, 7, -2.4F).efficiency(5.0F).set(MSItemTypes.SWORD_TOOL).add(OnHitEffect.SWEEP).add(OnHitEffect.SORD_DROP), new Item.Properties().tab(MSItemGroup.WEAPONS).rarity(Rarity.UNCOMMON)));
 	public static final RegistryObject<Item> SUBTRACTSHUMIDIRE_ZOMORRODNEGATIVE = REGISTER.register("subtractshumidire_zomorrodnegative", () -> new WeaponItem(new WeaponItem.Builder(MSItemTypes.URANIUM_TIER, 6, -2.6F).efficiency(5.0F).set(MSItemTypes.SWORD_TOOL).add(OnHitEffect.SWEEP).add(OnHitEffect.setOnFire(30)), new Item.Properties().tab(MSItemGroup.WEAPONS)));
+	public static final RegistryObject<Item> MUSIC_SWORD = REGISTER.register("music_sword", () -> new MusicPlayerItem(new WeaponItem.Builder(Tiers.DIAMOND, 6, -2.4F).efficiency(5.0F).set(MSItemTypes.SWORD_TOOL).add(OnHitEffect.SWEEP), new Item.Properties().tab(MSItemGroup.WEAPONS).rarity(Rarity.EPIC)));
```
Dweblenod, line 101: Reduce the rarity. Currently rarity for alchemy items tries to consider the rarity of the ingredients used to make it and the quality of the weapon. Epic rarity is equivalent to the popular high tier weapons from canon. It matters since the rarity factors into the grist cost calculation

src/main/java/com/mraof/minestuck/item/weapon/MusicPlayerItem.java
```diff
@@ -0,0 +1,194 @@
...
+						packet = MusicPlayerPacket.createPacket(playerIn, cassette.cassetteID, volume, pitch);
+						musicPlayingCap.setMusicPlaying(musicPlayer, cassette.cassetteID);
+					} else
+					{
+						packet = MusicPlayerPacket.createPacket(playerIn, EnumCassetteType.NONE, 0, 0);
+						musicPlayingCap.setMusicPlaying(ItemStack.EMPTY, EnumCassetteType.NONE);
+					}
+					MSPacketHandler.sendToTrackingAndSelf(packet, playerIn);
+				}
+			}
```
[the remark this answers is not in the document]
kirderf1: Looked some more at the problem, and there seems to be two aspects to it:
1. If the music is stopped immediately after it was set to start, the sound engine doesn't seem to register the stop call.  This can happen if there's a packet for playing music immediately followed by another music packet for the same player. This is a problem in `PlayerMusicClientHandler` and need to be fixed there. That class believes that it stopped the music from the first packet during the second packet, but the sound engine didn't actually stop it from playing. Solving this may involve figuring out exactly why the sound engine didn't handle the call to stop the music. It's possible that the end solution may involve playing and stopping music in bulk during some tick event instead of one by one through the execute function of the packet.
2. We're getting a "play music" and a "stop music" packet in the same tick! This is probably because the first music player item doesn't block interaction with the second when used. That's because this function is not returning the proper result when some sort of interaction is attempted or succeeds. This part is the most important one to fix!

src/main/java/com/mraof/minestuck/item/weapon/MusicPlayerItem.java
LunyaticCat: how would you suggest I use getShareTag() instead of inventoryTick()?

src/main/java/com/mraof/minestuck/block/EnumCassetteType.java
```diff
@@ -1,24 +1,88 @@
...
+			0.1F)),
+	FAR(() -> SoundEvents.MUSIC_DISC_FAR, userEffect(() -> new MobEffectInstance(MobEffects.MOVEMENT_SPEED, 100, 0,
+			false, false, false))),
+	MALL(() -> SoundEvents.MUSIC_DISC_MALL, userEffect(() -> new MobEffectInstance(MobEffects.WATER_BREATHING, 100, 0,
+			false, false, false))),
+	MELLOHI(() -> SoundEvents.MUSIC_DISC_MELLOHI, onHitEffect(() -> new MobEffectInstance(MobEffects.LEVITATION, 60, 0),
+			0.20F)),
+	PIGSTEP(() -> SoundEvents.MUSIC_DISC_PIGSTEP, userEffect(() -> new MobEffectInstance(MobEffects.FIRE_RESISTANCE, 100, 0,
+			false, false, false))),
+	RETRO_BATTLE_THEME(MSSoundEvents.MUSIC_DISC_RETRO_BATTLE_THEME, userEffect(() -> new MobEffectInstance(MobEffects.HEALTH_BOOST, 100, 2,
```
Dweblenod, line 40: This health boost effect adds 6 hearts(12 health) to the player. I dont necessarily have an issue with it but I did want to know if it was intended

Dweblenod: This may complicate how the code works but it may be smarter to make health boosting a weapon attribute modifier instead of a potion effect. An issue with health boost is that every time it is reapplied, the player loses that additional health. Meaning that they are cut back down to their original health every ~3 seconds, which doesnt help them with healing and could be considered a waste of the players hunger

src/main/java/com/mraof/minestuck/item/weapon/MusicPlayerItem.java
```diff
@@ -0,0 +1,234 @@
...
+					MSPacketHandler.sendToTrackingAndSelf(packet, player);
+				}
+			}
+			
+			if(player.level.getGameTime() % 50 == 0 && musicPlayingCap.getCassetteType() != EnumCassetteType.NONE)
+			{
+				EnumCassetteType.EffectContainer effectContainer = musicPlayingCap.getCassetteType().getEffectContainer();
+				if(!effectContainer.onHit()) //Apply the cassette buff every 50 ticks, if there is any
+					player.addEffect(effectContainer.effect().get());
+			}
```
Dweblenod, line 177: I dont think its worth changing, but I am going to note that the weapon continues to give the user buffs(if the cassette does that) even after the music has stopped playing

LunyaticCat: I would have liked to loop the music, but the soundinstance class im using do not allow it.

src/main/java/com/mraof/minestuck/item/weapon/MusicPlayerItem.java
Dweblenod: `@Nullable
	@Override
	public CompoundTag getShareTag(ItemStack stack)
	{
		CompoundTag tag = stack.getTag();
		if(stack.getItem() instanceof MusicPlayerItem && tag != null)
		{
			ItemStack itemInMusicPlayer = getItemStackHandlerMusicPlayer(stack).getStackInSlot(0);
			tag.putBoolean("HasCassette", itemInMusicPlayer != ItemStack.EMPTY);
			
			return tag;
		}
		
		return super.getShareTag(stack);
	}`

LunyaticCat: I thought about that but that would not help me change the tag according to a change in the container.

src/main/java/com/mraof/minestuck/client/sounds/PlayerMusicClientHandler.java
```diff
@@ -0,0 +1,61 @@
...
+import net.minecraft.sounds.SoundSource;
+import net.minecraft.world.entity.Entity;
+import net.minecraft.world.entity.player.Player;
+
+import java.util.HashMap;
+import java.util.Map;
+
+/**
+ * Keep track of every player entities listening to the music.
+ * Create a sound instance on every client matching the music currently playing.
```
kirderf1, line 17:
```suggestion
 * Keeps track of sound instances for music being played by nearby players.
 * These sound instances are created client-side upon receiving a MusicPlayerPacket.
```

src/main/java/com/mraof/minestuck/inventory/musicplayer/IMusicPlaying.java
```diff
@@ -0,0 +1,21 @@
+package com.mraof.minestuck.inventory.musicplayer;
+
+import com.mraof.minestuck.block.EnumCassetteType;
+import com.mraof.minestuck.item.weapon.MusicPlayerItem;
+import net.minecraft.world.item.ItemStack;
+
+/**
+ * This capability interface is meant to keep track of the current music being played,
+ * as well as which music player is currently playing it.
```
kirderf1, line 9:
```suggestion
 * The music playing capability keeps track of
 * the music cassette currently played by the player that the capability is attached to,
 * as well as which music player item stack that is being used to play it.
 * This interface defines how to interact with this capability.
```

src/main/java/com/mraof/minestuck/item/weapon/MusicPlayerItem.java
```diff
@@ -0,0 +1,234 @@
...
+			{
+				MusicPlayerPacket packet = MusicPlayerPacket.createPacket(playerIn, EnumCassetteType.NONE, 0, 0);
+				musicPlayingCap.setMusicPlaying(ItemStack.EMPTY, EnumCassetteType.NONE);
+				MSPacketHandler.sendToTrackingAndSelf(packet, playerIn); //This will stop the music before opening the GUI
+				
+				NetworkHooks.openGui((ServerPlayer) playerIn, new SimpleMenuProvider((pContainerId, pInventory, pPlayer) ->
+						new CassetteContainerMenu(pContainerId, pInventory, itemStackHandlerMusicPlayer, musicPlayer),
+						new TranslatableComponent(TITLE)));
+			}
+			return InteractionResultHolder.success(musicPlayer);
```
kirderf1, line 139: If any interaction would be a success, you should return a success regardless of the [logical side](https://forge.gemwire.uk/wiki/Sides/1.18). With that change, you should also be using `InteractionResultHolder.sidedSuccess()` instead of `InterectionResultHolder.success()`.

src/main/java/com/mraof/minestuck/client/sounds/PlayerMusicClientHandler.java
```diff
@@ -0,0 +1,61 @@
...
+		checkEntitiesInMap();
+		
+		Entity entity = Minecraft.getInstance().level.getEntity(entityID);
+		if(entity instanceof Player)
+		{
+			SoundManager soundManager = Minecraft.getInstance().getSoundManager();
+			
+			if(entitiesMap.containsKey(entityID))
+			{
+				soundManager.stop(entitiesMap.remove(entityID));
```
kirderf1, line 37:
```suggestion
				//TODO: if the sound instance was started the same tick, it fails to stop properly
				soundManager.stop(entitiesMap.remove(entityID));
```
Leave a todo for problem 1 from a review comment on the item class.

src/main/java/com/mraof/minestuck/item/weapon/MusicPlayerItem.java
kirderf1: Spontaneous thought: should a cassette player be able to loop in the first place? At some point further down the line, we could try implementing cassettes so that they need to be rewound to play from the beginning.

src/main/java/com/mraof/minestuck/util/MSCapabilities.java
```diff
@@ -0,0 +1,44 @@
...
+	{
+	});
+	
+	public static void register(RegisterCapabilitiesEvent event)
+	{
+		event.register(IMusicPlaying.class);
+	}
+	
+	/**
+	 * Attach the MusicPlayingCapability to the player
```
kirderf1, line 31:
```suggestion
	 * Attaches a provider of the music playing capability to any player
```

src/main/java/com/mraof/minestuck/network/MusicPlayerPacket.java
```diff
@@ -0,0 +1,64 @@
...
+{
+	private final int entityID;
+	private final EnumCassetteType cassetteType;
+	private final float volume;
+	private final float pitch;
+	
+	/**
+	 * Create a packet to the client to start a music based on the cassetteType given.
+	 * If the cassetteType NONE is given, the previous music stop if there
+	 * was any, and no music is played.
```
kirderf1, line 18:
```suggestion
	 * Creates a packet to the client for starting a cassette music track played by a specified player, and stopping the previous music track played by that player if there was any.
	 * If the cassetteType NONE is given, no new track is played.
```

src/main/java/com/mraof/minestuck/item/weapon/MusicPlayerItem.java
kirderf1: `getShareTag()` is used if a change is detected in the item stack or in any serialized capability, so this should work to change the tag according to changes in the inventory/item handler. However, the instanceof check should not be needed, and you should return a tag with "HasCassette" set even if the item stack does not have a tag. Just create a new tag in that case. `CompoundTag tag = Objects.requireNonNullElseGet(stack.getTag(), CompoundTag::new);`

kirderf1 (changes requested):

src/main/java/com/mraof/minestuck/item/weapon/MusicPlayerItem.java
```diff
@@ -0,0 +1,224 @@
...
+ */
+
+@Mod.EventBusSubscriber(modid = Minestuck.MOD_ID)
+public class MusicPlayerItem extends WeaponItem
+{
+	public static final String TITLE = "minestuck.music_player";
+	private final float volume;
+	private final float pitch;
+	
+	private static IItemHandler getItemStackHandlerMusicPlayer(ItemStack itemStack)
```
kirderf1, line 66: I suggest renaming this to just `getItemHandler()`

src/main/java/com/mraof/minestuck/item/weapon/MusicPlayerItem.java
```diff
@@ -0,0 +1,224 @@
...
+		{
+			Player player = tickEvent.player;
+			IMusicPlaying musicPlayingCap = getMusicPlaying(player);
+			
+			if(!(player.getItemInHand(InteractionHand.MAIN_HAND) == musicPlayingCap.getCurrentMusicPlayer() ||
+					player.getItemInHand(InteractionHand.OFF_HAND) == musicPlayingCap.getCurrentMusicPlayer()) ||
+					musicPlayingCap.getCurrentMusicPlayer().isEmpty())
+			{
+				if(musicPlayingCap.getCassetteType() != EnumCassetteType.NONE)
+				{ //If the Cassette player isn't in hand and is playing a music, stop it
```
kirderf1, line 166:
```suggestion
				{ //If the Cassette player isn't in hand and is playing music, stop it
```

src/main/java/com/mraof/minestuck/item/weapon/MusicPlayerItem.java
```diff
@@ -0,0 +1,224 @@
...
+ * <p>
+ * The sprite of the item can change depending of if a cassette is currently inside or not.
+ * The tag used to do so is HasCassette.
+ * </p>
+ * {@link #inventoryTick(ItemStack, Level, Entity, int, boolean)}
+ * {@link #hasCassette(ItemStack)}
+ */
+
+@Mod.EventBusSubscriber(modid = Minestuck.MOD_ID)
+public class MusicPlayerItem extends WeaponItem
```
kirderf1, line 60: Rename to `MusicPlayerWeaponItem` or `MusicPlayerWeapon`

kirderf1 (commented):
Everything looks fine from my perspective, with just some minor comment/naming changes left.

I have heard of three ideas that have not yet been implemented:
1. Have an attack speed that matches the beat of the song played
2. A different weapon damage when the music player does not have a cassette
3. Looping the music track
4. Cassettes needing to be rewound to play from the beginning

Which of these are still of interest? Which do you want to implement as part of this PR before merging?
From my perspective, it seems relevant to implement 2 as part of this PR, and to get back to 4 in a separate PR some time in the future.

LunyaticCat:
> Which of these are still of interest? Which do you want to implement as part of this PR before merging?
From my perspective, it seems relevant to implement 2 as part of this PR, and to get back to 4 in a separate PR some time in the future.

2 as been done with the last commit, the other 3 are not as important and be addressed in another pull request indeed.

src/main/java/com/mraof/minestuck/item/weapon/MusicPlayerWeapon.java
```diff
@@ -0,0 +1,236 @@
...
+		}
+		return super.hurtEnemy(stack, target, attacker);
+	}
+	
+	public static boolean hasCassette(ItemStack stack)
+	{
+		if(!stack.hasTag() || !stack.getTag().contains("HasCasette"))
+			return !getItemHandler(stack).getStackInSlot(0).isEmpty();
+		else return stack.getTag().getBoolean("HasCassette");
+	}
```
kirderf1, line 235: Could be moved so that it is side-by-side with the corresponding function `getShareTag()`

kirderf1 (approved):
Just a minor comment, but otherwise I think that everything is good for merging!

src/main/java/com/mraof/minestuck/item/weapon/MusicPlayerWeapon.java
```diff
@@ -0,0 +1,236 @@
...
+	{
+		CompoundTag tag = stack.getTag() != null ? stack.getTag().copy() : new CompoundTag();
+		tag.putBoolean("HasCassette",
+				!getItemHandler(stack).getStackInSlot(0).isEmpty());
+		return tag;
+	}
+	
+	public static boolean hasCassette(ItemStack stack)
+	{
+		if(!stack.hasTag() || !stack.getTag().contains("HasCasette"))
```
kirderf1, line 232:
```suggestion
		if(!stack.hasTag() || !stack.getTag().contains("HasCassette"))
```

src/main/java/com/mraof/minestuck/item/weapon/MusicPlayerWeapon.java
```diff
@@ -0,0 +1,236 @@
...
+			return !getItemHandler(stack).getStackInSlot(0).isEmpty();
+		else return stack.getTag().getBoolean("HasCassette");
+	}
```
kirderf1, line 235:
```suggestion
	@Nullable
	@Override
	public CompoundTag getShareTag(ItemStack stack)
	{
		IItemHandler iitemHandler = getItemHandler(stack);
		CompoundTag nbt = stack.getTag() != null ? stack.getTag() : new CompoundTag();
		if(iitemHandler instanceof ItemStackHandler itemHandler)
			nbt.put("cassette", itemHandler.serializeNBT());
		return nbt;
	}
	
	@Override
	public void readShareTag(ItemStack stack, @Nullable CompoundTag nbt)
	{
		if(nbt == null)
			stack.setTag(null);
		else
		{
			IItemHandler iitemHandler = getItemHandler(stack);
			if(iitemHandler instanceof ItemStackHandler itemHandler)
				itemHandler.deserializeNBT(nbt.getCompound("cassette"));
			stack.setTag(nbt);
		}
	}
	
	public static boolean hasCassette(ItemStack stack)
	{
		return !getItemHandler(stack).getStackInSlot(0).isEmpty();
	}
```

Dweblenod:
Seems good testing it in game! I noticed what I think is an inconsequential bug. Which is that if you activate a cassette weapon in your offhand, and then wield another cassette weapon in your main hand without activating it, the player will attack with the effect of the offhand weapon.

Dweblenod (approved):
I think its good to go!

PR #367, opened 2022-09-06
Updating overworld modded ore generation

Dweblenod:
Created a deepslate variant of uranium ore(texture work by Sipher Nil) and changed around the heights at which cruxite and uranium ore generate.
Cruxite has been made much more abundant near the surface with a pyramidal distribution that peaks at y=64 but generates no farther down than y=0. Uranium also has a pyramidal distribution that expands below the bottom of the world, which means it is at peak abundance at y=-46 and goes no farther up than y=32


kirderf1:
You've modified height and amounts for both overworld ores and land ores, but lands have different terrain and depth from the overworld, and should probably not use identical placement anymore. Can you restore the old placement values for land ores or redefine placement values specifically for lands?

Dweblenod:
Restored old placement values, added a TODO for if we change land world height

kirderf1 (approved):

PR #369, opened 2022-09-11
Renaming tile entities to block entities

Dweblenod:
Using the refactoring tool in IntelliJ to rename tile entities to match 1.18's naming scheme


src/main/java/com/mraof/minestuck/computer/ComputerReference.java
```diff
@@ -9,7 +9,7 @@
 
 public interface ComputerReference
 {
-	static ComputerReference of(ComputerTileEntity te)
+	static ComputerReference of(ComputerBlockEntity te)
```
Dweblenod, line 12: on line 20 below, is it appropriate to change tile_entity to block_entity if I also change it in TEComputerReference? Also may TEComputerReference be renamed to BEComputerReference?

kirderf1: Yes and yes

Dweblenod: And to double confirm, I do not need to change that string(tile_entity) in any other file to the best of your knowledge?

kirderf1: Indeed, that nbt key should only be used in those two places.

Dweblenod: If thats the case then there are no other issues I can detect with this PR

kirderf1 (approved):
I don't think there's much reason to review this one thoroughly, so here's an approval.

PR #370, opened 2022-09-15
Making some push-able blocks also pull-able

Dweblenod:
Added an enum to determine whether Pushable Blocks could be pushed, pulled, or both. If its both than the player crouches to pull it. Currently the only Pushable Block has been made to be able to do both


kirderf1 (approved):
Your enum check to determine if the block is pushable/pullable could be simplified with two boolean fields for "pushable" and "pullable" inside the enum. However, it's not important to do here since the enum is particularly small and unlikely to grow larger.

PR #371, opened 2022-09-16
Fixing redstone power related block updating for a few puzzle blocks

Dweblenod:
Platform generators, trajectory blocks, and wireless redstone transmitters all had issues with their second blockstate property relating to redstone power overwriting the change in the first property. Other blocks utilizing this integer property that was being overwritten were investigated but these seem to be the only ones that needed any sort of fixing


kirderf1:
You're calling the old state `receiverState` in these changes, which isn't accurate. Can you rename those variables?

Dweblenod:
whoops! I saw that when I first started but thought I fixed it

kirderf1 (approved):

PR #373, opened 2022-09-18
Adding functionality to Mine and Grist

Dweblenod:
Mine and Grist was a pickaxe whose implementation was not complete, now when it mines something it is meant to break it drops the grist of that block instead of the item drops. 

Includes code by Kirderf, which turns the grist dropping into a global loot table. A datapack can edit the multiplier feature of the loot table, meaning they can cause the pickaxe to give a smaller or larger return of grist


kirderf1:
At a first glance, it seems more appropriate to make use of `item.mineBlock()` through `DestroyBlockEffect` rather than to listen to the break block event. Is there a specific reason you're using the event rather than an item effect?

Dweblenod:
> At a first glance, it seems more appropriate to make use of `item.mineBlock()` through `DestroyBlockEffect` rather than to listen to the break block event. Is there a specific reason you're using the event rather than an item effect?

That is originally what I had done, but at the time I did not know how to prevent the item that is normally dropped from dropping as well. Although perhaps the method I am using to remove the block would do that. Speaking of, I just realized that would probably not end well if it was used on anything with a container/menu such as chests

src/main/java/com/mraof/minestuck/event/ServerEventHandler.java
```diff
@@ -232,6 +240,32 @@ public static void onPlayerInjured(LivingHurtEvent event)
...
+	{
+		Player player = event.getPlayer();
+		BlockPos pos = event.getPos();
+		Level level = player.level;
+		ItemStack stack = player.getMainHandItem(); //pickaxes cannot be swung in the offhand
+		BlockState state = event.getState();
+		
+		if(stack.getItem() == MSItems.MINE_AND_GRIST.get() && !level.isClientSide && stack.isCorrectToolForDrops(state)) //will only drop grist instead of an item if it can be mined by the pick
+		{
+			ItemStack droppedStack = state.getCloneItemStack(null, level, pos, player);
```
kirderf1, line 254: I wonder if we could gristify the actual block loot drops instead of the pick-block result.

Dweblenod: Resolved by you!

kirderf1 (approved):

PR #374, opened 2022-09-23
Reimplement Grist layers

kirderf1:
Reimplements grist layers. At its base, grist types are chosen per region as usual. However with the old biome layers removed, a new method was designed for creating irregular, random borders between grist regions. The new implementation uses noise fields to randomly shift the position used to get the base grist type.

To help tweak the noise field settings and shift factor, a debugging render tool was created to visualize grist types in layers in an area.


Dweblenod (approved):
Fantastic!

PR #390, opened 2022-10-15
Items rebase

KingBalDoro:
re-based

PR #368, opened 2022-09-09
1.18.1 items' recipes grist costs involving alchemy

KingBalDoro:
IT BURNS

PR #396, opened 2022-10-25
Doro 1.18 item alchemy recipes

Dweblenod:
Redo of #368 and #390 that has been rebased. The majority of the work is still recognized as KingBalDoro's

Aside from new recipes for blocks/items from 1.17 and 1.18, this also importantly changes how copper is handled in alchemy as prior to this it had been the expensive forge tag variety


src/main/java/com/mraof/minestuck/data/recipe/MinestuckGristCostsProvider.java
```diff
@@ -312,6 +336,7 @@ protected void buildCraftingRecipes(Consumer<FinishedRecipe> recipeSaver)
 		ContainerGristCostBuilder.of(Items.WATER_BUCKET).grist(COBALT, 4).buildFor(recipeSaver, Minestuck.MOD_ID);
 		ContainerGristCostBuilder.of(Items.LAVA_BUCKET).grist(TAR, 16).buildFor(recipeSaver, Minestuck.MOD_ID);
 		ContainerGristCostBuilder.of(Items.MILK_BUCKET).grist(CHALK, 8).buildFor(recipeSaver, Minestuck.MOD_ID);
+		GristCostRecipeBuilder.of(Items.POWDER_SNOW_BUCKET).grist(RUST, 27).grist(CHALK, 4).buildFor(recipeSaver, Minestuck.MOD_ID); //TODO find way to handle SolidBucketItem instances like BucketItem
```
kirderf1, line 339: You could use a `SourceGristCostBuilder` and specify the bucket as a source directly. You could also use that grist cost type for grist cost relations between metal items, such as setting the ingot cost to be sourced from the raw material, for example.

PR #397, opened 2022-10-29
Small tweaks to Rock lands and fixed flooding of all caves in Lands

Dweblenod:
Standard rock lands now have glow lichen and all rock lands have amethyst geodes as well as all three default carvers. Eventually trying to get a handle on Noise generated caves(as opposed to the more traditionally created Carver based caves) to see if they can generate in lands specifically if those lands already have caves

Currently aquifers/wet caves only seem to generate under ocean biomes


src/main/java/com/mraof/minestuck/world/gen/LandGenSettings.java
```diff
@@ -92,7 +92,7 @@ Holder<NoiseGeneratorSettings> createDimensionSettings(Registry<DensityFunction>
 		
 		NoiseGeneratorSettings settings = new NoiseGeneratorSettings(noiseSettings, blockRegistry.getBlockState("ground"), blockRegistry.getBlockState("ocean"),
 				MSDensityFunctions.makeLandNoiseRouter(densityFunctions),
-				surfaceRule, 64, false, false, false, false);
+				surfaceRule, 64, false, true, false, false);
```
kirderf1, line 95: I'm guessing this is the boolean that enables the aquifier system. Do you have any details on what this change does in practice which solves cave flooding?

Dweblenod: I don't have the full picture to understand all the reasons of why it works, but the result is that instead of the entirety of a cave being filled with fluid(lava at y=10 and below, with everything above that being the ocean fluid) there are select caves that are saturated with liquid and than others that have none. When it was set to false, it didnt matter if a cave had contact with the surface or not, it would always be wet

Dweblenod: I forgot to include, it seems the wet caves occur under ocean biomes with this change(so its not a full fix)

kirderf1 (approved):

PR #398, opened 2022-11-04
Held item deletion bug patch

Dweblenod:
When an item in the players inventory menu was being carried(held by mouse) and the button to enter the sylladex from the Minestuck menu was clicked, the item would not be transitioned to the resulting container menu and would instead be deleted. Players are no longer able to open their sylladex when holding an item


kirderf1:
The problem does not have to do with the sylladex button specifically, it happens in this container menu is closed, regardless of if it was closed with the button or not. Normally, the held item gets moved back to the inventory or dropped. This happens in `abstractContainerMenu.removed()`. However, `CaptchaDeckMenu` overrides this method without calling `super.removed()`, which is the real problem. I'll close this PR and fix the issue directly.

PR #404, opened 2022-11-27
Geckolib Horse clock

Dweblenod:
Implements a clock. It is shaped like a horse. And it is beautiful...

Geckolib animated multiblock that lets off redstone signals as it counts time. 
- Textures, model, and animation by Riotmode
- Chimes at the start of a new day or when the base of the multiblock is right clicked
- The dials on its chest display the approximate time
- Sound effects by Dweblenod(edits of royalty free sources, listed in metadata)
- It will not give off any sound if over top a vibration occluding block, the redstone clock block was changed to match(had been  using wool tag since occlusion tag was not yet created)
- reorganized some packages relating to animated block entities and entities
- runData seems to have resulted in some unexpected changes that were not merged with the switch to 1.18 for geckolibs_remodels. Also deleted unused models that snuck in

Redstone signal strengths:
-tick: 4
-tock: 2
-chime: 15


src/main/java/com/mraof/minestuck/item/block/HorseClockItem.java
```diff
@@ -0,0 +1,33 @@
...
+			Player player = context.getPlayer();
+			ItemStack itemstack = context.getItemInHand();
+			
+			if (player == null || !player.getAbilities().instabuild) {
+				itemstack.shrink(1);
+			}
+		}
+		
+		return superBool;
+	}
```
kirderf1, line 32: I can't tell why this is done. Should it not already be the case that one item is consumed if it is placed in survival?

Dweblenod: It extends MultiblockItem, which it itself does not have a super call. Meaning stack shrinking needs to be added back in

kirderf1: Ah, that is a bug in the implementation of the multiblock item then. I'll push a fix on the main branch.

Dweblenod: Updated and now have deleted this class

src/main/java/com/mraof/minestuck/block/HorseClockBlock.java
```diff
@@ -0,0 +1,141 @@
...
+	
+	@Nullable
+	@Override
+	public BlockEntity newBlockEntity(BlockPos pos, BlockState state)
+	{
+		if(state.is(MSBlocks.HORSE_CLOCK.BOTTOM.get())) //prevents non block entity blocks from trying to tick
+		{
+			return new HorseClockBlockEntity(pos, state);
+		} else
+			return null;
```
kirderf1, line 62: I think I recall some sort of minor problem when returning a null block entity. Perhaps it is relevant to avoid this with a new class for the block entity/ticker block, akin to other multiblocks?

src/main/java/com/mraof/minestuck/block/HorseClockBlock.java
```diff
@@ -0,0 +1,141 @@
...
+import net.minecraft.world.level.block.state.properties.BlockStateProperties;
+import net.minecraft.world.level.block.state.properties.IntegerProperty;
+import net.minecraft.world.phys.BlockHitResult;
+
+import javax.annotation.Nullable;
+import java.util.Random;
+
+public class HorseClockBlock extends MultiMachineBlock implements EntityBlock
+{
+	public static final IntegerProperty POWER = BlockStateProperties.POWER;
```
kirderf1, line 32: Are all three blocks making use of the power property? Also, when writing a javadoc, you could add a description on how the power property is used.

kirderf1: For the javadoc on this class, you could just refer to the block entity class if you don't want to add anything to the block class specifically.

src/main/java/com/mraof/minestuck/block/HorseClockBlock.java
```diff
@@ -0,0 +1,141 @@
...
+	
+	@Override
+	public void neighborChanged(BlockState state, Level level, BlockPos pos, Block block, BlockPos fromPos, boolean isMoving)
+	{
+		updateSurvival(state, level, pos);
+	}
+	
+	public static void updateSurvival(BlockState state, Level level, BlockPos pos)
+	{
+		if(!level.isClientSide() && state.getBlock() instanceof HorseClockBlock)
```
kirderf1, line 78: Is there a point to this instance check?

src/main/java/com/mraof/minestuck/block/HorseClockBlock.java
```diff
@@ -0,0 +1,141 @@
...
+		{
+			if(state.is(MSBlocks.HORSE_CLOCK.BOTTOM.get()) &&
+					!level.getBlockState(pos.above()).is(MSBlocks.HORSE_CLOCK.CENTER.get()))
+				level.destroyBlock(pos, true);
+			else if(state.is(MSBlocks.HORSE_CLOCK.CENTER.get()) &&
+					(!level.getBlockState(pos.above()).is(MSBlocks.HORSE_CLOCK.TOP.get()) || !level.getBlockState(pos.below()).is(MSBlocks.HORSE_CLOCK.BOTTOM.get())))
+				level.destroyBlock(pos, true);
+			else if(state.is(MSBlocks.HORSE_CLOCK.TOP.get()) &&
+					!level.getBlockState(pos.below()).is(MSBlocks.HORSE_CLOCK.CENTER.get()))
+				level.destroyBlock(pos, true);
```
kirderf1, line 88: It would've been nice to use a more general function from the multiblock to handle these validity checks, though I'm not sure if there's something that's readily available for this.

Dweblenod: I thought we might have had it, and looked briefly, but didnt find anything. I would have added it but did not feel inspired

src/main/java/com/mraof/minestuck/block/HorseClockBlock.java
Dweblenod: If there had been, it was no longer relevant and has been removed

src/main/java/com/mraof/minestuck/block/HorseClockBlock.java
Dweblenod: Created a class for Bottom within HorseClockBlock

src/main/java/com/mraof/minestuck/block/HorseClockBlock.java
```diff
@@ -0,0 +1,141 @@
...
+			return !level.isClientSide ? BlockUtil.checkTypeForTicker(placedType, MSBlockEntityTypes.HORSE_CLOCK.get(), HorseClockBlockEntity::serverTick) : null;
+		}
+	}
+	
+	@Override
+	protected void createBlockStateDefinition(StateDefinition.Builder<Block, BlockState> builder)
+	{
+		super.createBlockStateDefinition(builder);
+		builder.add(POWER);
+	}
```
kirderf1, line 140: Move this up so that the inner class is at the bottom.

kirderf1 (approved):
Should be good to go. Just an overriden function that got separated from the rest of the block's functions by the new inner block class.

PR #412, opened 2023-01-31
Translation keys update

LunyaticCat:


kirderf1:
Because of the large amounts of other changes from data generation, I'd like to put those changes on 1.19 separately first. I should be able to get to it tonight, though you might need to merge them afterwards, depending on how it affects this PR.

kirderf1 (approved):

PR #419, opened 2023-02-24
update various gui item and block textures

akisephila:
:33 < updates captcha card and cruxite dowel gui silhouettes, wooden cactus block, golden seeds crop, blooming cactus blocks to use the new cactus palette, and treated and dead planks, gold/iron boat items, and cactaceae cutlass sword since it missed the big texture update


Dweblenod:
The commit history is clean, the description is accurate, and the textures changes look good! If you make any new commits, try to focus on 1.19

PR #420, opened 2023-02-26
update a few canes, a couple blocks that used programmer art as a base, and the boat entities

akisephila:


akisephila:
:33 < h33h33 420

Dweblenod:
The changes seem appropriate, however there is an issue in that this contains changes from your previous PR. Reset your 1.18 branch to the mraof 1.18 branch and then commit the new changes from there

PR #421, opened 2023-02-27
:33 < fixed version of my previous pr with updated branch

akisephila:
:33 < touches up the cactaceae cutlass again, updates the remaining canes that missed the texture update, and updates boat entities


akisephila:
:(( < no more 420

PR #433, opened 2023-03-21
Cleanup of small machines

kirderf1:
Does some cleanup of the machine block entities and screens, focusing on taking steps from inheritance towards composition, while also fixing issues along the way.
Primarily, from `MachineBlockEntity`, I created `ProgressTracker` which keeps track of progress and the different modes that a machine can be in.
Before the PR is ready, I'm gonna do more with the screen-side of things, and probably also redesigning how the GO button behaves on clicks to something a little bit more intuitive.


Dweblenod (changes requested):
I cant seem to get the mini totem lathe and mini punch designix to work. As soon as I press start it immediately stops. 

Likewise, perhaps unrelated, when using the full sized totem lathe, it stops being usable after one use as the dowel block entity refuses to give up its dowel. The card holding block entity also refuses to be destroyed permanently when broken if it has an item stored in it. It will simple reappear as if nothing happened.

kirderf1:
Fixed the mentioned mini machine issues. The others sounds like a problem from #431, so I'll fix it in a commit directly on the 1.19 branch.

PR #437, opened 2023-03-30
Skaianet Denier block

Dweblenod:
A block which prevents anyone from Entering if the block would be taken as part of the Entry island. Stops the Entry process in a manner similar to trying to take someone else's computer or command blocks (while not in creative).

Crafted with iron ingots, an energy core, computer parts, and a lightning rod

Textures made by Akisephila


Dweblenod:
Should be simple to review, just wanted to make sure theres no additional considerations I am forgetting so I didnt want to merge it before it gets checked

kirderf1 (approved):

PR #443, opened 2023-04-05
Geckolib model and animation changes

LunyaticCat:
Changed consorts and underling models


Dweblenod:
Please add an idle animation controller for imps and liches. You can base it off of the consort idle animation controller. Ogres do not yet have an idle animation, and Basilisks have the tail flick + tongue flick, but it will take more work to get those added.

Please also add the giclops model and animation posted!

LunyaticCat:
things to fix :
- imp wrong texture
- imp no idle animation
- lich raising their arms while walking
- ogre no animation
- basilisk are small and have a detail problem

Dweblenod:
The issue with the imp and basilisk textures are that the default layer and the details layer were not separated out properly. The first layer should be all white. The imps detail layer does not contain the eyes. The reason any eyes show up at all is because of the messed up default layer

PR #444, opened 2023-04-07
Land gen tweaks

seriousCoolness:
Tweaked a few values in the LandGenSettings and MSNoiseParameters to make the terrain gen less flat, and more interesting to navigate.


Dweblenod:
Make sure to add to the CHANGELOG file!

kirderf1 (approved):
Looks all good and I'd be fine merging this as it is.

If you want to do some more experimenting, it might be interesting to have a higher angle for rough terrain than for normal terrain.
Is that something you want to look into before I merge?

seriousCoolness:
Nah, I am satisfied with the state of it for now. It’s a good idea for after 4/13, though. Go ahead and merge it.

PR #445, opened 2023-04-08
Structure nbt defined shadewood trees and Shade rough biome tweaks

Dweblenod:
Adding 4 manually built "ornate" shadewood trees made by Carnie and tweaked oil pool/shade stone blob frequency


src/main/java/com/mraof/minestuck/world/gen/feature/MSPlacedFeatures.java
kirderf1, line 1: Realized that the oil pool and shade stone block blob (and maybe others?) additions doesn't seem to already be mentioned in the changelog. Perhaps something that should be mentioned there?

kirderf1 (approved):
Other than the thing that may be missing from the changelog, this looks fine. Feel free to merge after the potential changelog addition.

src/main/java/com/mraof/minestuck/world/gen/feature/MSPlacedFeatures.java
Dweblenod: Since the oil pool and shade block blobs were added in this Update, it didnt seem like there needed to be a changelog for tweaks to something yet to be in the mod

Dweblenod (commented):
Found unexpected issues with the structure defined trees when placed as saplings

src/main/java/com/mraof/minestuck/world/gen/feature/MSPlacedFeatures.java
kirderf1: Yeah, the tweaking of the features shouldn't be in the changelog for that reason. But the addition of the features should be in the changelog. (Which was previously missed)

PR #449, opened 2023-04-09
Entry cleanup

kirderf1:
Code cleanup for entry-related code. Ideally fixing performance-related problems as well.

Not yet sure if it will be ready for 4/13.


Dweblenod:
I have no issues with implementing what youve made so far!

src/main/java/com/mraof/minestuck/entry/PostEntryTask.java
```diff
@@ -26,7 +23,7 @@
 	 * The maximum amount of time (in milliseconds) to spend updating blocks,
 	 * before leaving the rest for the next game tick.
 	 */
-	private static final long MIN_TIME = 20;
+	private static final long MIN_TIME = 15;
```
Dweblenod, line 26: Is it possible to have something similar for the pre entry section?

Dweblenod (approved):
I didnt test it in a dedicated server but it looks good when testing single player! In the future it seems like it would make sense to spread the pre entry process over several ticks(would need some logic to help prevent block duplication I imagine)

PR #454, opened 2023-04-13
Updated Rainbow Tree Textures

akisephila:
Updated Rainbow Leaves, Logs, Planks, and Saplings, textures by Carnie. 
:33 < couldnt be bothered to make a new branch for this so here
:33 < accidentally made a pr to the wrong branch


kirderf1 (commented):
Too bad that the sapling lost its rgb visual properties. However with the properties removed visually, I don't think it makes much sense to keep the properties in the block. As such, I believe that these should be removed unless the sapling is changed to use these properties again.

With the release of 1.10.0.0, the changelog entry should be moved to the top since this PR didn't get merged into that update.

A slight worry I have with the current textures and animations is that it might become a bit more homogeneous than otherwise intended. Not necessarily something that needs changing though.

CHANGELOG.md
```diff
@@ -11,6 +11,14 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 - Added 75 more blocks to the default atheneum list.
 - Transportalizer particle
 
+### Changed
+
+- Updated textures for Rainbow Logs, Planks, Leaves and Saplings
+
+### Removed
+
+- Removed unused blockstates and models for the Rainbow Sapling
```
kirderf1, line 20: Calling them unused is a bit misleading since they weren't unused before this update. Perhaps phrase it as a change to a simpler growth process without colored state properties?

CHANGELOG.md
```diff
@@ -25,7 +33,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 
 ### Contributors for this release
 
-- kirderf1, Caldw3ll, Dweblenod, Riotmode
+- kirderf1, Caldw3ll, Dweblenod, Riotmode, Carnie, Akisephila
```
kirderf1, line 36: That update has been released already, so I recommend you merge 1.19 and then move the changelog and contributors to the new unreleased block at the top

kirderf1 (approved):

src/main/java/com/mraof/minestuck/block/MSBlocks.java
```diff
@@ -290,7 +291,7 @@ private static Block coalOre(BlockBehaviour.Properties properties)
 	public static final RegistryObject<Block> SHADEWOOD_LEAVES = REGISTER.register("shadewood_leaves", () -> new FlammableLeavesBlock(Block.Properties.of(Material.LEAVES).strength(0.2F).randomTicks().sound(SoundType.GRASS).noOcclusion().isValidSpawn(MSBlocks::leafSpawns).isSuffocating(MSBlocks::never).isViewBlocking(MSBlocks::never)));
 	public static final RegistryObject<Block> SHROOMY_SHADEWOOD_LEAVES = REGISTER.register("shroomy_shadewood_leaves", () -> new FlammableLeavesBlock(Block.Properties.of(Material.LEAVES).strength(0.2F).randomTicks().lightLevel(state -> 11).sound(SoundType.GRASS).noOcclusion().isValidSpawn(MSBlocks::leafSpawns).isSuffocating(MSBlocks::never).isViewBlocking(MSBlocks::never)));
 	
-	public static final RegistryObject<BushBlock> RAINBOW_SAPLING = REGISTER.register("rainbow_sapling", () -> new RainbowSaplingBlock(Block.Properties.of(Material.PLANT).noCollission().randomTicks().strength(0).sound(SoundType.GRASS)));
+	public static final RegistryObject<BushBlock> RAINBOW_SAPLING = REGISTER.register("rainbow_sapling", () -> new SaplingBlock(new RainbowTree(), Block.Properties.of(Material.PLANT).noCollission().randomTicks().strength(0).sound(SoundType.GRASS)));
```
Dweblenod, line 294: The only issue I have with this is that the SaplingBlock cannot survive on wool or terracotta. The Feature that places the rainbow trees checks to see whether a rainbow sapling would be capable of being placed on the ground and since theres never a dirt material there for it to pick from, there ends up being no trees in the land. You may need to reimplement RainbowSaplingBlock and make it match closer to AspectSaplingBlock or EndSaplingBlock

kirderf1: Ah right, I forgot about that when I originally suggested replacing the block class with `SaplingBlock`. But yeah, the solution is to reintroduce `RainbowSaplingBlock`, but extending `SaplingBlock` and only overriding whatever it needs to change the supported ground block.

kirderf1:
Was just reminded of this PR. Shall we set someone on fixing the sapling problem Dweblenod pointed out so that we can finally get this merged?

kirderf1:
With no update or activity for a while and a new release coming up, I'll take care of that myself then.

PR #457, opened 2023-04-14
Client cache cleanup

kirderf1:
Better handling of access to grist caches client-side, as well as fixing client-side problems relating to the cache limit.


Dweblenod:
Do you need help testing it in-game?

kirderf1:
Yeah, that would be appreciated.

Dweblenod:
Grist picked up off the ground that are entering the gutter do not provide a toast. Is this intended behavior?

Dweblenod:
When testing in a fresh dedicated server setting, I could not trigger any bugs while connected to myself. However as was discussed on Discord, an old world had an issue where the build grist was higher than the cache should have allowed as the players grist set was established before the cache was in place

kirderf1:
> However as was discussed on Discord, an old world had an issue where the build grist was higher than the cache should have allowed as the players grist set was established before the cache was in place

My approach implemented in KingBalDoro#9 when a cache is already over capacity is to let it sit there and have the cache behave as if it currently was at max capacity for that grist type, neither discarding or sending away any of that grist. I suppose the grist cache screen should follow that behavior as well, drawing the bar as if it was full even when it's overfull.

> Grist picked up off the ground that are entering the gutter do not provide a toast. Is this intended behavior?

Yeah, problems with what toasts should show when the gutter gets involved were noticed during my review of the original gutter PR, but without enough time to adress it.
While it should change from how it behaves right now, I lack a vision for how it should actually behave.

Dweblenod:
> Yeah, problems with what toasts should show when the gutter gets involved were noticed during my review of the original gutter PR, but without enough time to adress it. While it should change from how it behaves right now, I lack a vision for how it should actually behave.

Making a Post in Discord with a suggestion to make a grist gutter Toast icon to show that its going to the gutter

src/main/java/com/mraof/minestuck/client/gui/MinestuckScreen.java
```diff
@@ -47,10 +44,10 @@ public void drawGrist(PoseStack poseStack, int xOffset, int yOffset, int xcor, i
 		//Show the name of the grist instead of the count if displaying a tooltip
 		boolean showName = false;
 		GristType tooltipType = null;
-		GristSet clientGrist = ClientPlayerData.getClientGrist();
-		int clientRung = ClientPlayerData.getRung();
-		long cacheLimit = Echeladder.getGristCapacity(clientRung);
-
+		ClientPlayerData.ClientCache cache = ClientPlayerData.getGristCache(ClientEditHandler.isActive() ? ClientPlayerData.CacheSource.EDITMODE : ClientPlayerData.CacheSource.PLAYER);
+		GristSet clientGrist = cache.set();
+		long cacheLimit = cache.limit();
```
Dweblenod, line 49: Perhaps we should put the GristTorrent icon next to the number if its in the process of downloading grist from the gutter

Dweblenod (approved):
Made a PR to fix the grist gutter toast situation, and earlier made a review comment suggesting we move the location of the grist torrent icon, but other than that I think its good to go!

src/main/java/com/mraof/minestuck/client/gui/MinestuckScreen.java
kirderf1: For this, I think we would need to sync which types of grist exist in the gutter, and have the torret icon display when the type in question exists in the gutter, as well as the cache is under capacity. This would then display for all grist types that *could* be sent from the gutter. We could of course instead consider displaying something when it actually gets sent.
Regardless, I think that it's outside the scope of this PR and could be looked at by someone after this.

PR #458, opened 2023-04-14
Editmode and atheneum fixes

seriousCoolness:
Added additional blocks to the atheneum and changed the order to be less cluttered looking.
Fixed bug where players in editmode could not break multiblock machines.


kirderf1 (approved):
Looks fine, though I only skimmed through the atheneum changes. (Which I doubt need a thorough review)

src/main/java/com/mraof/minestuck/computer/editmode/DeployList.java
```diff
@@ -77,71 +77,145 @@ public static void registerItems()
 		registerItem("card_punched_card", new GristSet(GristTypes.BUILD, 25), null, 0, config(MinestuckConfig.SERVER.deployCard), (sburbConnection, world) -> AlchemyHelper.createCard(new ItemStack(MSItems.CAPTCHA_CARD.get()), true), EntryLists.DEPLOY);
 		
 		//Atheneum
-		registerItem("stone", new ItemStack(Blocks.STONE), new GristSet(GristTypes.BUILD, 1), 0, EntryLists.ATHENEUM);
```
Dweblenod, line 80: Do you think it would make sense to add dirt/grass/stone to the atheneum but on a later tier? That way they cant be exploited at first but are still available. Or do you think its best to focus on blocks that must be crafted?

seriousCoolness: I think that dirt, stone, and grass would still be too overpowered to use, even at 4th or 5th tier.

Dweblenod (approved):

PR #459, opened 2023-04-14
Minestuck Particles registry and transportalizer particle

Dweblenod:
Deferred registry for new particles, first of which is the transportalizer particle made by Riotmode

Transportalizer particle plays at both the entry and exit 'portalizers


src/main/java/com/mraof/minestuck/client/particle/MSParticleType.java
```diff
@@ -0,0 +1,32 @@
...
+import net.minecraftforge.registries.RegistryObject;
+
+@Mod.EventBusSubscriber(modid = Minestuck.MOD_ID, bus = Mod.EventBusSubscriber.Bus.MOD, value = Dist.CLIENT)
+public class MSParticleType
+{
+	public static final DeferredRegister<ParticleType<?>> REGISTER = DeferredRegister.create(ForgeRegistries.PARTICLE_TYPES, Minestuck.MOD_ID);
+	
+	public static final RegistryObject<SimpleParticleType> TRANSPORTALIZER = REGISTER.register("transportalizer", () -> new SimpleParticleType(false));
+	
+	@OnlyIn(Dist.CLIENT)
```
kirderf1, line 24: This should generally be avoided in mod code. I recommend separating client-only code (particle provider registration) from common code (the register and transportalizer particle type), so that the common code isn't stored in the client package.

src/main/java/com/mraof/minestuck/client/particle/MSParticleType.java
```diff
@@ -0,0 +1,32 @@
...
+	@SubscribeEvent
+	public static void registerFactories(RegisterParticleProvidersEvent event)
+	{
+		ParticleEngine particles = Minecraft.getInstance().particleEngine;
+		
+		particles.register(MSParticleType.TRANSPORTALIZER.get(), TransportalizerParticle.Provider::new);
```
kirderf1, line 30: There's an `event.register()` function that I think you're meant to be using.

src/main/java/com/mraof/minestuck/client/particle/TransportalizerParticle.java
```diff
@@ -0,0 +1,54 @@
...
+	}
+	
+	@Override
+	public void tick()
+	{
+		super.tick();
+		setSpriteFromAge(this.spriteSet);
+	}
+	
+	@OnlyIn(Dist.CLIENT)
```
kirderf1, line 36: Not meant to be used in mod code

kirderf1 (changes requested):

src/main/java/com/mraof/minestuck/client/particle/MSParticleType.java
Dweblenod: Should I move MSParticleTypes to a different package? If so what should be left in the client package?

kirderf1: Yeah, that'd be good. Since the particle provider should only be registered client-side, that's what should remain in the client package. If you want, you could try putting that function in `ClientProxy` until it grows large enough to warrant its own class.

src/main/java/com/mraof/minestuck/client/particle/TransportalizerParticle.java
Dweblenod: Deleting it

src/main/java/com/mraof/minestuck/client/particle/MSParticleType.java
Dweblenod: Changed to use event.register()

src/main/java/com/mraof/minestuck/client/particle/MSParticleType.java
Dweblenod: Ended up creating a new package, let me know if you find how its organized now suitable

src/main/java/com/mraof/minestuck/blockentity/TransportalizerBlockEntity.java
```diff
@@ -167,9 +177,19 @@ protected void teleport(Entity entity)
 				return;
 			}
 			
-			entity = Teleport.teleportEntity(entity, (ServerLevel) destTransportalizer.level, destination.pos().getX() + 0.5, destination.pos().getY() + 0.6, destination.pos().getZ() + 0.5, entity.getYRot(), entity.getXRot());
+			ServerLevel originServerLevel = (ServerLevel) this.level;
+			ServerLevel destinationServerLevel = (ServerLevel) destTransportalizer.level;
```
kirderf1, line 181: Just `originLevel` and `destinationLevel` should be enough.

src/main/java/com/mraof/minestuck/client/particle/MSParticleType.java
kirderf1: It's kind of the opposite problem with `TransportalizerParticle` now. Since I doubt that the common side of particles won't grow beyond `MSParticleTypes`, it might be better to put that specific class in util instead, and make a particle package in client for `TransportalizerParticle` (and future particle classes of that kind)

src/main/java/com/mraof/minestuck/blockentity/TransportalizerBlockEntity.java
Dweblenod: Reorganized

src/main/java/com/mraof/minestuck/client/particle/MSParticleType.java
Dweblenod: Let me know if I got it how you want it!

kirderf1 (approved):

PR #460, opened 2023-04-15
Entry delay and effect

kirderf1:
Puts the entry at a delay from when the land dimension is generated to when the actual entry happens. This gives land worldgen a couple of ticks to do its thing, ideally improving the performance of entry. Since how much time worldgen needs depends on the performance of the machine running it, this delay is configurable in the common config.
To ensure that the game tries to load the chunks (and keeps them loaded), a chunk ticket gets added that is removed once entry is done.

An effect has been added that spawns light-blue particles in the entry area during this time.

Playtesting of this would be appreciated, at the very least to see if the current default value is suitable or should be higher, as well as if the entry effect should be adjusted in some way.

I've also considered a command that triggers land generation + chunk ticket earlier than that (for GMs to use), but I might wait on that for a later time unless someone wants me to add it with this PR.


Dweblenod:
I will admit it sound counter intuitive to what this PR accomplishes, but with the improvements to performance and with the lower world generation limit do you think it would make sense to increase the depth/total blocks that players Enter with?

kirderf1:
I did consider it, though at the moment I think we should hold off and see how things works out with this delay first.

Dweblenod:
> I did consider it, though at the moment I think we should hold off and see how things works out with this delay first.

Makes sense!

src/main/java/com/mraof/minestuck/MinestuckConfig.java
```diff
@@ -32,6 +33,11 @@ private Common(ForgeConfigSpec.Builder builder)
 			logItemsWithRecipeAndCost = builder.comment("Makes the recipe-generated grist cost process log any items that has a grist cost, but which could also be provided as a recipe generated cost. Useful for finding items that probably do not need manual grist costs.")
 					.define("logItemsWithRecipeAndCost", false);
 			builder.pop();
+			
+			builder.push("performance");
+			entryDelay = builder.comment("Time reserved for worldgen between generating the land and the actual entry. Measured in ticks. Try increasing this if entry halts the game too much.")
+					.defineInRange("entryDelay", 40, 0, Integer.MAX_VALUE);
```
Dweblenod, line 39: Recommendation to double the default value

Dweblenod (commented):
First round of testing went well! Tried double entry, getting a bunch of moving blocks to see if there would be any duplications, etc. Next will try it with a dedicated server

src/main/java/com/mraof/minestuck/MinestuckConfig.java
Dweblenod: Actually perhaps even more. That way performance is optimal and players get more of a chance to see the particle effects

Dweblenod:
I may be misunderstanding it's use, but perhaps the MIN_TIME field in PostEntryTask should also be increased from 15 to 20 to improve the post entry strain on the server

kirderf1:
Ah, I'll add a piece of javadoc for that

kirderf1:
Or no, see over the javadoc that's already there

Dweblenod:
> Or no, see over the javadoc that's already there
Ah! Then perhaps it should be lowered from 15 to 10 to reduce strain

src/main/java/com/mraof/minestuck/entry/EntryProcess.java
```diff
@@ -40,104 +49,171 @@
...
+		ServerLevel landWorld = Objects.requireNonNull(player.getServer()).getLevel(land);
+		if(landWorld == null)
+		{
+			player.sendSystemMessage(Component.literal("Unable to find land dimension. Something is not as it should be!"));
+			return;
+		}
+		
+		//Teleports the player to their home in the Medium, without any bells or whistles.
+		BlockPos pos = new BlockPos(0, 100, 0);
+		Teleport.teleportEntity(player, landWorld, pos.getX() + 0.5F, pos.getY(), pos.getZ() + 0.5F, player.getYRot(), player.getXRot());
```
Dweblenod, line 158: In testing, its likely the player will reenter underground

Dweblenod (approved):
Besides from the second entry placement, didnt find any bugs so far in a dedicated server or in single player!

I created a PR for a recommendation I have been thinking of if you would be willing to look at it before you merge this but other than that and the consideration to implement custom particle effects as discussed on Discord it seems good!

Dweblenod (changes requested):
Now that players can move around in the period before entry actually occurs, they have the ability to move elsewhere. If they move out of the entry region they may fall to their death or suffocate when they are teleported with the land as they will teleport to a spot off the entry platform

src/main/java/com/mraof/minestuck/entry/EntryProcess.java
kirderf1: Has now been changed to use the heightmap for motion blocking blocks.

Dweblenod:
Doing another in-game test now

Dweblenod (approved):

Dweblenod:
Seems good to go!

PR #463, opened 2023-04-15
Creating Title-Subtitle style Land Entry message

Dweblenod:
Does not replace existing chat message, just another version! Introduces Carima font used by Homestuck for their Land Entry messages


src/main/resources/assets/minestuck/font/carima.ttf
kirderf1, line 1: What is this font licensed under? Is there somewhere we can note which license that applies?

src/main/java/com/mraof/minestuck/data/MinestuckEnUsLanguageProvider.java
```diff
@@ -1699,7 +1699,8 @@ protected void addTranslations()
 		addLand(MonstersLandType.MONSTERS, "Monsters");
 		addLand(TowersLandType.TOWERS, "Towers");
 		add(LandTypePair.FORMAT, "Land of %s and %s");
-		add(SburbHandler.LAND_ENTRY, "Welcome to %s.");
+		add(SburbHandler.CHAT_LAND_ENTRY, "Welcome to %s.");
+		add(SburbHandler.TITLE_LAND_ENTRY, "%s");
```
kirderf1, line 1703: What is the point to this specific lang entry?

Dweblenod: I was under the impression we wanted to keep it translatable, in which case two arguments are needed unless I can just "%s" into SburbHandler

src/main/resources/assets/minestuck/font/carima.ttf
Dweblenod: It has been determined that we would be distributing this, which goes against copyright. Removing

src/main/java/com/mraof/minestuck/data/MinestuckEnUsLanguageProvider.java
kirderf1: Keeping it translatable? Two arguments are needed? I'm not sure what problem you're trying to solve.

Dweblenod: Issue resolved, thanks for help!

kirderf1 (approved):

PR #464, opened 2023-04-16
Cone-based hit detection for underlings

seriousCoolness:
Animated melee attacks now take an attack direction and a width angle, and will use those to check to make sure the target is within the attack's cone, so that underlings cannot attack things that are behind them, and the actual area where they can attack matches closer to their animations.


src/main/java/com/mraof/minestuck/entity/ai/attack/AnimatedAttackWhenInRangeGoal.java
```diff
@@ -110,6 +117,18 @@ protected boolean withinRange(@Nonnull LivingEntity target)
...
+	{
+		Vec3 attackDirection = Vec3.directionFromRotation(0, this.entity.getVisualRotationYInDegrees() + this.attackDirectionOffset);
+		Vec3 targetDirection = this.entity.position().vectorTo(target.position()).normalize();
+		targetDirection = new Vec3(targetDirection.x, 0, targetDirection.z);
+		
+		float angleOfTargetFromConeCenter = (180.0F / 2.0F) * ((((float)attackDirection.dot(targetDirection)) * -1.0F) + 1.0F);
+		
+		return angleOfTargetFromConeCenter <= this.attackConeAngle;
+		
+	}
```
Dweblenod, line 130: The method name is pretty self descriptive(you did a good job!) but mind adding a comment to explain a little bit of the math to help anyone who might try adding new attacks?

src/main/java/com/mraof/minestuck/entity/underling/BasiliskEntity.java
```diff
@@ -72,8 +72,8 @@ public static AttributeSupplier.Builder basiliskAttributes()
 	protected void registerGoals()
 	{
 		super.registerGoals();
-		this.goalSelector.addGoal(2, new AnimatedAttackWhenInRangeGoal<>(this, TAIL_SWING_ANIMATION, 0, 3.5F, 40));
-		this.goalSelector.addGoal(2, new AnimatedAttackWhenInRangeGoal<>(this, BITE_ANIMATION, 3.5F, AnimatedAttackWhenInRangeGoal.STANDARD_MELEE_RANGE, 40));
+		this.goalSelector.addGoal(2, new AnimatedAttackWhenInRangeGoal<>(this, TAIL_SWING_ANIMATION, 0, 3.5F, 40, 180.0F, AnimatedAttackWhenInRangeGoal.STANDARD_MELEE_ARC));
+		this.goalSelector.addGoal(2, new AnimatedAttackWhenInRangeGoal<>(this, BITE_ANIMATION, 3.5F, AnimatedAttackWhenInRangeGoal.STANDARD_MELEE_RANGE, 40, 0, 45.0F));
```
Dweblenod, line 76: Ok I actually really like that you can stand on their back now without problems. However there is a (in my opinion) too generous of a region on the front half of their body when standing close to their head where they cannot hurt you at all. See about either slightly expanding the cone or range of either the bite or whip attacks to compensate

Dweblenod (commented):
Overall looks good! Added two review comments on places where I think it could use improvement, and I have one additional thought to share. The standard range and the calculations seem pretty solid, but I do find it to be an issue that the attacker will sometimes try to attack multiple times before they change their look direction due to the fact that often their look direction is locked in place by the phased animation. It results in players being able to strafe behind an enemies back and wait their while getting missed a couple times, with the animation effectively locking that enemy in place(making them an easier target without as much skill). Let me know what you think of the problem and if you do think it is one we can brainstorm a solution

src/main/java/com/mraof/minestuck/entity/underling/BasiliskEntity.java
Dweblenod: Sorry to say but I feel the issue is still there. If you walk straight up to the front of the basilisk, you can get closer than where the bite attack activates. You can continue standing inside its head/neck and while the whip animation will play, it does not injure

seriousCoolness: Why would we want the whip animation to injure you there? That doesn't seem like it would be intuitive, since it doesn't match the animation.

Dweblenod: What I mean is that overall there needs to be adjustments in order to reduce the size of the safe spot. If you dont think increasing the tail whip range/cone makes sense, you can adjust the bite range

seriousCoolness: Oh, I see the problem now. I thought you meant that you were safe when standing behind the head to the right or left, but you can also LITERALLY stand INSIDE the head and be fine, because it won't do the bite attack.

Dweblenod: Hehe yep! It would probably be a smart choice to add the cone to the canUse() method to also prevent them from repeatedly trying to attack where they cannot

seriousCoolness: Okay, now you will get damaged when standing inside the basilisk's neck, while still keeping the hit-cones mostly accurate to the attack animations. I also made the tail-swipe have a slightly wider arc, so that it isn't _too_ easy to stay in the basilisk's blind-spots.

PR #465, opened 2023-04-17
Left-hand punch attack for ogres

seriousCoolness:
My previous change where I added the cone-based attack collision left a huge hole in the Ogre's offense, since they only have a right-hand punch animation, so I've quickly made a mirrored version of the punch to allow them to not be such pushovers.


kirderf1:
Realized that this didn't get added to the changelog. @Dweblenod could you look at it?

kirderf1:
Nevermind, I'll add it myself now

PR #467, opened 2023-04-18
fixed otherside and 5 crafting recipe was swapped

LunyaticCat:


Dweblenod (commented):
Use runData and then this will be ready to merge. However if you feel like it could you also make sure all the cassettes have grist costs? I believe some of the ones that were added before did not have one

kirderf1:
Adding the changes to the changelog would also be nice

PR #468, opened 2023-04-19
Worldgen datagen

kirderf1:
Does two things:
- Lands now properly get worldgen features from the world/datapack registry, rather than directly the builtin registry.
- Features and structures are now defined through datagen. Registry objects have been replaced by resource keys.

Both of these prepares us for the eventual move to 1.19.3.


kirderf1:
Realised that I should also do this for density functions and noise parameters. Marking the PR as ready to review once that's done as well.

Dweblenod:
Will have to test!

Dweblenod:
Checked out 8 random lands, did not see any issues with the generation!

However does this PR change the noise settings? I know Caldwell made a PR that did so recently but I didnt remember the changes being this pronounced

seriousCoolness:
> However does this PR change the noise settings? I know Caldwell made a PR that did so recently but I didnt remember the changes being this pronounced

That happens on the main branch too. I've found that sometimes the terrain changes are more pronounced for some reason. I think certain seeds just create more variation than others.

src/main/java/com/mraof/minestuck/blockentity/redstone/StructureCoreBlockEntity.java
```diff
@@ -33,7 +32,7 @@ public class StructureCoreBlockEntity extends BlockEntity
 	private ActionType actionType;
 	private int shutdownRange = 32;
 	private boolean hasWiped = false;
-	public final Structure[] SOLVABLE_STRUCTURES = new Structure[]{MSConfiguredStructures.FROG_TEMPLE.get()};
+	public final Structure[] SOLVABLE_STRUCTURES = new Structure[]{};
```
Dweblenod, line 35: Why remove frog temples?

Dweblenod (commented):
A question, is it possible with this setup to define the density/noise functions on a land by land basis?

src/main/java/com/mraof/minestuck/blockentity/redstone/StructureCoreBlockEntity.java
kirderf1: I got the impression that the feature was already not working as it should (since it was getting the structure from the wrong registry), as well as being an incomplete and unused feature. When I then changed `MSConfiguredStructures.FROG_TEMPLE` from a registry object to a resource key, I didn't bother changing/fixing it for the previously mentioned reasons, and instead just removed the structure from the array.

kirderf1:
> A question, is it possible with this setup to define the density/noise functions on a land by land basis?

These changes make no difference on that regard.

Allowing lands to do more with density functions would involve designing an API through which land types can do so, and change `LandGenSettings.makeLandNoiseRouter()` accordingly.

kirderf1:
@Dweblenod if I could take your attention back to this PR whenever you have time.

Dweblenod (commented):
Explored multiple lands with no issues. Will continue checking through code!

src/main/java/com/mraof/minestuck/data/worldgen/MSPlacedFeatureProvider.java
```diff
@@ -0,0 +1,305 @@
...
+				worldGenModifiers(CountPlacement.of(UniformInt.of(100, 200)), HeightRangePlacement.uniform(VerticalAnchor.bottom(), VerticalAnchor.absolute(60)))));
+		consumer.accept(OCEANIC_POINTED_DRIPSTONE, placed(features, CaveFeatures.POINTED_DRIPSTONE,
+				worldGenModifiers(CountPlacement.of(UniformInt.of(164, 256)), HeightRangePlacement.uniform(VerticalAnchor.bottom(), VerticalAnchor.absolute(48)))));
+		consumer.accept(LUSH_CAVES_VEGETATION, placed(features, CaveFeatures.MOSS_PATCH,
+				worldGenModifiers(CountPlacement.of(125), HeightRangePlacement.uniform(VerticalAnchor.bottom(), VerticalAnchor.absolute(64)), EnvironmentScanPlacement.scanningFor(Direction.DOWN, BlockPredicate.solid(), BlockPredicate.ONLY_IN_AIR_PREDICATE, 12), RandomOffsetPlacement.vertical(ConstantInt.of(1)))));
+		consumer.accept(LUSH_CAVES_CEILING_VEGETATION, placed(features, CaveFeatures.MOSS_PATCH_CEILING,
+				worldGenModifiers(CountPlacement.of(125), HeightRangePlacement.uniform(VerticalAnchor.bottom(), VerticalAnchor.absolute(64)), EnvironmentScanPlacement.scanningFor(Direction.UP, BlockPredicate.solid(), BlockPredicate.ONLY_IN_AIR_PREDICATE, 12), RandomOffsetPlacement.vertical(ConstantInt.of(-1)))));
```
Dweblenod, line 142: Still trying to determine if it was already like that or if there was a mistranslation, but this generates at a visually higher rate than intended

kirderf1: It looks to be the same as the [original](https://github.com/mraof/Minestuck/blob/948469d4e40072ef9e85476fa113069946e6803a/src/main/java/com/mraof/minestuck/world/gen/feature/MSPlacedFeatures.java#L116-L117), so I assume that it was already like that.

Dweblenod (approved):
I cant think of any further suggestions or questions!

kirderf1:
Alright, I'll be looking at merging this sometime tomorrow. (Unless you get to it before me)

PR #469, opened 2023-04-20
Grist set interfaces

kirderf1:
Changes the structure of grist set classes to introduce two interfaces implemented by multiple classes.
Some notable changes:
- `GristSet` -> `MutableGristSet`. Used when you need a grist set that should be changeable, such as with `addGrist()` or `setGrist()`.
- `ImmutableGristSet` -> `DefaultImmutableGristSet`. For grist sets that shouldn't be changed. No longer extends the mutable variant, so it no longer has any `addGrist()` or `setGrist()` functions in the first place. However with a new interface, this should only be used when you want to create a new grist set.
- New interface `GristSet`. Typically used as a function parameter when you only need to check the content of the grist set.
- New interface `ImmutableGristSet`. Extends `GristSet` without any new methods. Suitable when you want to be sure that the grist set can't change. Suitable when a grist set will remain for a longer time, but should not change, for example as fields in a recipe.
- `GristAmount` now implements `ImmutableGristSet`, making it useable in any functions that accepts a `GristSet`.

Since these changes are rather wide, it'd be good if this gets priority before too many people start making conflicting changes.
But also because it changes a lot of how grist sets are used, feedback on these changes are important as well.


kirderf1:
The `NonNegativeGristSet` is one part which I do not have a fully clear image of whether it's good as is or should be done in a different way.

I should probably move the buffer read/write functions to somewhere more appropriate.

One idea I've had is to make `GristCache` and `GristGutter` implement `GristSet`, redirecting function calls to their inner grist set. I'm not sure whether or not there's enough of a point in doing so.

kirderf1:
Oh and a reminder to self that I was thinking of shortening `gristSet.addGrist()` to `gristSet.add()`

src/main/java/com/mraof/minestuck/alchemy/MutableGristSet.java
```diff
@@ -0,0 +1,203 @@
+package com.mraof.minestuck.alchemy;
+
+import com.mojang.serialization.Codec;
+
+import java.util.*;
+import java.util.function.Supplier;
+
+public class MutableGristSet implements GristSet
```
Dweblenod, line 8: Mind adding javadoc to explain when it should be used?

src/main/java/com/mraof/minestuck/alchemy/NonNegativeGristSet.java
```diff
@@ -1,58 +1,54 @@
 package com.mraof.minestuck.alchemy;
 
-import net.minecraft.nbt.ListTag;
+import com.mojang.serialization.Codec;
 
-public class NonNegativeGristSet extends GristSet
+import java.util.List;
+
+public class NonNegativeGristSet extends MutableGristSet
```
Dweblenod, line 7: Mind adding javadoc to explain when this should be used?

Dweblenod (approved):
Did a few minutes of testing with runServer and in single player and havent found any bugs resulting from the changes.

The changes to inheritance and the implementation of an interface now make sense to me so I have no issues

PR #474, opened 2023-04-26
rung name tweak

KingBalDoro:
This pr is meant to add new rung names for both aspects and potentially classpects.


KingBalDoro:
not finished, more names are required before I can runData

kirderf1:
I'm considering closing this PR considering its current state. Does that seem fine?

KingBalDoro:
this PR was intended to add new, specialized rung names. the scope of this intent far outweighed my abilities at the time; i think it should be fully closed as it's not quite worth the amount of effort it'd take, but i will be keeping these ideas noted for when i become capable of easily completing these.

PR #413, opened 2023-02-04
Updates to lotus flower

Dweblenod:
Incorporating model/animation changes made by Riotmode to the lotus flower and halving the default regrow value (now takes 5 minutes to regrow instead of 10)

PR #477, opened 2023-04-29
Fix for TPZ command not accepting TRANSportalizers

Dweblenod:
Now uses an instanceof check(TransportalizerBlock) instead of checking for a specific Block. If this approach is not desired, then I can revert it and do a check for TRANSportalizer and transportalizer specifically

Includes changelog changes from PR #413


kirderf1:
The transportalizer itself checks that the destination has a transportalizer block entity rather than a transportalizer block. It might be sensible to take the same approach here.

Dweblenod:
> The transportalizer itself checks that the destination has a transportalizer block entity rather than a transportalizer block. It might be sensible to take the same approach here.

Sounds reasonable! Im also noticing that the command is case sensitive despite the transportalizers converting names to all caps. Is this intentional?

kirderf1:
> Im also noticing that the command is case sensitive despite the transportalizers converting names to all caps. Is this intentional?

I can't imagine why that would be intentional, so I don't think it is.

Dweblenod:
> 

Added a .toUpperCase(Locale.ROOT) to the end of line 36 to fix the case sensitivity but also noticing now that the tpz command does not accept addresses with special characters like ♥. I cannot think of an immediate solution to this

kirderf1:
Doesn't look like you've pushed the upper case fix yet

Dweblenod:
Woops yea you are right. I think I got distracted trying to imagine a solution to the non-alphanumeric issue

src/main/java/com/mraof/minestuck/command/TransportalizerCommand.java
```diff
@@ -33,7 +33,7 @@
 	public static void register(CommandDispatcher<CommandSourceStack> dispatcher)
 	{
 		dispatcher.register(Commands.literal("tpz").requires(commandSource -> commandSource.hasPermission(2))
-				.then(Commands.argument("code", StringArgumentType.word()).executes(context -> teleport(context.getSource(), Collections.singleton(context.getSource().getEntityOrException()), StringArgumentType.getString(context, "code")))
+				.then(Commands.argument("code", StringArgumentType.word()).executes(context -> teleport(context.getSource(), Collections.singleton(context.getSource().getEntityOrException()), StringArgumentType.getString(context, "code").toUpperCase(Locale.ROOT)))
```
kirderf1, line 36: I think that it would be better to perform the conversion to upper case inside of the `teleport()` function instead to make it less easy to miss when reading through the code, as well as make less things happen in this part of the code where there's already a bunch going on. Plus, by putting it inside of `teleport()`, it'll also apply to the missed call on line 37.

CHANGELOG.md
```diff
@@ -10,9 +10,13 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 
 - Cinnamon Sword
 
+### Fixed
+
+- TPZ command now accepts TRANSportalizers and is no longer case sensitive
```
kirderf1, line 15: Just a personal taste, but I'd list them as two separate things

kirderf1:
```suggestion
- /tpz command now works with TRANSportalizers
- /tpz command code argument is no longer case sensitive
```

PR #478, opened 2023-05-01
Intellibeam lazerstation

KingBalDoro:
this PR adds:
- intellibeam lazerstation block
- unreadable item tag

additional contributors, **blankMushroom**


KingBalDoro:
as of now i have yet to run data

src/main/java/com/mraof/minestuck/block/machine/IntellibeamLaserstationBlock.java
```diff
@@ -0,0 +1,114 @@
...
+	public BlockEntity newBlockEntity(BlockPos pos, BlockState state)
+	{
+		return new HolopadBlockEntity(pos, state);
+	}
+	
+	@Nullable
+	@Override
+	public <T extends BlockEntity> BlockEntityTicker<T> getTicker(Level level, BlockState state, BlockEntityType<T> placedType)
+	{
+		return level.isClientSide ? BlockUtil.checkTypeForTicker(placedType, MSBlockEntityTypes.HOLOPAD.get(), HolopadBlockEntity::clientTick) : null;
```
Dweblenod, line 63: switch to using MSBlockEntityTypes.INTELLIBEAM_LASERSTATION.get() and IntellibeamLaserstationBlockEntity and then remove the import for the holopad

src/main/java/com/mraof/minestuck/block/machine/IntellibeamLaserstationBlock.java
```diff
@@ -0,0 +1,114 @@
...
+	{
+		super(builder);
+		registerDefaultState(defaultBlockState().setValue(HAS_CARD, false));
+	}
+	
+	@Nullable
+	@Override
+	public BlockEntity newBlockEntity(BlockPos pos, BlockState state)
+	{
+		return new HolopadBlockEntity(pos, state);
```
Dweblenod, line 56: switch to using IntellibeamLaserstationBlockEntity

CHANGELOG.md
```diff
@@ -9,14 +9,16 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
...
+- Unreadable tag
 
 ### Fixed
 
 - Basilisks can now despawn again
 
 ### Contributors for this release
 
-- Vivian-Zane, rose_bushes_
+- Vivian-Zane, rose_bushes_, Doro
```
Dweblenod, line 21: Make sure to include blankMushroom

src/main/java/com/mraof/minestuck/block/machine/IntellibeamLaserstationBlock.java
KingBalDoro: this whole function doesn't actually need to exist

src/main/java/com/mraof/minestuck/block/machine/IntellibeamLaserstationBlock.java
KingBalDoro: done

CHANGELOG.md
KingBalDoro: done

src/main/java/com/mraof/minestuck/block/MSBlockShapes.java
```diff
@@ -100,6 +100,7 @@
 	public static CustomVoxelShape SMALL_ALCHEMITER = new CustomVoxelShape(new double[]{0,0,0,16,4,16}, new double[]{1,4,1,15,6,15}, new double[]{2,6,2,14,8,14}, new double[]{0.5,4,6.75,1,5,9.25}, new double[]{6.75,4,0.5,9.25,5,1}, new double[]{15,4,6.75,15.5,5,9.25}, new double[]{6.75,4,15,9.25,5,15.5}, new double[]{13.5,4,13.5,15.5,9,15.5}, new double[]{11.5,4,14.5,12.5,15,15.5}, new double[]{13,11,14.5,14,15,15.5}, new double[]{14.5,11,14.5,15.5,15,15.5}, new double[]{12,15,14.5,13.5,15.5,15.5}, new double[]{13.5,10.5,14.5,15,11,15.5}, new double[]{14,15,14,16,17,16});
 	
 	//Misc Machines
+	public static CustomVoxelShape INTELLIBEAM_LAZERSTATION = new CustomVoxelShape(new double[]{2,0,12,9,7,17}, new double[]{0,0,6,10,8,12}, new double[]{10,0,4,19,8,12}, new double[]{-1,7,7,20,10,11}, new double[]{4,7,11,9,18,15}, new double[]{2,18,9,12,24,15}, new double[]{-3,18,8,2,25,16});
```
kirderf1, line 103: Looks like it's a bit larger than a block. Are you confident that it is within the size that vanilla can handle correctly for one block?

src/main/java/com/mraof/minestuck/data/tag/MinestuckItemTagsProvider.java
```diff
@@ -126,6 +126,7 @@ protected void addTags()
 		tag(BUGS).add(BUG_ON_A_STICK.get(), CHOCOLATE_BEETLE.get(), CONE_OF_FLIES.get(), GRASSHOPPER.get(), CICADA.get(), JAR_OF_BUGS.get());
 		tag(CONSORT_SNACKS).add(Items.COOKIE).addTag(BUGS);
 		tag(CREATIVE_SHOCK_RIGHT_CLICK_LIMIT).add(Items.CHORUS_FRUIT);
+		tag(UNREADABLE).add(CLIENT_DISK.get(), SERVER_DISK.get(), GUTTER_THUMB_DRIVE.get(), GUTTER_BALL.get(), CAPTCHA_CARD.get(), CUEBALL.get(), BLACK_QUEENS_RING.get(), WHITE_QUEENS_RING.get(), BLACK_KINGS_SCEPTER.get(), WHITE_KINGS_SCEPTER.get()).add(Items.DRAGON_EGG).add(Items.DRAGON_HEAD).add(Items.DRAGON_BREATH).add(Items.NETHER_STAR).add(Items.COMMAND_BLOCK).add(Items.COMMAND_BLOCK_MINECART).add(Items.CHAIN_COMMAND_BLOCK).add(Items.REPEATING_COMMAND_BLOCK).add(Items.END_CRYSTAL);
```
kirderf1, line 129: Having captcha cards in this tag would remove one method of getting spare captcha cards for alchemy during the early game. Has this been considered properly?

src/main/java/com/mraof/minestuck/blockentity/machine/PunchDesignixBlockEntity.java
```diff
@@ -114,8 +117,13 @@ private void handleKeyboardClick(ServerPlayer player)
 		{
 			if(AlchemyHelper.hasDecodedItem(heldStack))
 			{
-				ItemStack output;
-				if(AlchemyHelper.isPunchedCard(getCard()))	//|| combination
+				ItemStack output = AlchemyHelper.getDecodedItem(heldStack);
+				if(output.is(MSTags.Items.UNREADABLE) && !AlchemyHelper.isReadableCard(output))
+				{
+					player.sendSystemMessage(Component.translatable("On closer inspection, the code on the back seems to be almost unreadable").withStyle(ChatFormatting.BOLD));
```
kirderf1, line 123: You're using a string literal instead of a translation key for the translatable message. You should change it to use a translation key which you then specify the text for in `MinestuckEnUsLanguageProvider`.

src/main/java/com/mraof/minestuck/block/MSBlockShapes.java
kirderf1: Or actually, is the shape defined here even used for the lazerstation?

src/main/java/com/mraof/minestuck/block/machine/IntellibeamLaserstationBlock.java
```diff
@@ -0,0 +1,115 @@
...
+			intellibeam.onRightClick(player);
+		return InteractionResult.SUCCESS;
+	}
+	
+	@Override
+	public void playerWillDestroy(Level level, BlockPos pos, BlockState state, Player player)
+	{
+		IntellibeamLaserstationBlockEntity intelibeamEntity = (IntellibeamLaserstationBlockEntity) level.getBlockEntity(pos);
+		
+		if(intelibeamEntity != null && !level.isClientSide)
```
kirderf1, line 85:
```suggestion
		if(!level.isClientSide && level.getBlockEntity(pos) instanceof IntellibeamLaserstationBlockEntity intelibeamEntity)
```
Using an instanceof check here should be safer than casting the block entity without a check.

src/main/java/com/mraof/minestuck/blockentity/IntellibeamLaserstationBlockEntity.java
```diff
@@ -0,0 +1,198 @@
...
+
+import static com.mraof.minestuck.alchemy.AlchemyHelper.createEncodedItem;
+
+public class IntellibeamLaserstationBlockEntity extends BlockEntity
+{
+	protected ItemStack card = ItemStack.EMPTY;
+	protected int EXP_LEVEL_CAPACITY = 10;
+	protected int EXPERIENCE_LEVEL = 0;
+	protected float SOUND_SCALER = 0F;
+	protected int WAIT_TIMER = 0;
```
kirderf1, line 32: Should these be private perhaps?

kirderf1, line 32: These are named like constants, but they lack the attributes of a constant. Are they actually meant to be regular fields?

src/main/java/com/mraof/minestuck/blockentity/IntellibeamLaserstationBlockEntity.java
```diff
@@ -0,0 +1,198 @@
...
+	}
+	
+	public boolean hasCard()
+	{
+		return !this.getCard().isEmpty();
+	}
+	
+	public void insertCard(ItemStack card)
+	{
+		if (card.getItem() == MSItems.CAPTCHA_CARD.get() || card.isEmpty())
```
kirderf1, line 108:
```suggestion
		if (card.is(MSItems.CAPTCHA_CARD.get()) || card.isEmpty())
```

src/main/java/com/mraof/minestuck/blockentity/IntellibeamLaserstationBlockEntity.java
```diff
@@ -0,0 +1,198 @@
...
+	
+	public ItemStack getCard()
+	{
+		return this.card;
+	}
+	
+	public static boolean isDecoded(ItemStack decodedItem)
+	{
+		return decodedItem.getItem() == MSItems.CAPTCHA_CARD.get() && decodedItem.hasTag() && decodedItem.getTag().getBoolean("decoded");
+	};
```
kirderf1, line 153: Looks a lot like `AlchemyHelper.isReadableCard()`
Perhaps you should use that one instead?

src/main/java/com/mraof/minestuck/blockentity/IntellibeamLaserstationBlockEntity.java
```diff
@@ -0,0 +1,198 @@
...
+	
+	public IntellibeamLaserstationBlockEntity(BlockPos pos, BlockState state)
+	{
+		super(MSBlockEntityTypes.INTELLIBEAM_LASERSTATION.get(), pos, state);
+	}
+	
+	public void onRightClick(Player player)
+	{
+		ItemStack heldItem = player.getMainHandItem();
+		ItemStack taggedItem = AlchemyHelper.getDecodedItem(heldItem);
```
kirderf1, line 42: `taggedItem` is a bit unclear. How about renaming it to `itemInsideCard`?

src/main/java/com/mraof/minestuck/blockentity/IntellibeamLaserstationBlockEntity.java
```diff
@@ -0,0 +1,198 @@
...
+		
+		if(isDecoded(taggedItem))
+		{
+			return;
+		}
+		if(card.isEmpty() && !this.hasCard() && !isDecoded(taggedItem))
```
kirderf1, line 48: `card.isEmpty()` and `!this.hasCard()` has the same meaning, right? If so, remove one of them.

src/main/java/com/mraof/minestuck/blockentity/IntellibeamLaserstationBlockEntity.java
```diff
@@ -0,0 +1,198 @@
...
+import net.minecraft.world.Containers;
+import net.minecraft.world.InteractionHand;
+import net.minecraft.world.entity.player.Player;
+import net.minecraft.world.item.ItemStack;
+import net.minecraft.world.level.Level;
+import net.minecraft.world.level.block.Block;
+import net.minecraft.world.level.block.entity.BlockEntity;
+import net.minecraft.world.level.block.state.BlockState;
+
+import static com.mraof.minestuck.alchemy.AlchemyHelper.createEncodedItem;
```
kirderf1, line 24: I would advice against this specific use of a static import

kirderf1 (changes requested):
Should the mini punch designix have a check similar to what was added to the regular punch designix?

KingBalDoro:
> Should the mini punch designix have a check similar to what was added to the regular punch designix?

yes i think so

src/main/java/com/mraof/minestuck/block/MSBlockShapes.java
KingBalDoro: it's not being used for anything at the moment because I wanted to put that off until I got the block most of the way finished, but after that I was thinking of changing the model to be 1, closer to the comic, and 2, better able to fit in minecraft

src/main/java/com/mraof/minestuck/data/tag/MinestuckItemTagsProvider.java
KingBalDoro: this was the decision that had come to after a discussion between caldw3ll, blank, and to my knowledge riotmode, and I. I'm honoring it although I don't remember the actual purpose behind why specifically the captcha card was to be made unreadable. as it stands I have it locked away as something i can give or take once the feature is ready for packaging and release

src/main/java/com/mraof/minestuck/blockentity/machine/PunchDesignixBlockEntity.java
KingBalDoro: can do

src/main/java/com/mraof/minestuck/blockentity/IntellibeamLaserstationBlockEntity.java
KingBalDoro: done

src/main/java/com/mraof/minestuck/blockentity/IntellibeamLaserstationBlockEntity.java
KingBalDoro: done

src/main/java/com/mraof/minestuck/blockentity/IntellibeamLaserstationBlockEntity.java
KingBalDoro: removed

src/main/java/com/mraof/minestuck/blockentity/IntellibeamLaserstationBlockEntity.java
KingBalDoro: why

KingBalDoro: aactually i see why, nevermind

src/main/java/com/mraof/minestuck/data/tag/MinestuckItemTagsProvider.java
seriousCoolness: Iirc we decided to make the captchalogue cards unreadable in order to emphasize their use as the “fuel” for alchemy, along with some other reasons related to the role of cruxite dowels in alchemy. I think the conversation was had in Homestuck-observations? Take this with a grain of salt btw, cause I could very well be misremembering this.

seriousCoolness: Wait, it was actually in debate-and-discussion.

kirderf1: Ah yes, making them purposefully more sparse in the early game for some gameplay purpose, instead relying on the recipe with raw cruxite and paper.
However that does make me wonder, while cruxite will always be available in lands through mining, I don't think the same is true for paper or sugar cane. That could be problematic if we expect all players to need to use that recipe.

KingBalDoro: we could create another more accommodating recipe, but that raises the question of whether or not it's worth it considering how needlessly gated it might make alchemy feel.

seriousCoolness: I think maybe we could give sugarcane/paper more ways to be obtained. Add alchemy combinations like seaweed && stick = Sugarcane? Stick || wheat = sugarcane? 
We could also make it so that you can substitute the 8 paper in the captchalogue card recipe with one of the playing-card items you can sometimes get from imp-dungeons. Or maybe we could make it so that you can craft captchalogue cards using item frames?

akisephila: :?? < i do think sugar cane and paper, and more items in general, having alchemy recipes and alternative ways to be obtained in the medium is a good idea. what i dont understand is making captcha cards unreadable. 
:?? < it doesnt gate access to them, just adds two more steps; alchemizing the ingredients in their crafting recipes. at that point, why not just eliminate the middle man of having to alchemize the ingredients first?

KingBalDoro: i'm in agreement with candy. it doesn't do much for captcha cards to be made unreadable considering their composite parts are already generally inexpensive and can be made in bulk.

seriousCoolness: I thought raw cruxite wasn’t alchemizable? Even if it is though, making the captchalogue card’s code unreadable at first provides a sense of progression to the player, by removing a redundant step once they progress far enough. It’s like how the jumper block extension removes certain redundant steps. Removing the redundancies is a reward for progression.

KingBalDoro: agreed, but the thing i'm questioning is whether or not it's worth it to make the alchemy experience bumpier when the speed with which you get to the same place remains the same. if we make cards unreadable, I think we should also look for ways to slow down the player's ability to get them to then emphasize their general value as a standalone item to an even greater extent. This would also smooth out the pacing of alchemy if making cards through crafting becomes the one available method.

seriousCoolness: I think the speed of the process as a whole being basically the same is a good thing. It means that the restriction won’t bother players all that much, but despite not being bothered by it too much, they’ll still get that little hit of dopamine once they find out how to get around it using the intellibeam laserstation.

seriousCoolness: That’s fair.

src/main/java/com/mraof/minestuck/item/MSItems.java
```diff
@@ -653,7 +653,7 @@
 	public static final RegistryObject<BlockItem> MINI_ALCHEMITER = registerBlockItem(MSBlocks.MINI_ALCHEMITER, MSItemGroup.MAIN);
 	public static final RegistryObject<BlockItem> MINI_PUNCH_DESIGNIX = registerBlockItem(MSBlocks.MINI_PUNCH_DESIGNIX, MSItemGroup.MAIN);
 	public static final RegistryObject<BlockItem> HOLOPAD = registerBlockItem(MSBlocks.HOLOPAD, MSItemGroup.MAIN);
-	
+	public static final RegistryObject<BlockItem> INTELLIBEAM_LAZERSTATION = registerBlockItem(MSBlocks.INTELLIBEAM_LASERSTATION, MSItemGroup.MAIN);
```
Dweblenod, line 656: change z to s

src/main/java/com/mraof/minestuck/data/MinestuckEnUsLanguageProvider.java
```diff
@@ -1405,6 +1406,8 @@ protected void addTranslations()
 		add("message.dice.3", "You rolled a 4.");
 		add("message.dice.4", "You rolled a 5.");
 		add("message.dice.5", "You rolled a 6.");
+		
+		add("punch_designix.code_rejected", "On closer inspection, the code on the back seems to be almost unreadable");
```
Dweblenod, line 1410: Create a public static final value in PunchDesignixBlockEntity with the location name. You can add it after line 1748 in this file, modeling it off of the RedstoneClockBlockEntity example that exists on line 1748

src/main/java/com/mraof/minestuck/data/MinestuckEnUsLanguageProvider.java
```diff
@@ -507,6 +507,7 @@ protected void addTranslations()
 		MSBlocks.ALCHEMITER.forEachBlock(block -> add(block, "Alchemiter"));
 		MSBlocks.PUNCH_DESIGNIX.forEachBlock(block -> add(block, "Punch Designix"));
 		addBlock(MSBlocks.HOLOPAD, "Holopad");
+		addBlock(MSBlocks.INTELLIBEAM_LASERSTATION, "Intellibeam Lazerstation");
```
Dweblenod, line 510: Change the z to an s

src/main/java/com/mraof/minestuck/blockentity/IntellibeamLaserstationBlockEntity.java
```diff
@@ -0,0 +1,196 @@
...
+import net.minecraft.world.level.block.entity.BlockEntity;
+import net.minecraft.world.level.block.state.BlockState;
+
+public class IntellibeamLaserstationBlockEntity extends BlockEntity
+{
+	protected ItemStack card = ItemStack.EMPTY;
+	protected int EXP_LEVEL_CAPACITY = 10;
+	protected int EXPERIENCE_LEVEL = 0;
+	protected float SOUND_SCALER = 0F;
+	protected static int WAIT_TIMER = 0;
```
Dweblenod, line 32: Two things about this section:
1. SCREAMING_SNAKE_CASE is usually only appropriate for static final variables, which remain unchanged while the code is running. The naming scheme for EXP_LEVEL_CAPACITY is appropriate due to this definition, and as a result should be changed from protected to private static final or public static final
2. The use of the protected access modifier means that these values can be edited by other Classes in this package. While I dont know if their usage was intentional, private would be most appropriate considering the nature of them and the fact that they are only currently used in this class

Heres a useful explanation of how the common modifiers work: https://stackoverflow.com/questions/5083232/usage-patterns-for-private-static-final-public-abstract-keywords-in-java

src/main/java/com/mraof/minestuck/blockentity/IntellibeamLaserstationBlockEntity.java
```diff
@@ -0,0 +1,196 @@
...
+import net.minecraft.world.Containers;
+import net.minecraft.world.InteractionHand;
+import net.minecraft.world.entity.player.Player;
+import net.minecraft.world.item.ItemStack;
+import net.minecraft.world.level.Level;
+import net.minecraft.world.level.block.Block;
+import net.minecraft.world.level.block.entity.BlockEntity;
+import net.minecraft.world.level.block.state.BlockState;
+
+public class IntellibeamLaserstationBlockEntity extends BlockEntity
```
Dweblenod, line 26: Use the Reformat Code option on this file

src/main/java/com/mraof/minestuck/block/machine/IntellibeamLaserstationBlock.java
```diff
@@ -0,0 +1,102 @@
...
+import net.minecraft.world.level.block.state.StateDefinition;
+import net.minecraft.world.level.block.state.properties.BooleanProperty;
+import net.minecraft.world.phys.BlockHitResult;
+import net.minecraft.world.phys.shapes.CollisionContext;
+import net.minecraft.world.phys.shapes.VoxelShape;
+
+import javax.annotation.Nullable;
+import java.util.Map;
+
+public class IntellibeamLaserstationBlock extends MachineBlock implements EntityBlock
```
Dweblenod, line 31: remove the double empty lines in this file

Dweblenod (commented):
Looking through the system as it is now, its pretty functional! A major change that I think is both desirable and achievable is to use a GUI instead of a relying on right click functionality. It would make it much more obvious what this does, what is needed to make it work, and allow the player to keep track of what theyve already done. Im down to write the code needed to incorporate the GUI should you want that assistance

src/main/java/com/mraof/minestuck/alchemy/AlchemyHelper.java
```diff
@@ -85,6 +85,11 @@ else if(tag.contains("contentSize") && tag.getInt("contentSize") >= 1)
 		
 	}
 	
+	public static boolean isReadableCard(ItemStack readItem)
+	{
+		return readItem.is(MSItems.CAPTCHA_CARD.get()) && readItem.hasTag() && readItem.getTag().getBoolean("decoded");
```
Dweblenod, line 90: I recommend also allowing it to be readable if its already been punched. Both because that means the player can already reverse engineer the code via interpreting the punch pattern and because the part of the card with the captcha on it likely has holes punched through it

Dweblenod:
Nothing game breaking but it appears you will be brought into the swimming/prone player state if you jump up while pressed against the machine when on the machines right side

src/main/java/com/mraof/minestuck/blockentity/IntellibeamLaserstationBlockEntity.java
```diff
@@ -0,0 +1,196 @@
...
+			this.level.playSound(null, this.worldPosition, MSSoundEvents.INTELLIBEAM_LAZERSTATION_REMOVE_CARD.get(), SoundSource.BLOCKS, 0.5F, 0.1F);
+			WAIT_TIMER = 10;
+			return;
+		}
+		if(card.isEmpty() && !AlchemyHelper.isReadableCard(card))
+		{
+			tryInsertCard(heldItem);
+			WAIT_TIMER = 10;
+			return;
+		}
```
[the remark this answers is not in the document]
KingBalDoro: a dedicated GUI, isn't needed since it's so conceptually simple. but if clarity is a worry then a small HWYLA styled gui that appears next to your cursor for each of the alchemy related items could do the trick. simply having an "experience gauge" would be all it takes. although i'm of the opinion that by the time the player will be able to get the intellibeam laserstation, Experience won't be much of a concern. So letting players stumble into an answer through experimentation here seems fine. but the aforementioned GUI would be a nice aesthetic touch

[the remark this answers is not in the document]
KingBalDoro: besides that, empty captcha cards do have codes, they're just 0000000.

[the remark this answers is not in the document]
KingBalDoro: the third return should prevent readable cards from actually being able to have experience added

src/main/java/com/mraof/minestuck/alchemy/AlchemyHelper.java
KingBalDoro: done

KingBalDoro: decided against it, although it make since functionally i will not be making this one function do two things like that and will instead just include the isPunched() function in all calls involving isReadable()

KingBalDoro: decided against this as i'm tired and don't want to think about how to make that work

src/main/java/com/mraof/minestuck/block/machine/IntellibeamLaserstationBlock.java
KingBalDoro: done

src/main/java/com/mraof/minestuck/blockentity/IntellibeamLaserstationBlockEntity.java
KingBalDoro: done

src/main/java/com/mraof/minestuck/blockentity/IntellibeamLaserstationBlockEntity.java
KingBalDoro: done

src/main/java/com/mraof/minestuck/data/MinestuckEnUsLanguageProvider.java
KingBalDoro: done

src/main/java/com/mraof/minestuck/alchemy/AlchemyHelper.java
Dweblenod: return isPunchedCard(readItem) || (readItem.is(MSItems.CAPTCHA_CARD.get()) && readItem.hasTag() && readItem.getTag().getBoolean("decoded"));

Dweblenod: The above just builds the check into this boolean. If the item in question meets either the punched card criteria or the criteria you created, it will succeed

KingBalDoro: i was referring to my first comment with my second comment

Dweblenod:
There are a couple larger scale changes I think the Intellibeam could benefit from. Because of that, I believe it would be easier for me to create a PR to your branch with my own code as opposed to making a long series of suggestions or straight up dictations. Would it be acceptable to make a PR that will be merged into this, or would you like to do the work directly?

KingBalDoro:
i've gotten all that i need out of the PR as i really could experience wise and i think the tedium of you working through me isn't worth it when all that's left to do is implement it. if you're fine with doing that extra leg work, it would be appreciated on my end.

CHANGELOG.md
```diff
@@ -9,6 +9,9 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 ### Added
 
 - Cinnamon Sword
+- Intellibeam Lazerstation
+- Unreadable tag
+- Decoded tag
```
kirderf1, line 14: These listings are too vague for the changelog

src/main/java/com/mraof/minestuck/blockentity/IntellibeamLaserstationBlockEntity.java
kirderf1: It is now named `cardWithItem`, which is not accurate. From my understanding, the item in the players hand is expected to be a card with an item inside. When that stack is "decoded", you get the item that's inside this card.

CHANGELOG.md
kirderf1: Looks like that change has been lost since then?

src/main/java/com/mraof/minestuck/alchemy/AlchemyHelper.java
Dweblenod: Does this mean you think punched cards should still require decoding?

Dweblenod:
Code will be added by a new PR for technical reasons, with due crediting

PR #480, opened 2023-05-05
Anthvil and power hub

Dweblenod:
Anthvil: Grist and uranium powered machine that repairs durability on items

Power hub: slowly generates uranium fuel that can be passed on to a uranium powered device above it (anthvil/sendificator/cookalyzer). Power hub now used to create hubtop


Dweblenod:
Still waiting on art feedback, but code is ready for review

src/main/java/com/mraof/minestuck/blockentity/machine/UraniumPowered.java
```diff
@@ -0,0 +1,13 @@
+package com.mraof.minestuck.blockentity.machine;
+
+/**
+ * Used to keep track of block entities that store uranium fuel and apply external sources of fuel.
+ */
+public interface UraniumPowered
+{
+	void addFuel(short fuelAmount);
+	
+	boolean canBeRefueled();
```
kirderf1, line 10: Should this function be part of the interface for any reason?

src/main/java/com/mraof/minestuck/blockentity/machine/AnthvilBlockEntity.java
```diff
@@ -0,0 +1,259 @@
...
+/**
+ * Mends an item at the cost of uranium power and grist. The grist used is the one with the highest value impact (weighted against build grist)
+ */
+public class AnthvilBlockEntity extends MachineProcessBlockEntity implements MenuProvider, UraniumPowered
+{
+	public static final String TITLE = "container.minestuck.anthvil";
+	public static final short MAX_FUEL = 128;
+	public static final short MEND_FUEL_COST = 5;
+	
+	private final ProgressTracker progressTracker = new ProgressTracker(ProgressTracker.RunType.ONCE, 0, this::setChanged, this::contentsValid);
```
kirderf1, line 43: Using a progress tracker seems like overkill when it doesn't need to track progress and doesn't need to track if it's looping. Perhaps you could simplify it out?

src/main/java/com/mraof/minestuck/blockentity/machine/AnthvilBlockEntity.java
```diff
@@ -0,0 +1,259 @@
...
+	
+	public static void attemptMend(AnthvilBlockEntity anthvil, ServerPlayer player)
+	{
+		Level level = anthvil.level;
+		ItemStack slotStack = anthvil.itemHandler.getStackInSlot(0);
+		GristCache playerCache = GristCache.get(player);
+		
+		ImmutableGristSet pickedGrist = mendingGrist(anthvil, playerCache, level, slotStack);
+		
+		if(!GristAmount.EMPTY.equals(pickedGrist))
```
kirderf1, line 144:
```suggestion
		if(!pickedGrist.isEmpty())
```

src/main/java/com/mraof/minestuck/blockentity/machine/AnthvilBlockEntity.java
```diff
@@ -0,0 +1,259 @@
...
+			if(fullSet != null && !fullSet.isEmpty())
+			{
+				GristAmount pickedGrist = getUsedGrist(fullSet);
+				
+				if(anthvil.fuel >= MEND_FUEL_COST && playerCache.canAfford(pickedGrist))
+					return pickedGrist;
+			}
+		}
+		
+		return GristAmount.EMPTY;
```
kirderf1, line 181:
```suggestion
		return GristSet.EMPTY;
```

src/main/java/com/mraof/minestuck/blockentity/machine/AnthvilBlockEntity.java
```diff
@@ -0,0 +1,259 @@
...
+		}
+	}
+	
+	public static void attemptMend(AnthvilBlockEntity anthvil, ServerPlayer player)
+	{
+		Level level = anthvil.level;
+		ItemStack slotStack = anthvil.itemHandler.getStackInSlot(0);
+		GristCache playerCache = GristCache.get(player);
+		
+		ImmutableGristSet pickedGrist = mendingGrist(anthvil, playerCache, level, slotStack);
```
kirderf1, line 142:
```suggestion
		GristSet pickedGrist = mendingGrist(anthvil, playerCache, level, slotStack);
```
No reason for this to be immutable, leave the type as `GristSet` instead

src/main/java/com/mraof/minestuck/blockentity/machine/AnthvilBlockEntity.java
```diff
@@ -0,0 +1,259 @@
...
+	{
+		if(level != null && !slotStack.isEmpty() && slotStack.isRepairable() && slotStack.isDamageableItem() && slotStack.isDamaged())
+		{
+			GristSet fullSet = GristCostRecipe.findCostForItem(slotStack, null, false, level);
+			
+			if(fullSet != null && !fullSet.isEmpty())
+			{
+				GristAmount pickedGrist = getUsedGrist(fullSet);
+				
+				if(anthvil.fuel >= MEND_FUEL_COST && playerCache.canAfford(pickedGrist))
```
kirderf1, line 176: Too many responsibilities in this function. I suggest separating the conditions for if the machine can repair the item from the code that looks up the repair cost, moving some of the conditions into `attemptMend()`.

src/main/java/com/mraof/minestuck/blockentity/machine/AnthvilBlockEntity.java
```diff
@@ -0,0 +1,259 @@
...
+	
+	/**
+	 * Only handles refreshing uranium
+	 */
+	private void processContents()
+	{
+		if(canBeRefueled())
+		{
+			//checks for a uranium itemstack in the right item slot, increases the fuel value if some is found and then removes one count from the fuel stack
+			if(itemHandler.getStackInSlot(1).is(ExtraForgeTags.Items.URANIUM_CHUNKS))
```
kirderf1, line 127: I suggest combining them into one if-statement with two conditions

src/main/java/com/mraof/minestuck/blockentity/machine/AnthvilBlockEntity.java
```diff
@@ -0,0 +1,259 @@
...
+			{
+				//Refill fuel
+				addFuel((short) FUEL_INCREASE);
+				itemHandler.extractItem(1, 1, false);
+			}
+		}
+	}
+	
+	public static void attemptMend(AnthvilBlockEntity anthvil, ServerPlayer player)
```
kirderf1, line 136: It may be a good idea to use some early returns in this function.

src/main/java/com/mraof/minestuck/blockentity/machine/AnthvilBlockEntity.java
```diff
@@ -0,0 +1,259 @@
...
+		List<GristAmount> gristAmounts = fullSet.asAmounts();
+		GristAmount pickedGrist = new GristAmount(GristTypes.BUILD.get(), 1);
+		
+		if(gristAmounts.size() > 1) //establishes which GristAmount in the set has the highest impact and uses that
+		{
+			for(GristAmount grist : fullSet.asAmounts())
+			{
+				double gristValue = grist.getValue();
+				if(grist.type() == GristTypes.BUILD.get())
+					gristValue /= 10; //reduces build grist value
```
kirderf1, line 198: Make a local/private function for this modified value and you could use a stream:
`fullSet.asAmounts().stream().max(Comparator.comparingDouble(AnthvilBlockEntity::modifiedGristValue))`
(this expression could then replace essentially the entirety of `getUsedGrist()`)

kirderf1: Oh actually, that wouldn't be true for items with a negative grist cost, but you can be the judge of that.

src/main/java/com/mraof/minestuck/network/AnthvilPacket.java
```diff
@@ -0,0 +1,41 @@
+package com.mraof.minestuck.network;
+
+import com.mraof.minestuck.blockentity.machine.AnthvilBlockEntity;
+import net.minecraft.core.BlockPos;
+import net.minecraft.network.FriendlyByteBuf;
+import net.minecraft.server.level.ServerPlayer;
+
+public class AnthvilPacket implements PlayToServerPacket
+{
+	private final BlockPos pos;
```
kirderf1, line 10: Instead of sending a position through the packet (to a block that the player may or may not reach), I suggest instead making use of `ContainerLevelAccess` from `machineContainerMenu.getPosition()` (See `SendificatorPacket` for an example)

src/main/java/com/mraof/minestuck/client/gui/MSScreenFactories.java
```diff
@@ -105,6 +110,11 @@ public static void displayAlchemiterScreen(AlchemiterBlockEntity blockEntity)
 		Minecraft.getInstance().setScreen(new AlchemiterScreen(blockEntity));
 	}
 	
+	/*public static void displayAnthvilScreen(AnthvilMenu screenContainer, Inventory inv, Component titleIn, AnthvilBlockEntity blockEntity)
+	{
+		Minecraft.getInstance().setScreen(new AnthvilScreen(MSMenuTypes.ANTHVIL.get(), inv, titleIn, blockEntity));
+	}*/
```
kirderf1, line 116: To be removed?

src/main/java/com/mraof/minestuck/client/gui/AnthvilScreen.java
```diff
@@ -0,0 +1,124 @@
...
+import net.minecraft.world.entity.player.Inventory;
+import net.minecraft.world.item.ItemStack;
+import net.minecraft.world.level.Level;
+import net.minecraftforge.client.gui.widget.ExtendedButton;
+
+import javax.annotation.ParametersAreNonnullByDefault;
+import java.util.List;
+
+@ParametersAreNonnullByDefault
+public class AnthvilScreen extends MachineScreen<AnthvilMenu>
```
kirderf1, line 26: You don't need to use `MachineScreen` unless you're gonna make use of the `goButton` field (and even then, I'd suggest against it)
```suggestion
public class AnthvilScreen extends AbstractContainerScreen<AnthvilMenu>
```

kirderf1 (changes requested):

src/main/java/com/mraof/minestuck/blockentity/machine/UraniumPowered.java
```diff
@@ -0,0 +1,13 @@
+package com.mraof.minestuck.blockentity.machine;
+
+/**
+ * Used to keep track of block entities that store uranium fuel and apply external sources of fuel.
+ */
+public interface UraniumPowered
```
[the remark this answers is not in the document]
Dweblenod: Agreed! I was considering a method of converting our uranium power/fuel to FE but it seemed too big for the scope of the PR

src/main/java/com/mraof/minestuck/blockentity/machine/UraniumPowered.java
Dweblenod: I suppose its not necessary, I believe I included it as a way to ensure standardization despite no external uses

src/main/java/com/mraof/minestuck/blockentity/machine/AnthvilBlockEntity.java
Dweblenod: I kept it in under the uncritical assumption that it was necessary for classes extending MachineProcessBlockEntity. When I messed with the progress tracker before it prevented the GUI from showing up. Do you have any suggestions or examples of how to simplify it?

src/main/java/com/mraof/minestuck/blockentity/machine/PowerHubBlockEntity.java
```diff
@@ -0,0 +1,75 @@
...
+import net.minecraft.world.level.block.entity.BlockEntity;
+import net.minecraft.world.level.block.state.BlockState;
+
+public class PowerHubBlockEntity extends BlockEntity
+{
+	public static final String POWER_PROMPT = "block.minestuck.power_hub.power_prompt";
+	
+	public static final short MAX_POWER = 256;
+	
+	private short power = 0;
```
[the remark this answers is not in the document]
Dweblenod: Made a method to handle it

src/main/java/com/mraof/minestuck/blockentity/machine/AnthvilBlockEntity.java
Dweblenod: Migrated parts of code to attemptMend and incorporated them as an early return checks

src/main/java/com/mraof/minestuck/blockentity/machine/AnthvilBlockEntity.java
kirderf1: You could cut out its usage bit by bit. First cut out use of the `MachinePacket.SetRunning` packet and trigger the refuel as part of `AnthvilPacket` and `AnthvilBlockEntity.attemptMend()` instead of through the progress tracker. After that, you could change `AnthvilMenu` to not require the progress tracker data by making it independent from `MachineContainerMenu` (that does mean that you need to handle the `ContainerLevelAccess` directly in `AnthvilMenu`, but that's no big deal. Once it's no longer used to trigger the refuel, and no longer synced through the menu, you can just remove it. You should be able to remove the ticker for the block entity at that point as well.

Does that seem reasonable with how refueling should happen?

src/main/java/com/mraof/minestuck/blockentity/machine/AnthvilBlockEntity.java
```diff
@@ -0,0 +1,264 @@
...
+		GristAmount defaultGrist = new GristAmount(GristTypes.BUILD.get(), 1);
+		
+		if(gristAmounts.size() > 1) //if theres more than one, establishes which GristAmount has the highest impact and uses that
+		{
+			GristAmount pickedGrist = fullSet.asAmounts().stream().max(Comparator.comparingDouble(AnthvilBlockEntity::getModifiedGristValue)).orElse(defaultGrist);
+			return new GristAmount(pickedGrist.type(), 1);
+		} else if(gristAmounts.get(0).type() != GristTypes.BUILD.get())
+			return new GristAmount(gristAmounts.get(0).type(), 1);
+		else
+			return defaultGrist;
```
kirderf1, line 206:
```suggestion
		GristType pickedGrist = fullSet.asAmounts().stream().max(Comparator.comparingDouble(AnthvilBlockEntity::getModifiedGristValue)).map(GristAmount::type).orElse(GristTypes.BUILD.get());
		return new GristAmount(pickedGrist, 1);
```
The stream will work equally well independent on the number of grist amounts, so this stream could replace the other if-cases too. Added another step in this suggestion too.

src/main/java/com/mraof/minestuck/blockentity/machine/UraniumPowered.java
kirderf1: In the respective machines, it is tied to the machines capability to consume one fuel item. I don't think it'll do us any good to ensure that future uranium powered machines also has this capability. I recommend removing it from the interface.

src/main/java/com/mraof/minestuck/client/gui/AnthvilScreen.java
```diff
@@ -0,0 +1,123 @@
...
+	private static final ResourceLocation BACKGROUND_TEXTURE = new ResourceLocation(Minestuck.MOD_ID, "textures/gui/anthvil.png");
+	private static final ResourceLocation FUEL_STATUS = new ResourceLocation(Minestuck.MOD_ID, "textures/gui/progress/uranium_level.png");
+	
+	private static final int FUEL_STATUS_X = 133;
+	private static final int FUEL_STATUS_Y = 7;
+	private static final int STATUS_WIDTH = 35;
+	private static final int STATUS_HEIGHT = 39;
+	private static final int BUTTON_X = 12;
+	private static final int BUTTON_Y = 38;
+	private final AnthvilBlockEntity anthvilBE;
```
kirderf1, line 36: Looks like this is only here for level access. I suggest removing it and getting the level from the minecraft instance attached to the screen instead (`MiniAlchemiterScreen`).

src/main/java/com/mraof/minestuck/blockentity/machine/PowerHubBlockEntity.java
```diff
@@ -0,0 +1,85 @@
...
+			changePower(true);
+		}
+	}
+	
+	public short getPower()
+	{
+		return power;
+	}
+	
+	private void changePower(boolean increasePower)
```
kirderf1, line 76: The boolean seems like extra complexity that doesn't serve much purpose. I suggest an int argument instead (or short?), so that you pass in 1 or -1 instead of true or false.

kirderf1 (changes requested):

src/main/java/com/mraof/minestuck/blockentity/machine/AnthvilBlockEntity.java
```diff
@@ -0,0 +1,264 @@
...
+	 */
+	private static ImmutableGristSet mendingGrist(GristCache playerCache, Level level, ItemStack slotStack)
+	{
+		GristSet fullSet = GristCostRecipe.findCostForItem(slotStack, null, false, level);
+		
+		if(fullSet != null && !fullSet.isEmpty())
+		{
+			GristAmount pickedGrist = getUsedGrist(fullSet);
+			
+			if(playerCache.canAfford(pickedGrist))
```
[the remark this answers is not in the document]
Dweblenod: Yup!

src/main/java/com/mraof/minestuck/blockentity/machine/AnthvilBlockEntity.java
Dweblenod: Have done so now!

src/main/java/com/mraof/minestuck/blockentity/machine/AnthvilBlockEntity.java
```diff
@@ -0,0 +1,264 @@
...
+		Level level = anthvil.level;
+		ItemStack slotStack = anthvil.itemHandler.getStackInSlot(0);
+		GristCache playerCache = GristCache.get(player);
+		
+		if(level == null || !isMendableItem(slotStack) || !hasEnoughFuel(anthvil))
+			return;
+		
+		GristSet pickedGrist = mendingGrist(playerCache, level, slotStack);
+		
+		if(!pickedGrist.isEmpty())
```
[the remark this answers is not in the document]
Dweblenod: Reorganized this

src/main/java/com/mraof/minestuck/inventory/AnthvilMenu.java
```diff
@@ -0,0 +1,111 @@
...
+{
+	private static final int uraniumInputX = 143;
+	private static final int uraniumInputY = 54;
+	private static final int itemInputX = 81;
+	private static final int itemInputY = 35;
+	
+	private final DataSlot fuelHolder;
+	private final ContainerLevelAccess levelAccess;
+	
+	public AnthvilMenu(int windowId, Inventory playerInventory, FriendlyByteBuf buffer)
```
kirderf1, line 28: The machine position is no longer needed for this menu and can be cleared out. The buffer argument can then be cleared out along with it (meaning that you'd skip the `IContainerFactory` cast in `MSMenuTypes` and the position argument when calling `NetworkHooks.openScreen()` from the block class.

src/main/java/com/mraof/minestuck/network/AnthvilPacket.java
```diff
@@ -0,0 +1,37 @@
...
+	
+	@Override
+	public void execute(ServerPlayer player)
+	{
+		AbstractContainerMenu playerContainer = player.containerMenu;
+		if(playerContainer instanceof AnthvilMenu anthvilMenu)
+		{
+			anthvilMenu.getPosition().execute((level, machinePos) -> {
+				AnthvilBlockEntity blockEntity = (AnthvilBlockEntity) level.getBlockEntity(machinePos);
+				if(blockEntity != null)
```
kirderf1, line 30: An instanceof check would be safer  than a cast followed by a null check. (It doesn't really matter much here specifically, but it's a good style to have for getting block entities generally)

kirderf1 (changes requested):

Dweblenod:
made changes

kirderf1 (approved):

PR #484, opened 2023-06-21
Union Buster

Estellarium:
-Feat(MSItems.java): Added Union Buster sword (concept by Riotmode) and corresponding assets/lang; added grist recipe but not crafting recipe, or loot. (Currently unobtainable)
-- Note: Parameters may need tweaking, especially durability and damage.
-Feat(OnHitEffect.java): Added "SPREAD_SWEEP" effect, derivative from "SWEEP", which measures a lateral dot product to spread hit enemies more than vanilla knockback. As it is only used for the Union Buster, it is also hardcoded to add the Union Bust effect.
-Feat(MSEffects.java): Added the Union Bust effect (incl. class) - it is meant to, when ticked, forcibly move the afflicted entity away from other entities with the effect. It was not implemented (currently commented), but it is also meant to disable mounting and calls for reinforcements.


Dweblenod (commented):
As mentioned in Discord, this is good work! 

IntelliJ has a tool called Reformat Code which helps us standardize our formatting, I may suggest using it now and again. Try using it on SuspicionEffect!

src/main/java/com/mraof/minestuck/item/MSItems.java
```diff
@@ -102,6 +103,7 @@ public class MSItems
...
 	public static final RegistryObject<Item> CINNAMON_SWORD = REGISTER.register("cinnamon_sword", () -> new WeaponItem(new WeaponItem.Builder(MSItemTypes.CANDY_TIER, 4, -2.4F).efficiency(1.0F).set(MSItemTypes.SWORD_TOOL).add(OnHitEffect.SWEEP).add(OnHitEffect.SET_CANDY_DROP_FLAG), new Item.Properties().tab(MSItemGroup.WEAPONS)));
+	public static final RegistryObject<Item> UNION_BUSTER = REGISTER.register("union_buster", () -> new WeaponItem(new WeaponItem.Builder(Tiers.GOLD, 7, -3.5F).efficiency(1.0f).set(MSItemTypes.SWORD_TOOL).add(OnHitEffect.SweepMultiEffect(OnHitEffect.SPREADING_KNOCKBACK, OnHitEffect.enemyPotionEffect(() -> new MobEffectInstance(MSEffects.SUSPICION.get(), 1200, 1, false, false)))), new Item.Properties().defaultDurability(202).tab(MSItemGroup.WEAPONS).rarity(Rarity.UNCOMMON)));
```
Dweblenod, line 106: The speed being what it is is an intelligent choice given the properties of the weapon. However the damage could probably do with an increase. In terms of DPS its not very good and its slow draw could make it difficult to swap between it and another weapon so ideally it should be able to hold its own ground. I could see it going as high as 11 in-game damage if this is meant to be a high tier weapon, but even just a bump of +1 to +2 would also be fine. If you want it to be a more readily accessible weapon then I have no complaints

src/main/java/com/mraof/minestuck/item/weapon/OnHitEffect.java
```diff
@@ -314,6 +340,41 @@ static OnHitEffect enemyPotionEffect(Supplier<MobEffectInstance> effect)
 		return (stack, target, attacker) -> target.addEffect(effect.get());
 	}
 	
+	//Nearly identical to Sweep, but it applies an optional array of effects
+	static OnHitEffect SweepMultiEffect(OnHitEffect... effects)
```
Dweblenod, line 344: Make the first letter lowercase. We use the lowercase variant of camelCase in such conditions

src/main/java/com/mraof/minestuck/effects/SuspicionEffect.java
```diff
@@ -0,0 +1,81 @@
...
+import net.minecraft.world.entity.Entity;
+import net.minecraft.world.entity.LivingEntity;
+import net.minecraftforge.event.entity.EntityMountEvent;
+import net.minecraftforge.event.entity.living.ZombieEvent;
+import net.minecraftforge.eventbus.api.SubscribeEvent;
+import net.minecraftforge.fml.common.Mod;
+import org.jetbrains.annotations.NotNull;
+
+@Mod.EventBusSubscriber(modid = Minestuck.MOD_ID, bus = Mod.EventBusSubscriber.Bus.FORGE)
+public class SuspicionEffect extends MobEffect
```
Dweblenod, line 15: Its a pretty simple class, but please add a short javadoc style description. We are trying to get into the habit of commenting our codebase to make things as quick to understand as possible

src/main/java/com/mraof/minestuck/data/MinestuckEnUsLanguageProvider.java
```diff
@@ -1364,6 +1364,8 @@ protected void addTranslations()
 		addItemTooltip(MSItems.GUTTER_BALL, "A hardened crystal in the shape of grist");
 		addItemExtra(MSItems.GUTTER_BALL, "press_shift", "Press §eSHIFT§r for more info");
 		addItemExtra(MSItems.GUTTER_BALL, "desc", "Carved from a common ore, it is believed by the natives to be a kind of food to the yet-born deities.");
+		addItem(MSItems.UNION_BUSTER, "Union Buster");
+		addItemTooltip(MSItems.UNION_BUSTER, "Does quite a lot of damage, both societal AND material!");
```
Dweblenod, line 1368: Each weapon category has a region of this file it occupies, where you should put this is between what is currently line 673 and 674

src/main/resources/assets/minestuck/textures/item/union_buster.png
Dweblenod, line 1: While it maps correctly and doesnt display differently in game, this file is actually 256x256 pixels across instead of the typical 16x16. When saved from discord you may have taken a version that was increased in size for display reasons

Dweblenod (commented):
Seems to work as intended both in single player and with a dedicated server, good job!

src/main/java/com/mraof/minestuck/effects/SuspicionEffect.java
```diff
@@ -0,0 +1,136 @@
...
+	private boolean hasInitialized = false;
+	/**
+	 * Used to enable the effect at all, blocking the {@link #applyEffectTick(LivingEntity, int)} and {@link #onMount(EntityMountEvent)} functions if false.
+	 * It is true if the afflicted entity is not an {@link net.minecraft.world.entity.decoration.ArmorStand} instance and doesn't have the "NoAI" NBT.
+	 */
+	private boolean isPushable = true;
+	/**
+	 * Used to enable the pushing effect on flying entities, as an alternative for the {@link LivingEntity#isOnGround()} check.
+	 */
+	private boolean isFlying = false;
```
kirderf1, line 39: Instances of `MobEffect` are global/akin to a singleton object, and NOT specific to a specific entity. You should therefore not use entity-specific info in fields here as the same effect and fields will be used for ALL entities. Try instead defining reusable functions that figure out if a given entity is pushable or is flying.

src/main/java/com/mraof/minestuck/effects/SuspicionEffect.java
```diff
@@ -0,0 +1,136 @@
...
+	
+	/**
+	 * Cancels out attempts to call for reinforcements.
+	 * Examples: {@link net.minecraft.world.entity.monster.Zombie#hurt(DamageSource, float)}, {@link net.minecraft.world.entity.monster.ZombifiedPiglin#hurt(DamageSource, float)})
+	 */
+	@SubscribeEvent
+	public void summonAttempt(ZombieEvent.SummonAidEvent event)
+	{
+		if(event.getEntity().hasEffect(this)) event.setCanceled(true);
+	}
```
kirderf1, line 134: The event handlers does not appear to be properly registered. Try changing them to be static instead: https://cdn.discordapp.com/attachments/437002009216942086/558994977430962196/eventsubscriber.png

src/main/java/com/mraof/minestuck/item/weapon/OnHitEffect.java
```diff
@@ -180,6 +180,32 @@ public interface OnHitEffect
...
+		double expDot = 1.5 * Math.pow(1 - (Math.abs(dot)), 0.3);
+		//Knockback direction vector
+		double knockbackX = -dirToEntityX + (pRightCos * Mth.sign(dot) * expDot);
+		double knockbackZ = -dirToEntityZ + (pRightSin * Mth.sign(dot) * expDot);
+		dirMagnitude = Math.sqrt((knockbackX * knockbackX + knockbackZ * knockbackZ));
+		knockbackX /= dirMagnitude;
+		knockbackZ /= dirMagnitude;
+		
+		target.knockback(0.7f, knockbackX, knockbackZ);
+	};
```
kirderf1, line 207: Would it make sense to use some vector objects to make the math here more readable?

src/main/java/com/mraof/minestuck/item/weapon/OnHitEffect.java
```diff
@@ -314,6 +340,41 @@ static OnHitEffect enemyPotionEffect(Supplier<MobEffectInstance> effect)
...
+								effect.onHit(stack, livingEntity, attacker);
+						}
+					}
+					
+					playerAttacker.level.playSound(null, playerAttacker.getX(), playerAttacker.getY(), playerAttacker.getZ(), SoundEvents.PLAYER_ATTACK_SWEEP, playerAttacker.getSoundSource(), 1.0F, 1.0F);
+					playerAttacker.sweepAttack();
+				}
+			}
+		};
+	}
```
kirderf1, line 376: The function could do with some transformations to improve readability. To start, I recommend [inverting](https://youtu.be/CFRhGnuXG-4) the first two if-statements and making an early return. After that, it could be useful to extract some part of the code inside the for-loop or if-statement conditions.

kirderf1 (changes requested):

src/main/java/com/mraof/minestuck/effects/SuspicionEffect.java
Estellarium: Ah, the usage of "MobEffectInstance" threw me off. Fixed, but I'm concerned that now there might be too many checks per tick. It depends on the performance impact.

src/main/java/com/mraof/minestuck/effects/SuspicionEffect.java
Estellarium: And I had changed them before because I thought they weren't working because they were static. Fixed.

src/main/java/com/mraof/minestuck/item/weapon/OnHitEffect.java
Estellarium: I was avoiding adding an import to the file. Change made as it is visually better, and does not require the redundant manual math. Same to the vector math used in SuspicionEffect.java.

src/main/java/com/mraof/minestuck/item/weapon/OnHitEffect.java
Estellarium: Inversions were made, I did not feel extractions were necessary. Inversions to reduce nesting were made to SuspicionEffect.applyEffectTick(LivingEntity, int) as well.

src/main/java/com/mraof/minestuck/item/weapon/OnHitEffect.java
```diff
@@ -180,6 +181,35 @@ public interface OnHitEffect
...
 	};
 	
+	/**
+	 * A knockback effect that does a right-vector (relative to the player's orientation) and direction to enemy dot calculation (relative to the player's orientation) to spread enemies horizontally via knockback.
+	 */
+	OnHitEffect SPREADING_KNOCKBACK = (stack, target, attacker) -> {
+		
+		//Attacker's right direction and position
+		double pRightSin = Math.sin(attacker.getYRot() * ((float) Math.PI / 180F) - Math.PI / 2.0);
+		double pRightCos = Math.cos(attacker.getYRot() * ((float) Math.PI / 180F) - Math.PI / 2.0);
```
kirderf1, line 191: A vector could ideally be used for these as well.

src/main/java/com/mraof/minestuck/item/weapon/OnHitEffect.java
```diff
@@ -180,6 +181,35 @@ public interface OnHitEffect
...
+		//Other entity position
+		double eX = target.getX(), eZ = target.getZ();
+		
+		//Direction vector player -> entity
+		Vec2 dirToEntity = new Vec2((float) (eX - pX), (float) (eZ - pZ));
+		dirToEntity = dirToEntity.normalized();
+		
+		//Dot product of direction and player's right
+		double dot = (pRightSin * dirToEntity.x) + (pRightCos * dirToEntity.y);
+		double expDot = 1.5 * Math.pow(1 - (Math.abs(dot)), 0.3);
```
kirderf1, line 202: If I understand the equation for `expDot` correctly, it's meant to make the sideways knockback stronger for entities directly in front of the player. Is that right?

kirderf1 (commented):
Good work so far! Here are a couple more things I noted. I also suggest running code formatting on both files once you're done.

src/main/java/com/mraof/minestuck/effects/SuspicionEffect.java
```diff
@@ -0,0 +1,129 @@
...
+@Mod.EventBusSubscriber(modid = Minestuck.MOD_ID, bus = Mod.EventBusSubscriber.Bus.FORGE)
+public class SuspicionEffect extends MobEffect
+{
+	/**
+	 * Used to enable the effect at all, blocking the {@link #applyEffectTick(LivingEntity, int)} and {@link #onMount(EntityMountEvent)} functions if false.
+	 * It is true if the afflicted entity is not an {@link net.minecraft.world.entity.decoration.ArmorStand}, doesn't have the "NoAI" NBT, or isn't a {@link net.minecraft.world.entity.player.Player}.
+	 */
+	private static boolean isPushable(Entity entity) {
+		return !((entity instanceof ArmorStand) ||
+				(entity.getPersistentData().contains("NoAI", 99)) ||
```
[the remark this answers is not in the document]
Estellarium: From testing it seemed to work, implemented this as it's better

src/main/java/com/mraof/minestuck/effects/SuspicionEffect.java
```diff
@@ -0,0 +1,129 @@
...
+	}
+	
+	/**
+	 * Cancels out attempts to ride (or be ridden by) an affected entity
+	 */
+	@SubscribeEvent
+	public static void onMount(EntityMountEvent event)
+	{
+		Entity mountEntity = event.getEntityMounting();
+		if(isPushable(mountEntity) && ((LivingEntity) mountEntity).hasEffect(MSEffects.SUSPICION.get()))
```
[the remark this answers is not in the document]
Estellarium: Fixed. The "if" statement, as it was identical, was also extracted into its own function.

src/main/java/com/mraof/minestuck/item/weapon/OnHitEffect.java
Estellarium: Change made, replaced instances of Vec2 with Vec3 for some more readability (use z component instead of y) and more built-in functions, which also reduced the amount of local variables.

src/main/java/com/mraof/minestuck/item/weapon/OnHitEffect.java
Estellarium: Exactly. It adds a force based on the relative position, the addition is weaker when closer to the player's right, that is, it becomes closer to default knockback.

src/main/java/com/mraof/minestuck/item/weapon/OnHitEffect.java
```diff
@@ -180,6 +181,38 @@ public interface OnHitEffect
...
+		
+		//Dot product of direction and attacker's right
+		double dot = attackerRight.dot(dirToTarget);
+		double expDot = 1.5 * Math.pow(1 - (Math.abs(dot)), 0.3);
+		
+		//Knockback direction vector
+		Vec3 knockback = new Vec3(
+				(float) (-dirToTarget.x + (attackerRight.x * Mth.sign(dot) * expDot)),
+				0,
+				(float) (-dirToTarget.z + (attackerRight.z * Mth.sign(dot) * expDot)));
```
kirderf1, line 209:
```suggestion
		Vec3 knockback = dirToTarget.scale(-1).add(attackerRight.scale(Mth.sign(dot) * expDot));
```
And consider replacing `Mth.sign(dot) * expDot` using:
```java
double sideKnockbackFactor = Mth.sign(dot) * 1.5 * Math.pow(1 - (Math.abs(dot)), 0.3);
```

src/main/java/com/mraof/minestuck/item/weapon/OnHitEffect.java
```diff
@@ -314,6 +347,47 @@ static OnHitEffect enemyPotionEffect(Supplier<MobEffectInstance> effect)
...
+			float attackDamage = (float) playerAttacker.getAttribute(Attributes.ATTACK_DAMAGE).getValue();
+			float sweepEnchantMod = 1.0F + EnchantmentHelper.getSweepingDamageRatio(playerAttacker) * attackDamage;
+			
+			for(LivingEntity livingEntity : playerAttacker.level.getEntitiesOfClass(
+					LivingEntity.class, target.getBoundingBox().inflate(2.0D, 0.25D, 2.0D)))
+			{
+				if(!(livingEntity != playerAttacker
+						&& !playerAttacker.isAlliedTo(livingEntity)
+						&& (!(livingEntity instanceof ArmorStand) || !((ArmorStand) livingEntity).isMarker())
+						&& playerAttacker.distanceToSqr(livingEntity) < 9.0D))
```
kirderf1, line 378: Double-inversions might make this harder to read, so I'd like to suggest this instead:
```suggestion
				if(livingEntity == playerAttacker
						|| playerAttacker.isAlliedTo(livingEntity)
						|| livingEntity instanceof ArmorStand stand && stand.isMarker()
						|| playerAttacker.distanceToSqr(livingEntity) >= 9.0D)
```

src/main/java/com/mraof/minestuck/item/weapon/OnHitEffect.java
Estellarium: Applied, as well as some other likewise simplifications.

src/main/java/com/mraof/minestuck/item/weapon/OnHitEffect.java
Estellarium: Applied changes to both if statements that had this structure.

kirderf1 (approved):
Alright, good job! I expect that to be all from me.

PR #485, opened 2023-06-25
Tag expansion and shadewood plank crafting expansion

Dweblenod:
Expanding Minestuck items/blocks to fit more appropriate tags maintained by Minecraft/Forge. Added shadewood planks to the appropriate tags so that they can be used as normal planks


kirderf1 (commented):
Looks good so far. The biggest question I have is if to really put spoons in the shovel tag or not. What are your thoughts on it?

Dweblenod:
> Looks good so far. The biggest question I have is if to really put spoons in the shovel tag or not. What are your thoughts on it?

Im thinking that anything themed similar to a tool that also has the corresponding tool functionality seems worth including. Perhaps my mind would change if I knew some general usages of the item tag. I am open to removing them if you think it seems like a poor choice!

kirderf1:
I'm fine with leaving them there then.

Dweblenod:
I have nothing else I can think to add to this

kirderf1 (approved):

kirderf1 (changes requested):
Ah right, changelog should be updated too.

PR #489, opened 2023-07-03
Intellibeam Laserstation and captcha codes

Dweblenod:
Continuation of @KingBalDoro's PR of the same name #478.

Gives each registered item its own captchalogue code which can now be typed into a Punch Designix. Certain items cannot be read by default and require the assistance of the Intellibeam Laserstation to read. XP is expended to decode those captchas.

Captcha for a given item is essentially random but tied to seed, so two different worlds with the same seed will have the same captchas for the same items but two different worlds with different seeds will have different captchas for the same items.

Intellibeam model originally from blankMushroom, with remodels by Doro


src/main/java/com/mraof/minestuck/alchemy/CardCaptchas.java
```diff
@@ -0,0 +1,272 @@
...
+	{
+		this.seedRandomizer = new Random(level.getSeed());
+		generateBackupCaptchas();
+	}
+	
+	@SubscribeEvent
+	public static void serverStarted(ServerStartedEvent event)
+	{
+		CardCaptchas captchas = new CardCaptchas(event.getServer().overworld());
+		captchas.iterateThroughRegistry();
```
Dweblenod, line 57: Perhaps I am overthinking it, but I feel like creating an instance like this is awkward. However I am not sure what a better solution would be

src/main/java/com/mraof/minestuck/blockentity/IntellibeamLaserstationBlockEntity.java
```diff
@@ -0,0 +1,237 @@
...
+		
+		for(int i = 0; i < numberOfBars; i++)
+		{
+			if(i < getCardItemExperience())
+				stringChars[i] = '▮';
+			else
+				stringChars[i] = '▯';
+		}
+		
+		return '<' + new String(stringChars) + '>'; //if two levels of experience have been invested, it will look like this: "<▮▮▯▯▯▯▯▯▯▯>"
```
Dweblenod, line 121: In game, the bar characters are smaller than the text. Not sure if there is a simple solution to make them similarly sized like how they appear in Github and IntelliJ

src/main/java/com/mraof/minestuck/blockentity/machine/PunchDesignixBlockEntity.java
```diff
@@ -104,32 +134,56 @@ else if (!player.getInventory().add(getCard()))
...
-		if(heldStack.getItem() != MSItems.CAPTCHA_CARD.get())
-			return;    //Not a valid item in hand
 		
-		if(getCard().getItem() == MSItems.CAPTCHA_CARD.get())
+		if(player.level.isClientSide)
 		{
-			if(AlchemyHelper.hasDecodedItem(heldStack))
+			//Go to screen if no valid item in hand or crouching
+			//TODO if the player is holding a captcha card and is crouching, it wont open the screen
+			if(player.isShiftKeyDown() || heldStack.getItem() != MSItems.CAPTCHA_CARD.get())
```
Dweblenod, line 145: Could use help with this issue

kirderf1: What might be happening is that the item gets priority by shifting, and then consumes the action by returning a non-pass result from the `use()` function. Does that seem correct?

One way to deal with it is to override `item.doesSneakBypassUse()`. Think about if that'd be approriate.

Dweblenod: Sounds logical! If thats the case, it does not sound appropriate. Extra crouch functionality is low priority in my book

src/main/java/com/mraof/minestuck/alchemy/CardCaptchas.java
```diff
@@ -0,0 +1,272 @@
...
+	
+	public static final String AVAILABLE_CHARACTERS = "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789!?";
+	public static final String EMPTY_CARD_CAPTCHA = "00000000";
+	
+	private static final BiMap<String, String> REGISTRY_MAP = HashBiMap.create();
+	
+	private final Set<String> backupCaptchas = new HashSet<>();
+	private final Set<String> backupCaptchasMaster = new HashSet<>(); //entries are not removed from here
+	
+	private final Random seedRandomizer;
```
kirderf1, line 45: Minecraft had semi-recently switched to use their own random class, and I think that it makes some sense to follow along. (See `RandomSource`)

src/main/java/com/mraof/minestuck/alchemy/CardCaptchas.java
```diff
@@ -0,0 +1,272 @@
...
+ */
+@Mod.EventBusSubscriber(modid = Minestuck.MOD_ID, bus = Mod.EventBusSubscriber.Bus.FORGE)
+public class CardCaptchas
+{
+	//TODO use of the character "#" not canonically accurate and suggests there are more punch holes in a captchalogue card then there actually are
+	
+	public static final String AVAILABLE_CHARACTERS = "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789!?";
+	public static final String EMPTY_CARD_CAPTCHA = "00000000";
+	
+	private static final BiMap<String, String> REGISTRY_MAP = HashBiMap.create();
```
kirderf1, line 40: Ah, I see that this field is used for captchas both client-side and server-side. Have you thought anything about thread safety? Is it being used in a sensible manner to avoid typical threading issues?

src/main/java/com/mraof/minestuck/alchemy/CardCaptchas.java
kirderf1: One option is to make `iterateThroughRegistry()` static, and create your object for backup captchas inside there instead.

src/main/java/com/mraof/minestuck/block/MSBlockShapes.java
```diff
@@ -100,6 +100,7 @@ public class MSBlockShapes
 	public static CustomVoxelShape SMALL_ALCHEMITER = new CustomVoxelShape(new double[]{0,0,0,16,4,16}, new double[]{1,4,1,15,6,15}, new double[]{2,6,2,14,8,14}, new double[]{0.5,4,6.75,1,5,9.25}, new double[]{6.75,4,0.5,9.25,5,1}, new double[]{15,4,6.75,15.5,5,9.25}, new double[]{6.75,4,15,9.25,5,15.5}, new double[]{13.5,4,13.5,15.5,9,15.5}, new double[]{11.5,4,14.5,12.5,15,15.5}, new double[]{13,11,14.5,14,15,15.5}, new double[]{14.5,11,14.5,15.5,15,15.5}, new double[]{12,15,14.5,13.5,15.5,15.5}, new double[]{13.5,10.5,14.5,15,11,15.5}, new double[]{14,15,14,16,17,16});
 	
 	//Misc Machines
+	public static CustomVoxelShape INTELLIBEAM_LAZERSTATION = new CustomVoxelShape(new double[]{1,6,0,6,12,4}, new double[]{1,0,1,6,7,6}, new double[]{4,0,12,11,7,16}, new double[]{0,0,6,8,8,12}, new double[]{8,0,4,15,8,12}, new double[]{-1,7,7,16,10,11}, new double[]{6,7,12,11,16,16}, new double[]{6,15,11,11,21,15}, new double[]{4,20,8,14,26,14}, new double[]{-1,20,7,4,27,15}, new double[]{5,19,5,12,22,10}, new double[]{6,22,6,11,24,8});
```
kirderf1, line 103: > Looks like it's a bit larger than a block. Are you confident that it is within the size that vanilla can handle correctly for one block?

From the previous PR, still relevant. I'm sure that there's some limit to how large a shape that the game handles without a problem, but I don't know how large that is or whether this is within that. Bringing this up mostly because it's larger than the shape of a fence.

kirderf1 (changes requested):

src/main/java/com/mraof/minestuck/alchemy/CardCaptchas.java
Dweblenod: How it is created and how it is distributed from server side to client side has been a concern of mine. I am familiarized with the concept of thread safety but do not actually know what is needed to prevent potential issues on that front. Any recommendations?

src/main/java/com/mraof/minestuck/block/MSBlockShapes.java
Dweblenod: Found a potential resource we could use to discuss this in Discord

src/main/java/com/mraof/minestuck/blockentity/machine/PunchDesignixBlockEntity.java
Dweblenod: Removed

src/main/java/com/mraof/minestuck/alchemy/CardCaptchas.java
```diff
@@ -0,0 +1,274 @@
...
+	public CardCaptchas(ServerLevel level)
+	{
+		this.seedRandomizer = RandomSource.create(level.getSeed());
+		generateBackupCaptchas();
+	}
+	
+	@SubscribeEvent
+	public static void serverStarted(ServerStartedEvent event)
+	{
+		//TODO the second time a world loads, and every subsequent time after that, the captchas are different from the first time the world loads. Both before and after the change, server and client side are synced correctly
```
Dweblenod, line 57: Could use help with this bug

src/main/java/com/mraof/minestuck/client/gui/PunchDesignixScreen.java
```diff
@@ -0,0 +1,87 @@
...
+		int yOffset = (this.height / 2) - (guiHeight / 2);
+		this.captchaTextField = new EditBox(this.font, this.width / 2 - 40, yOffset + 25, 80, 20, Component.translatable(ENTER_CAPTCHA_MESSAGE));
+		this.captchaTextField.setCanLoseFocus(false);
+		
+		String storedCaptcha = be.getCaptcha();
+		if(storedCaptcha != null)
+			this.captchaTextField.setValue(storedCaptcha);
+		
+		this.captchaTextField.setMaxLength(8);
+		this.captchaTextField.setFilter((text) -> Pattern.matches("^[0-9a-zA-Z!#?]+$", text) || text.isEmpty()); //"^[0-9a-zA-Z!#?]+$" indicates that it will only accept characters used to represent hex or the characters "!"/"#"/"?"
```
Dweblenod, line 52: need to update comment to say alphanumeric instead of hex

src/main/java/com/mraof/minestuck/client/gui/PunchDesignixScreen.java
```diff
@@ -0,0 +1,87 @@
...
+	public void render(PoseStack poseStack, int mouseX, int mouseY, float partialTicks)
+	{
+		this.renderBackground(poseStack);
+		
+		int yOffset = (this.height / 2) - (guiHeight / 2);
+		
+		RenderSystem.setShader(GameRenderer::getPositionTexShader);
+		RenderSystem.setShaderColor(1, 1, 1, 1);
+		RenderSystem.setShaderTexture(0, guiBackground);
+		this.blit(poseStack, (this.width / 2) - (guiWidth / 2), yOffset, 0, 0, guiWidth, guiHeight);
```
Dweblenod, line 71: `(this.width / 2) - (guiWidth / 2)` should be a variable next to yOffset or yOffset should be inline with this, for consistency

src/main/java/com/mraof/minestuck/network/data/ShareCaptchasPacket.java
```diff
@@ -0,0 +1,51 @@
...
+	{
+		this.captchas = captchas;
+	}
+	
+	@Override
+	public void encode(FriendlyByteBuf buffer)
+	{
+		for(Map.Entry<String, String> captchaEntry : captchas.entrySet())
+		{
+			buffer.writeUtf(captchaEntry.getKey(), 256); //TODO confirm this number is accurate
```
Dweblenod, line 24: If anyone is able to confirm this, I cant remember my source

src/main/java/com/mraof/minestuck/network/PunchDesignixPacket.java
```diff
@@ -0,0 +1,46 @@
...
+		BlockPos pos = buffer.readBlockPos();
+		String captcha = buffer.readUtf(8);
+		
+		return new PunchDesignixPacket(pos, captcha);
+	}
+	
+	@Override
+	public void execute(ServerPlayer player)
+	{
+		if(player.getCommandSenderWorld().isAreaLoaded(pos, 0))
```
kirderf1, line 37: Should check that the player is close enough too.

src/main/java/com/mraof/minestuck/client/gui/PunchDesignixScreen.java
kirderf1: I suggest adding `xOffset` then.

kirderf1 (changes requested):

src/main/java/com/mraof/minestuck/alchemy/CardCaptchas.java
Dweblenod: Am I overthinking it however?

src/main/java/com/mraof/minestuck/alchemy/CardCaptchas.java
Dweblenod: Have made a separate class for client side, let me know if it actually resolves the issue

src/main/java/com/mraof/minestuck/alchemy/CardCaptchas.java
```diff
@@ -0,0 +1,274 @@
...
+			return CardCaptchas.getCaptcha(registryName);
+		else
+			return null;
+	}
+	
+	public static String getRegistryNameFromItem(Item item)
+	{
+		Optional<ResourceKey<Item>> resourceKey = item.builtInRegistryHolder().unwrapKey();
+		
+		return resourceKey.map(itemResourceKey -> itemResourceKey.location().toString()).orElse(null);
```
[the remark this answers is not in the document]
Dweblenod: Switched to .orElseThrow()

src/main/java/com/mraof/minestuck/alchemy/CardCaptchas.java
```diff
@@ -0,0 +1,274 @@
...
+ */
+@Mod.EventBusSubscriber(modid = Minestuck.MOD_ID, bus = Mod.EventBusSubscriber.Bus.FORGE)
+public class CardCaptchas
+{
+	//TODO use of the character "#" not canonically accurate and suggests there are more punch holes in a captchalogue card then there actually are
+	
+	public static final String AVAILABLE_CHARACTERS = "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789!?";
+	public static final String EMPTY_CARD_CAPTCHA = "00000000";
+	
+	private static final BiMap<String, String> REGISTRY_MAP = HashBiMap.create();
```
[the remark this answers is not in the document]
Dweblenod: Using the registry name (as a string) just seemed to make sense. Eventually it also seemed to be easier to transfer when it came to packets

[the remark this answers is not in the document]
kirderf1: For the packet part, I'd say that it is as easy as if it was resource locations or items (using `buffer.writeResourceLocation()`/`buffer.readResourceLocation()` and `buffer.writeRegistryId()`/`buffer.readRegistryId()` respectively). There is no harm in using items since this all happens after item codes have been set up. And I'd say that it is better, as it is more narrow by disallowing non-valid item registry names. Plus, after codes have been set up, I assume that the map is only used to find a code from an item object or vice versa.
(Oops, a bit longer than intended)

src/main/java/com/mraof/minestuck/alchemy/CardCaptchas.java
kirderf1, line 1: Looks like there might be more public functions than are actually there to be used outside this function. I suggest privating those that are only used for generating codes and making sure that those meant to be used elsewhere (item and code getter) are above other functions in the class.

src/main/java/com/mraof/minestuck/alchemy/CardCaptchas.java
kirderf1: Dunno, but not not by much if so. I'd say I share your feeling on this instance creation. Doing it inside `iterateThroughRegistry()` makes more sense to me since it's first there that it gets used.

kirderf1 (changes requested):

src/main/java/com/mraof/minestuck/data/tag/MinestuckItemTagsProvider.java
```diff
@@ -138,6 +138,7 @@ protected void addTags()
 		tag(BUGS).add(BUG_ON_A_STICK.get(), CHOCOLATE_BEETLE.get(), CONE_OF_FLIES.get(), GRASSHOPPER.get(), CICADA.get(), JAR_OF_BUGS.get());
 		tag(CONSORT_SNACKS).add(Items.COOKIE).addTag(BUGS);
 		tag(CREATIVE_SHOCK_RIGHT_CLICK_LIMIT).add(Items.CHORUS_FRUIT);
+		tag(UNREADABLE).add(CRUEL_FATE_CRUCIBLE.get(), ROYAL_DERINGER.get(), TRANSPORTALIZER.get(), TRANS_PORTALIZER.get(), FEAR_NO_ANVIL.get(), TYPHONIC_TRIVIALIZER.get(), QUILL_OF_ECHIDNA.get(), UMBRAL_INFILTRATOR.get(), FLUORITE_OCTET.get(), CLIENT_DISK.get(), SERVER_DISK.get(), GUTTER_THUMB_DRIVE.get(), GUTTER_BALL.get(), CAPTCHA_CARD.get(), CUEBALL.get(), BLACK_QUEENS_RING.get(), WHITE_QUEENS_RING.get(), BLACK_KINGS_SCEPTER.get(), WHITE_KINGS_SCEPTER.get()).add(Items.DRAGON_EGG).add(Items.DRAGON_HEAD).add(Items.DRAGON_BREATH).add(Items.NETHER_STAR).add(Items.COMMAND_BLOCK).add(Items.COMMAND_BLOCK_MINECART).add(Items.CHAIN_COMMAND_BLOCK).add(Items.REPEATING_COMMAND_BLOCK).add(Items.END_CRYSTAL).add(Items.EXPERIENCE_BOTTLE).add(Items.ELYTRA).add(Items.TOTEM_OF_UNDYING).add(Items.BEACON).add(Items.BEDROCK).add(Items.PUMPKIN);
```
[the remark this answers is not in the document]
Dweblenod: I feel it helps extend the usefulness of raw cruxite. But I would be open to reverting it for the time being

src/main/java/com/mraof/minestuck/item/CaptchaCardItem.java
```diff
@@ -69,10 +71,33 @@ public void appendHoverText(ItemStack stack, @Nullable Level level, List<Compone
...
+		{
+			tooltip.add(Component.literal(CardCaptchas.EMPTY_CARD_CAPTCHA));
 			tooltip.add(makeTooltipInfo(Component.translatable(getDescriptionId() + ".empty")));
+		}
+	}
+	
+	//TODO consider obfuscated characters for unreadable unpunched card
+	private String getCaptcha(ItemStack decodedStack, ItemStack content)
+	{
+		if(AlchemyHelper.isReadableCard(decodedStack))
```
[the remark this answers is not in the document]
Dweblenod: Following discovery that the getShareTag() only worked in multiplayer, the new approach was to add the captcha code to the nbt of the captcha card upon creation, with the full map of captchas only being created and accessible server-side

src/main/java/com/mraof/minestuck/alchemy/CardCaptchas.java
```diff
@@ -0,0 +1,231 @@
...
+	public static Item getItemFromCaptcha(String captcha)
+	{
+		return REGISTRY_MAP.inverse().get(captcha);
+	}
+	
+	@SubscribeEvent
+	public static void serverStarted(ServerStartedEvent event)
+	{
+		//TODO the second time a world loads, and every subsequent time after that, the captchas are different from the first time the world loads. Both before and after the change, server and client side are synced correctly
+		// Is this still the case, or has a recent change fixed it?
```
kirderf1, line 57: This should be checked up on.

src/main/java/com/mraof/minestuck/blockentity/machine/IntellibeamLaserstationBlockEntity.java
```diff
@@ -0,0 +1,238 @@
...
+			if(level != null)
+			{
+				updateState();
+			}
+		}
+	}
+	
+	public void applyReadableNBT(ItemStack taggedCard)
+	{
+		taggedCard.getOrCreateTag().putBoolean("decoded", true);
```
kirderf1, line 162: I suppose we don't need "decoded" anymore, and could instead check whether the stack has the "captcha_code" tag or now?

kirderf1 (commented):
There's also serialization of item codes that would be good for added stability, as well as lazily computing codes.

src/main/java/com/mraof/minestuck/alchemy/CardCaptchas.java
Dweblenod: It no longer appears to be the case, will remove!

src/main/java/com/mraof/minestuck/alchemy/AlchemyHelper.java
```diff
@@ -85,6 +86,15 @@ else if(tag.contains("contentSize") && tag.getInt("contentSize") >= 1)
 		
 	}
 	
+	public static boolean isReadableCard(ItemStack card)
+	{
+		if(!card.is(MSItems.CAPTCHA_CARD.get()) || !hasDecodedItem(card))
+			return false;
+		
+		//either has an existing captcha code or isnt in UNREADABLE item tag to begin with
+		return (card.hasTag() && !card.getTag().getString("captcha_code").isBlank()) || !getDecodedItem(card).is(MSTags.Items.UNREADABLE);
```
kirderf1, line 95:
```suggestion
		return (card.hasTag() && card.getTag().contains("captcha_code", Tag.TAG_STRING)) || !getDecodedItem(card).is(MSTags.Items.UNREADABLE);
```

PR #491, opened 2023-07-09
added datagen for item and block models, and blockstates.

GideonHead:


src/main/java/com/mraof/minestuck/data/MinestuckData.java
```diff
@@ -63,6 +63,9 @@ public static void gatherData(GatherDataEvent event)
 		
 		gen.addProvider(event.includeServer(), new StartingModusProvider(gen, Minestuck.MOD_ID));
 		
+		gen.addProvider(true, new MinestuckBlockStateProvider(gen, fileHelper));
+		gen.addProvider(true, new MinestuckItemModelProvider(gen, fileHelper));
```
kirderf1, line 67:
```suggestion
		gen.addProvider(event.includeClient(), new MinestuckBlockStateProvider(gen, fileHelper));
		gen.addProvider(event.includeClient(), new MinestuckItemModelProvider(gen, fileHelper));
```

kirderf1 (commented):
Should be a good base for future use. Though the two new classes need a little bit or reformatting to use the same style as the rest of the project.

GideonHead:
Done, what other changes need to be made?

GideonHead:
I did some reformatting, lmk if any more needs to be done

kirderf1:
That was reformatting in the wrong direction as the brace style is now specifically what we're not using (we have braces on new lines). Make sure that you're using the code style configuration included in the project when you reformat.

GideonHead:
ah alright, I moved the braces to the new lines

GideonHead:
alright all of the formatting issues are fixed with intellij auto-reformat, so I think this is finished

kirderf1 (approved):
Should be good like this. I suppose you could also convert more blocks and items to use it, but I also think it's fine to leave it for the future on a on-need-basis.

PR #494, opened 2023-07-24
Vinneli's transportalizer textures

InfectedGriffon:
why was the trans flag ever vertical in the first place


Dweblenod:
Edited it so that it will merge into 1.20.1 since thats what we are moving towards now

Dweblenod:
Slap some updates to the CHANGELOG in here and it will be good to go!

PR #495, opened 2023-07-27
Bookshelves and Ladders retextures.

SurrealDude:
Retextured some bookshelves and ladders to be on par with recent changes to some wood types (mostly since Rainbow Bookshelf looks really out of place against any other Rainbow block). These changes apply to land wood types and some of the (unused) aspect wood types


Dweblenod:
Sorry for taking so long to look at this, I will start reviewing now!

Dweblenod:
How do you want to be credited?

PR #499, opened 2023-07-29
Armor Piercing damage type and magic damage comments

Dweblenod:
Adds new damage source for use in armorBypassingDamageMod() OnHitEffect. Removes TODO comments regarding Minestuck magical attacks, but it can be discussed below

### DIFF
--- a/src/main/generated/resources/assets/minestuck/lang/en_us.json
+++ b/src/main/generated/resources/assets/minestuck/lang/en_us.json
@@ -983,8 +983,6 @@
   "container.minestuck.mini_totem_lathe": "Miniature Totem Lathe",
   "container.minestuck.sendificator": "Sendificator",
   "container.minestuck.uranium_cooker": "Cookalyzer",
-  "death.attack.minestuck.armor_pierce": "%s suffered a direct strike to their body",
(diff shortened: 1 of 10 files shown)


Dweblenod:
Regarding the magic damage, I can see indirectMagic() as an alternative to magic(). Not sure how others feels about the distinction

kirderf1:
Yeah, `indirectMagic()` seems like a good choice.

Dweblenod:
> Yeah, `indirectMagic()` seems like a good choice.

Looking at it closer, indirectMagic() requires both a projectile entity and the projectile entities owner. The projectile entity cannot be null

kirderf1:
You could use the same entity for both. From `Guardian.GuardianAttackGoal.tick()`:
```java
livingentity.hurt(this.guardian.damageSources().indirectMagic(this.guardian, this.guardian), f);
```

Dweblenod:
It hath been done

kirderf1:
Would it make sense to rework the armor pierce damage type to have an entity source?

Dweblenod:
> Would it make sense to rework the armor pierce damage type to have an entity source?

I would be down, but am not currently sure how to!

kirderf1:
Alright did it for you and switched it to use player damage death messages.

kirderf1 (approved):

PR #501, opened 2023-08-01
Minor change to player data handling

kirderf1:
Removes some complexity by removing the need to call `PlayerSavedData.setDirty()` when some data changes, leading to removal or replacement of  `PlayerSavedData` fields and parameters.

Normally, instances of `SavedData` use `setDirty()` to signal that some data has changed, and thus only saving when needed rather than every time that the world saves data. However just like with skaianet data, I believe that the complexity of our data makes keeping it and maintaining it not worth it for the performance gain that it might bring. So as with skaianet data, it'll override `isDirty()` to always return true so that it always saves its data.

### DIFF
--- a/src/main/java/com/mraof/minestuck/inventory/captchalogue/CaptchaDeckHandler.java
+++ b/src/main/java/com/mraof/minestuck/inventory/captchalogue/CaptchaDeckHandler.java
@@ -26,6 +26,7 @@ import net.minecraft.world.level.GameRules;
 import net.minecraftforge.event.entity.living.LivingDropsEvent;
 import net.minecraftforge.eventbus.api.EventPriority;
 import net.minecraftforge.eventbus.api.SubscribeEvent;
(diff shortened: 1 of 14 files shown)


Dweblenod:
Is there a way you can recommend to test this?

kirderf1:
One type of test that could be done is to change `isDirty()` to return false or to simply not be overriding, check that player data changes don't get saved properly. Then revert the change and see if it now saves corectly.

Dweblenod:
Grist Gutter does not appear to be saving in either single player or in a server. Other than that, things seem to be working

kirderf1:
The grist gutter is stored in skaianet data, and this PR shouldn't have touched that. So that might've been an issue even before this PR.

Dweblenod (approved):
In light of learning the Grist Gutter is not tied to PlayerSavedData, I think this is good to go!

PR #502, opened 2023-08-05
Grist cleanup and api

kirderf1:
Cleans up and separates grist types, grist sets, grist costs and combination recipes into a new api package.


InfectedGriffon (approved):
literally just some tiny nitpicks, this whole api seems really good so far!

src/main/java/com/mraof/minestuck/api/alchemy/recipe/ContainerGristCostBuilder.java
```diff
@@ -86,11 +98,21 @@ public void build(Consumer<FinishedRecipe> recipeSaver, ResourceLocation id)
 		recipeSaver.accept(new Result(new ResourceLocation(id.getNamespace(), "grist_costs/"+id.getPath()), ingredient, new DefaultImmutableGristSet(costBuilder), priority));
```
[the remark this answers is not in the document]
kirderf1: Yeah, I think it's the appropriate to do. I just wasn't very thorough with it. I'll go through the grist cost builders and take care of that.

src/main/java/com/mraof/minestuck/alchemy/GristHelper.java
```diff
@@ -57,7 +61,7 @@ public static MutableGristSet generateUnderlingGristDrops(UnderlingEntity entity
 		GristType primary = entity.getGristType();
 		GristType secondary = getSecondaryGrist(random, primary);
 		
-		MutableGristSet set = new MutableGristSet();
+		MutableGristSet set = MutableGristSet.newDefault();
```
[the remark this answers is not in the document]
kirderf1: Yeah that's valid too. But I feel that doing it this way matches the intent better:
`MutableGristSet set = new DefaultMutableGristSet();` we want a mutable grist set, and it should specifically be `DefaultMutableGristSet`.
`MutableGristSet set = MutableGristSet.newDefault();` we want a mutable grist set that behaves normally, and that happens to be `DefaultMutableGristSet`.

src/main/java/com/mraof/minestuck/api/alchemy/GristSet.java
```diff
@@ -0,0 +1,136 @@
...
+	
+	static ImmutableGristSet of(Supplier<GristType> type1, long amount1, Supplier<GristType> type2, long amount2, Supplier<GristType> type3, long amount3)
+	{
+		return new DefaultImmutableGristSet(ImmutableMap.<GristType, Long>builder().put(type1.get(), amount1).put(type2.get(), amount2).put(type3.get(), amount3));
+	}
+	
+	static ImmutableGristSet of(Supplier<GristType> type1, long amount1, Supplier<GristType> type2, long amount2, Supplier<GristType> type3, long amount3, Supplier<GristType> type4, long amount4)
+	{
+		return new DefaultImmutableGristSet(ImmutableMap.<GristType, Long>builder().put(type1.get(), amount1).put(type2.get(), amount2).put(type3.get(), amount3).put(type4.get(), amount4));
+	}
```
Dweblenod, line 118: Would it make sense to create an object and pass any number of grist type+amount together using a vararg

src/main/java/com/mraof/minestuck/api/alchemy/GristType.java
```diff
@@ -0,0 +1,205 @@
...
+		{
+			Objects.requireNonNull(item);
+			return candyStack(() -> new ItemStack(item.get()));
+		}
+		
+		public Properties candyStack(Supplier<ItemStack> stack)
+		{
+			candyItem = Objects.requireNonNull(stack);
+			return this;
+		}
```
Dweblenod, line 161: Is there a technical reason or advantage for requiring a candy to exist for a grist type? I understand if it makes it harder in scenarios like the candy weapons

src/main/java/com/mraof/minestuck/api/alchemy/GristTypeSpawnCategory.java
```diff
@@ -0,0 +1,33 @@
+package com.mraof.minestuck.api.alchemy;
+
+import com.mraof.minestuck.Minestuck;
+import net.minecraft.resources.ResourceLocation;
+import net.minecraft.tags.TagKey;
+
+import java.util.Objects;
+import java.util.stream.Stream;
+
+public enum GristTypeSpawnCategory    //Which categories can a certain grist type appear under (for spawning underlings)
```
Dweblenod, line 10: May make sense to reference or briefly explain grist layers

src/main/java/com/mraof/minestuck/api/alchemy/GristType.java
```diff
@@ -0,0 +1,205 @@
...
+		public Properties underlingType(int color)
+		{
+			return this.underlingType(1 / this.rarity, color);
+		}
+		
+		public Properties underlingType(float power, int color)
+		{
+			this.underlingData = new UnderlingData(color, power);
+			return this;
+		}
```
Dweblenod, line 172: This seems to replace the method color, but without context or a description it seems kind of confusing

Dweblenod (commented):
Just some minor thoughts. And just like Hadean had to say, this looks nice!

src/main/java/com/mraof/minestuck/api/alchemy/GristSet.java
InfectedGriffon: varargs wouldn't work here because we need both multiple types and amounts

src/main/java/com/mraof/minestuck/api/alchemy/GristType.java
InfectedGriffon: it defaults to `ItemStack.EMPTY` without calling either method I believe

src/main/java/com/mraof/minestuck/api/alchemy/GristSet.java
Dweblenod: Perhaps im missing something, but what I am suggesting is squishing a grist type + amount in an object so that the object can be passed in

InfectedGriffon: I guess `GristAmount`s would fit that role?

src/main/java/com/mraof/minestuck/api/alchemy/GristType.java
kirderf1: Hmm, actually while it seems appropriate to have the supplier return an empty stack for grist types that lack a candy item, the code that uses the candy item doesn't check for that. I should make it check for that.

src/main/java/com/mraof/minestuck/api/alchemy/GristType.java
kirderf1: Would the problem be solved with some javadoc saying that the color parameter is used to color the texture for underlings of that type?

Dweblenod: I think so!

src/main/java/com/mraof/minestuck/api/alchemy/GristSet.java
kirderf1: Perhaps so.
One of the longest usages of this function is:
```java
GristSet.of(GristTypes.BUILD, 1, GristTypes.COBALT, 1, GristTypes.CHALK, 1, GristTypes.AMBER, 1)
```
Using a vararg with grist amounts as is, we would instead write this as:
```java
GristSet.of(new GristAmount(GristTypes.BUILD.get(), 1), new GristAmount(GristTypes.COBALT.get(), 1), new GristAmount(GristTypes.CHALK.get(), 1), new GristAmount(GristTypes.AMBER.get(), 1))
```
By overloading the constructor, we could reduce this to:
```java
GristSet.of(new GristAmount(GristTypes.BUILD, 1), new GristAmount(GristTypes.COBALT, 1), new GristAmount(GristTypes.CHALK, 1), new GristAmount(GristTypes.AMBER, 1))
```
If we define a function `public static GristAmount amount(long)` for grist types, we could instead write:
```java
GristSet.of(GristTypes.BUILD.get().amount(1), GristTypes.COBALT.get().amount(1), GristTypes.CHALK.get().amount(1), GristTypes.AMBER.get().amount(1))
```
Another alternative is to define a builder object:
```java
GristSet.Builder.new().with(GristTypes.BUILD, 1).with(GristTypes.COBALT, 1).with(GristTypes.CHALK, 1).with(GristTypes.AMBER, 1).build()
```

My current thought is to keep the current functions for fewer arguments (because of how compact those functions are), and add a builder for grist sets with a larger number of grist types.
What are your thoughts?

kirderf1: Or actually, as I just sent that, I changed my mind and like the `amount()` functions a bit better because they group together the grist type and its corresponding amount more clearly than the current thing.

InfectedGriffon: I feel like the `amount()` functions would be both easier to implement and more readable than the builder. Plus, they could be used in other places where GristAmounts would be needed.

kirderf1:
Alright, that should be taken care of. I was reminded of an old thought about grist amount naming. The fields in a grist amount are currently `GristType type` and `long amount`. If you access the latter, it might look like `amount.amount()` or `gristAmount.amount()`. This looks a bit weird, which makes me think that we should rename either the record itself or the number field so that they're not using the same word. I have had no idea what the rename would actually be. Feel free to write something down if you've got an idea!

kirderf1:
I'm thinking of leaving the amount rename for another day to get this merged and out of my focus. Are others in agreement?

Dweblenod:
> I'm thinking of leaving the amount rename for another day to get this merged and out of my focus. Are others in agreement?

That seems reasonable to me!

PR #506, opened 2023-08-12
Editmode locations with some code cleanup for editmode and computer programs

Dweblenod:
Changes the way the region an editmode player can move through is handled. Now an editmode player can move around anything that can "run" the sburb client disk, extending the potential range beyond the Entry island

Each location is composed of the center block position, the relevant dimension, and the source of the location. The source is either
- Entry: Location created upon Entering
- Block: Location created by computer blocks when it possesses the sburb client disk
- Entity: Unused location that could in the future be used for mobile computers that use the sburb client disk

Editmode assets made by glubtier


Dweblenod:
Due to the complex structure of the data stored, I think I messed up somewhere in MSExtraData. I mean its a map storing a multimap storing a pair of BlockPos and enum. The issues with saving/loading were there before the Pair was introduced and I took several attempts at diagnosing and fixing

Dweblenod:
Would it make sense to continue using MSExtraData or would it make more sense to move to something like PlayerData?

src/main/java/com/mraof/minestuck/world/storage/MSExtraData.java
```diff
@@ -181,6 +211,43 @@ public void recoverConnections(Consumer<EditData.ConnectionRecovery> recover)
 		}
 	}
 	
+	/**
+	 * Gets the places the editmode server player can be for their client
+	 */
+	public EditmodeLocations getEditmodeLocations(UUID playerID)
```
kirderf1, line 217: Is the player id meant to be for the client player? Should the editmode locations instance be tied to a specific sburb connection? The choice of player id used doesn't seem to be consistent at the moment. This is pretty important for reviewing the rest of the PR.

Also, it looks like this can return null, and should be annotated as such.

Dweblenod: I wasnt sure if it was smart to have it related to a connection with the idea that the locations shouldnt need to be remade if a connection is changed or broken

src/main/java/com/mraof/minestuck/block/machine/ComputerBlock.java
```diff
@@ -32,6 +35,7 @@
 import net.minecraft.world.phys.shapes.VoxelShape;
 
 import javax.annotation.Nullable;
+import javax.swing.text.html.HTMLDocument;
```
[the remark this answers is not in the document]
Dweblenod: Whoops! I was doing a trial run of another branch, guess the import slipped past me

src/main/java/com/mraof/minestuck/world/storage/MSExtraData.java
kirderf1: Sounds like you might have an example scenario in mind?
Also if you could answer the player id question too that'd be nice.

Dweblenod: Not sure what I was necessarily thinking, but I decided after all to store the locations inside SburbConnection, which now works successfully

src/main/java/com/mraof/minestuck/network/EditmodeTeleportPacket.java
```diff
@@ -0,0 +1,47 @@
...
+			{
+				BlockPos teleportPos = locations.getClosestPosInDimension(player);
+				
+				if(teleportPos != null)
+				{
+					player.teleportTo(teleportPos.getX() + 0.5D, teleportPos.getY() + 1.0D, teleportPos.getZ() + 0.5D);
+				}
+			}
+		}
+	}
```
kirderf1, line 46: Apply early returns?

src/main/java/com/mraof/minestuck/computer/editmode/EditmodeLocations.java
```diff
@@ -0,0 +1,371 @@
...
+	}
+	
+	/**
+	 * Takes in a player and looks through each source (such as a computer) providing editmode to see if the player resides within one of the spaces.
+	 * Used both server side and client side
+	 *
+	 * @param editPlayer Player in editmode
+	 * @return Whether the player is standing in a supported region
+	 */
+	public boolean canMoveAtPosition(Player editPlayer, double defaultRange)
```
kirderf1, line 104: This description might be misleading as the function doesn't just check if the player can move in their position, but might also change their position or change their velocity.

src/main/java/com/mraof/minestuck/computer/editmode/EditmodeLocations.java
```diff
@@ -0,0 +1,371 @@
...
+			}
+			
+			if(localXInBounds && localYInBounds && localZInBounds)
+			{
+				allMatch = true;
+				break;
+			}
+		}
+		
+		return handleMovement(editPlayer, xInBounds, yInBounds, zInBounds, allMatch, noneMatch);
```
kirderf1, line 165: If I understand this loop correctly, you can replace `noneMatch` here with `!xInBounds && !yInBounds && !zInBounds`.

kirderf1, line 165: From my understanding, you don't handle movement if all match. My refactor instincts is telling me that you could move out that return from `handleMovement()` out here

src/main/java/com/mraof/minestuck/computer/editmode/EditmodeLocations.java
```diff
@@ -0,0 +1,371 @@
...
+		
+		return false;
+	}
+	
+	/**
+	 * Takes all the flags from canMoveAtPosition() and modifies the players movement/position as needed.
+	 *
+	 * @return Whether the player is out of one or more coordinate bounds
+	 */
+	private boolean handleMovement(Player editPlayer, boolean xInBounds, boolean yInBounds, boolean zInBounds, boolean allMatch, boolean noneMatch)
```
kirderf1, line 322: One thing that would be a nice safeguard against getting stuck is to instead of setting velocity to 0 along certain axis, we would instead use vector math to remove velocity in the direction away from the closest source.
We could look into it after taking care of some refactor and code cleanup first.
I should have no trouble figuring out how to do it, so feel free to make me do it if you're unsure yourself.

kirderf1 (changes requested):

src/main/java/com/mraof/minestuck/client/gui/playerStats/InventoryEditmodeScreen.java
```diff
@@ -46,11 +48,19 @@ protected void renderBg(GuiGraphics guiGraphics, float par1, int xcor, int ycor)
...
 		
-		guiGraphics.blit(GUI_BACKGROUND, xOffset+ LEFT_ARROW_X, yOffset+ ARROW_Y, guiWidth + (b2?36:0), (less?0:18) + (b1?36:0), 18, 18);
-		guiGraphics.blit(GUI_BACKGROUND, xOffset+ RIGHT_ARROW_X, yOffset+ ARROW_Y, guiWidth+18 + (b2?36:0), (more?0:18) + (b1?36:0), 18, 18);
+		guiGraphics.blit(GUI_BACKGROUND, xOffset + LEFT_ARROW_X, yOffset + ARROW_Y, guiWidth + (b2 ? 36 : 0), (less ? 0 : 18) + (b1 ? 36 : 0), 18, 18);
+		guiGraphics.blit(GUI_BACKGROUND, xOffset + RIGHT_ARROW_X, yOffset + ARROW_Y, guiWidth + 18 + (b2 ? 36 : 0), (more ? 0 : 18) + (b1 ? 36 : 0), 18, 18);
 		
 		drawActiveTabAndIcons(guiGraphics);
 		
+		//TODO add more options (potentially work with icons)
+		addRenderableWidget(new ExtendedButton(xOffset + LEFT_ARROW_X + 31, yOffset + ARROW_Y + 31, 100, 16, Component.literal("TELEPORT CLOSEST"), button -> teleport()));
```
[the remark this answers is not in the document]
Dweblenod: I like the suggestion, but im not sure I feel its more intuitive! I will agree that what I currently have put could use work. And yeah it needs a language key.

An alternative thought I have had would be to create a secondary menu of sorts where all editmode locations are listed. It could also have toggles for theoretical content such as enabling no-clip, disabling selection based block placing/breaking, moving individual blocks around without placing/breaking, or copy pasting selections of blocks

src/main/java/com/mraof/minestuck/computer/editmode/EditmodeLocations.java
Dweblenod: That would be appreciated! I wasnt sure how to make movement smoother along the edges

src/main/java/com/mraof/minestuck/block/machine/ComputerBlock.java
```diff
@@ -138,6 +151,12 @@ private boolean insertDisk(ComputerBlockEntity blockEntity, BlockState state, Le
 		{
 			if(!level.isClientSide && blockEntity.installedPrograms.size() < 3)
 			{
+				SburbConnection c = SkaianetHandler.get(level).getClientConnection(blockEntity);
+				if(c != null)
+				{
+					c.removeClientEditmodeLocations(level.dimension(), pos, EditmodeLocations.Source.BLOCK);
+				}
```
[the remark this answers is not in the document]
Dweblenod: Would this make sense?

`public static void onClientClosed(ComputerBlockEntity be)
	{
		if(be.getLevel() != null)
		{
			Level level = be.getLevel();
			SkaianetHandler handler = SkaianetHandler.get(level);
			
			SburbConnection c = handler.getClientConnection(be);
			if(c != null)
			{
				c.removeClientEditmodeLocations(level.dimension(), be.getBlockPos(), EditmodeLocations.Source.BLOCK);
			}
			
			handler.closeClientConnection(be);	//Can safely be done even if this computer isn't in a connection
		}
	}`

src/main/java/com/mraof/minestuck/client/gui/playerStats/InventoryEditmodeScreen.java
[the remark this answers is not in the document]
Dweblenod: I decided to go through with creating a pop up menu for additional editmode functionality. We could perhaps write out a section labelled "editmode sources" and have explanatory text when hovering over it? Open to suggestions!

src/main/java/com/mraof/minestuck/client/gui/EditmodeSettingsScreen.java
```diff
@@ -0,0 +1,254 @@
...
+		return super.mouseClicked(pMouseX, pMouseY, pButton);
+	}
+	
+	@Override
+	public void resize(Minecraft pMinecraft, int pWidth, int pHeight)
+	{
+		super.resize(pMinecraft, pWidth, pHeight);
+		
+		//TODO resizing does not work for button rendering
+		renderButtons();
```
Dweblenod, line 189: Not sure how to fix

kirderf1:
I am pretty much done with all changes that I had in mind, which did extend to some old code that was adjacent.
What I think is left:
- [ ] Testing for any regression I might have caused, or if there are any undiscovered problems left in general. This I prefer to leave to someone else if possible.
- [ ] TODO-handling. There is one in `EditmodeLocations` about possibly sending a message when an editmode player gets teleported due to a removed computer. I did not want to make a decision myself, but it should still either be done or dropped.
- [ ] I should review the PR diff again and see if there's anything else that I think should be changed.
- [ ] Translation keys in editmode settings screen?

CHANGELOG.md
kirderf1, line 1: Missing changelog entries!

kirderf1 (changes requested):
Aside from the changelog, the only thing that stood out to me was how the icon "buttons" were implemented. I think that it would be better to implement them as actual widgets or other object in order to bundle coordinates and dimensions.
However with the current design being a recurring pattern in our gui screens, I am fine with leaving it to be looked at for all gui screens in its own PR.

CHANGELOG.md
Dweblenod: Changelog entries added! Feel free to critique word choice

kirderf1 (approved):

PR #508, opened 2023-08-16
Temple Scanner Changes:

VivianZane:
* Modified function; now it displays a visual indicator instead of sending a message in the user's chatbox.
* The animated texture of the temple scanner has been split into the base and all its lights, for elegance, efficiency, and ease of development.
* Temple Scanner uses durability to detect charge, but does not break upon reaching full damage.


src/main/java/com/mraof/minestuck/item/StructureScannerItem.java
```diff
@@ -21,65 +37,112 @@ public class StructureScannerItem extends Item
...
+					{
+						return GlobalPos.of(level.dimension(), NbtUtils.readBlockPos(stack.getTag().getCompound("TargetLocation")));
+					} else
+					{
+						return null;
+					}
+				}));
+		
+		ItemProperties.register(this, new ResourceLocation(Minestuck.MOD_ID, "powered"),
+				((pStack, pLevel, pEntity, pSeed) -> pStack.hasTag() && pStack.getTag().getBoolean("Powered") ? 1 : 0));
```
Dweblenod, line 68: I dont believe these should exist in the constructor. It causes a crash when used with a dedicated server. From memory, it may need to go into ClientProxy instead

src/main/java/com/mraof/minestuck/item/StructureScannerItem.java
```diff
@@ -21,65 +37,112 @@ public class StructureScannerItem extends Item
 {
 	private final TagKey<Structure> structure;
 	private final Supplier<Item> fuelItem;
+	private final String scannerName;
 	
-	public StructureScannerItem(Properties properties, TagKey<Structure> structure, Supplier<Item> fuelItem)
+	public StructureScannerItem(Properties properties, TagKey<Structure> structure, Supplier<Item> fuelItem, String scannerName)
```
Dweblenod, line 42: what is the purpose of "scannerName"? If its needed later, it can be acquired via the registry name of the item instance

src/main/java/com/mraof/minestuck/item/StructureScannerItem.java
```diff
@@ -1,14 +1,30 @@
...
+import net.minecraft.server.level.ServerPlayer;
 import net.minecraft.sounds.SoundEvents;
 import net.minecraft.sounds.SoundSource;
 import net.minecraft.tags.TagKey;
 import net.minecraft.world.InteractionHand;
 import net.minecraft.world.InteractionResultHolder;
+import net.minecraft.world.entity.Entity;
+import net.minecraft.world.entity.EntityType;
+import net.minecraft.world.entity.EquipmentSlot;
+import net.minecraft.world.entity.item.ItemEntity;
```
Dweblenod, line 27: A lot of these new imports are unused, make sure to delete the unused stuff!

src/main/java/com/mraof/minestuck/item/StructureScannerItem.java
```diff
@@ -21,65 +37,112 @@ public class StructureScannerItem extends Item
...
+					}
+				}));
+		
+		ItemProperties.register(this, new ResourceLocation(Minestuck.MOD_ID, "powered"),
+				((pStack, pLevel, pEntity, pSeed) -> pStack.hasTag() && pStack.getTag().getBoolean("Powered") ? 1 : 0));
 	}
 	
 	@Override
-	public InteractionResultHolder<ItemStack> use(Level levelIn, Player playerIn, InteractionHand handIn)
+	public InteractionResultHolder<ItemStack> use(Level pLevel, Player pPlayer, InteractionHand pUsedHand)
```
Dweblenod, line 72: why rename the parameters?

src/main/resources/assets/minestuck/models/item/temple_scanner.json
```diff
@@ -1,6 +1,19 @@
...
+  		{ "predicate": { "damage": 0 }, "model": "minestuck:item/temple_scanner_base" },
+  		{ "predicate": { "damage": 0.0001, "angle": 0.0000 }, "model": "minestuck:item/temple_scanner_north" },
+        { "predicate": { "damage": 0.0001, "angle": 0.0625 }, "model": "minestuck:item/temple_scanner_northeast" },
+        { "predicate": { "damage": 0.0001, "angle": 0.1875 }, "model": "minestuck:item/temple_scanner_east" },
+        { "predicate": { "damage": 0.0001, "angle": 0.3125 }, "model": "minestuck:item/temple_scanner_southeast" },
+        { "predicate": { "damage": 0.0001, "angle": 0.4375 }, "model": "minestuck:item/temple_scanner_south" },
+        { "predicate": { "damage": 0.0001, "angle": 0.5625 }, "model": "minestuck:item/temple_scanner_southwest" },
+        { "predicate": { "damage": 0.0001, "angle": 0.6875 }, "model": "minestuck:item/temple_scanner_west" },
+        { "predicate": { "damage": 0.0001, "angle": 0.8125 }, "model": "minestuck:item/temple_scanner_northwest" },
+        { "predicate": { "damage": 0.0001, "angle": 0.9375 }, "model": "minestuck:item/temple_scanner_north" },
```
Dweblenod, line 16: This may just be how it needs to be, but is there a reason 0.0001 was picked for the damage?

Dweblenod (commented):
Very nice!

There are a few things to fix up but good work!

I would recommend adding a message to the screen to indicate what is occurring when the item is active. All it takes is a single click to identify the closest temple and after that point it can be shut off.

I would suggest making it so that creative mode players do not consume uranium or durability

src/main/resources/assets/minestuck/models/item/temple_scanner.json
VivianZane: Yeah; if damage is 0, we'd want the scanner to show the "turned off" sprite. In order to do this with predicates, the best I can do is to set a value close to zero as the first value to get checked.

src/main/java/com/mraof/minestuck/item/StructureScannerItem.java
```diff
@@ -21,65 +37,112 @@ public class StructureScannerItem extends Item
...
 		}
+	}
+	
+	public boolean CheckFuelItem(Player pPlayer, Level pLevel)
+	{
+		ItemStack fuelStack = new ItemStack(fuelItem.get());
+		
 		
-		if(levelIn instanceof ServerLevel level)
+		if(fuelItem != null)
```
kirderf1, line 130: If `fuelItem` is allowed to be null, it should be annotated with `@Nullable`. And if it is, you shouldn't use it to create `fuelStack` before checking if `fuelItem` is null. If it isn't allowed to be null, then this check is unneeded.

src/main/java/com/mraof/minestuck/item/StructureScannerItem.java
```diff
@@ -21,65 +37,112 @@ public class StructureScannerItem extends Item
...
+					pStack.getTag().putBoolean("Powered", false);
 				}
 			}
-		} else
-		{
-			foundItem = true;
 		}
+	}
+	
+	public boolean CheckFuelItem(Player pPlayer, Level pLevel)
```
kirderf1, line 125: The standard for java function names is to use camelCase.

kirderf1 (changes requested):
Don't forget to add a changelog entry.

src/main/java/com/mraof/minestuck/item/StructureScannerItem.java
VivianZane: Because it overrides the use() function in the Item class. I changed the parameter names to the names used in the original function.

src/main/java/com/mraof/minestuck/item/StructureScannerItem.java
VivianZane: '~' ) I... don't know, honestly. I forget why that's there.

Deleted scannerName.

src/main/java/com/mraof/minestuck/item/StructureScannerItem.java
VivianZane: Alright, annotated with `@Nullable` for both the class item and the constructor's parameter. Also moved the fuelStack initialization inside of the check.

kirderf1:
@VivianZane I'm not seeing any pushed changes for the things you've said has been fixed

src/main/java/com/mraof/minestuck/item/StructureScannerItem.java
kirderf1: Correct. Move these to the other `ItemProperties.register()` calls inside `ClientProxy.init()`

src/main/java/com/mraof/minestuck/item/StructureScannerItem.java
kirderf1: Consistency with Parchment parameter names is not as important as just having sensible names. Personally, I always avoid copying the "p_" or "_In" from from such parameter names, however it's not particularly important and shouldn't block the PR from being merged IMO.

src/main/java/com/mraof/minestuck/item/StructureScannerItem.java
```diff
@@ -1,85 +1,132 @@
...
 	@Override
-	public InteractionResultHolder<ItemStack> use(Level levelIn, Player playerIn, InteractionHand handIn)
+	public InteractionResultHolder<ItemStack> use(Level pLevel, Player pPlayer, InteractionHand pUsedHand)
 	{
-		levelIn.playSound(playerIn, playerIn.getX(), playerIn.getY(), playerIn.getZ(), SoundEvents.UI_BUTTON_CLICK.value(), SoundSource.AMBIENT, 0.8F, 1.3F);
+		ItemStack stack = pPlayer.getItemInHand(pUsedHand);
 		
-		ItemStack item = playerIn.getItemInHand(handIn);
-		boolean foundItem = false;
+		if(pPlayer.isCreative() | stack.getDamageValue() != stack.getMaxDamage() | checkFuelItem(pPlayer, pLevel))
```
kirderf1, line 61: I suspect that you want the conditional OR instead of the logical OR here due to the side effects of `checkFuelItem()`. Is that right, or is the logical OR intentional here?

src/main/java/com/mraof/minestuck/item/StructureScannerItem.java
VivianZane: Alright then. I'll leave it as is, for the time being.

src/main/java/com/mraof/minestuck/item/StructureScannerItem.java
```diff
@@ -1,85 +1,132 @@
...
-		if(fuelItem != null)
+	}
+	
+	public static boolean isCharged(ItemStack stack)
+	{
+		return stack.getDamageValue() < stack.getMaxDamage();
+	}
+	
+	@Override
+	public void inventoryTick(ItemStack pStack, Level pLevel, Entity pEntity, int pSlotId, boolean pIsSelected)
```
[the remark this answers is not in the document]
VivianZane: I've decided to break it into several functions, since there are a lot of moving parts here. It should be much more readable and easier to edit, now.

src/main/java/com/mraof/minestuck/item/StructureScannerItem.java
VivianZane: Ah, is that the difference? I was wondering why it allowed me to use a single symbol. I'll resolve that right away.

src/main/java/com/mraof/minestuck/item/StructureScannerItem.java
VivianZane: Ah, I wasn't aware ClientProxy was a Minestuck addition. I was thinking it was a Minecraft component, and thus uneditable. Well that's handy!

Moved the item properties to the Client Proxy init function.

src/main/java/com/mraof/minestuck/item/StructureScannerItem.java
```diff
@@ -1,85 +1,183 @@
...
 		
-		ItemStack item = playerIn.getItemInHand(handIn);
-		boolean foundItem = false;
+		if(pPlayer.isCreative() || stack.getDamageValue() != stack.getMaxDamage() && checkFuelNeeded(pPlayer, pLevel))
+		{
+			stack.getOrCreateTag().putBoolean("Powered", true);
+			pLevel.playSound(pPlayer, pPlayer.getX(), pPlayer.getY(), pPlayer.getZ(), SoundEvents.UI_BUTTON_CLICK.get(), SoundSource.AMBIENT, 0.8F, 1.3F);
+			
+			MutableComponent message = Component.translatable("message.temple_scanner.on");
+			pPlayer.sendSystemMessage(message.withStyle(ChatFormatting.DARK_GREEN));
```
Dweblenod, line 62: this should be server side only

src/main/java/com/mraof/minestuck/item/StructureScannerItem.java
```diff
@@ -1,85 +1,183 @@
...
+	 * Check that the fuel item is set; if not, don't reduce charge.
+	 * Scanner charge is represented by durability and damage. Damage is dealt every 20 ticks.
+	 * If it runs out of charge, power it off.
+	 *
+	 * @param pStack current item
+	 * @param pEntity player entity
+	 * @param pLevel player's current level
+	 * @return Boolean check for if deactivated, used to reactivate if players still has fuel.
+	 */
+	public boolean reduceCharge(ItemStack pStack, Entity pEntity, Level pLevel)
```
Dweblenod, line 128: do you still plan on using the boolean return? Otherwise this would be suitable to turn into a void

src/main/java/com/mraof/minestuck/item/StructureScannerItem.java
```diff
@@ -1,85 +1,183 @@
...
+	}
+	
+	public void useFuelItem(ItemStack invItem, Player pPlayer, Level pLevel)
+	{
+		if(!pLevel.isClientSide)
+		{
+			invItem.shrink(1);
+		}
+		pLevel.playSound(pPlayer, pPlayer.getX(), pPlayer.getY(), pPlayer.getZ(), SoundEvents.BLAZE_SHOOT, SoundSource.AMBIENT, 0.4F, 2F);
 	}
```
Dweblenod, line 182: does not trigger when in survival, meaning the item cannot be refueled once used

src/main/java/com/mraof/minestuck/item/StructureScannerItem.java
VivianZane: Reverted to void.

CHANGELOG.md
```diff
@@ -8,6 +8,16 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 
 ### Changed
 
+- Temple Scanner now displays direction on the item itself, rather than printing coordinates in the chatbox.
+
+### Contributors for this release
+
+- Vivian Zane, Dweblenod, kirderf1
+
+## [Unreleased]
```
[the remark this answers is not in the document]
VivianZane: Which one? You've highlighted several.

src/main/java/com/mraof/minestuck/item/StructureScannerItem.java
VivianZane: I put in a check along the lines of
`if (!sLevel.isClientSide)`
that seems to do the trick.

src/main/java/com/mraof/minestuck/item/StructureScannerItem.java
```diff
@@ -2,84 +2,180 @@
...
+	{
+		return stack.getDamageValue() < stack.getMaxDamage();
+	}
+	
+	public void resetCharge(ItemStack pStack)
+	{
+		pStack.setDamageValue(0);
+	}
+	
+	public BlockPos setLocation(Entity pEntity, ServerLevel sLevel)
```
kirderf1, line 101: I guess you meant `getLocation()`?

src/main/java/com/mraof/minestuck/item/StructureScannerItem.java
```diff
@@ -2,84 +2,180 @@
...
+	{
+		return sLevel.findNearestMapStructure(structure, pEntity.blockPosition(), 100, false);
+	}
+	
+	public void checkLocation(ItemStack pStack, BlockPos pos)
```
kirderf1, line 106: I guess you meant `setLocation()`?

src/main/java/com/mraof/minestuck/item/StructureScannerItem.java
```diff
@@ -2,84 +2,180 @@
...
 		{
-			foundItem = true;
+			pStack.getTag().put("TargetLocation", NbtUtils.writeBlockPos(pos));
 		}
-		
-		if(levelIn instanceof ServerLevel level)
+	}
+	
+	/**
+	 * Check that the fuel item is set; if not, don't reduce charge.
```
kirderf1, line 118:
```suggestion
	 * Does nothing if this scanner type has no fuel item that it uses.
```
Clears up ambiguity on what the fuel item is set to.

src/main/java/com/mraof/minestuck/item/StructureScannerItem.java
```diff
@@ -2,84 +2,180 @@
...
-						levelIn.playSound(playerIn, playerIn.getX(), playerIn.getY(), playerIn.getZ(), SoundEvents.BLAZE_SHOOT, SoundSource.AMBIENT, 0.4F, 2F);
-					}
-					
-					break;
-				}
+				MutableComponent message = Component.translatable("message.temple_scanner.on");
+				pPlayer.sendSystemMessage(message.withStyle(ChatFormatting.DARK_GREEN));
+			}
+		}
+		return InteractionResultHolder.success(stack);
```
kirderf1, line 63: Would it be reasonable to return a success result when it was successfully turned on, and failure if fuel was missing?

src/main/java/com/mraof/minestuck/item/StructureScannerItem.java
```diff
@@ -2,84 +2,180 @@
...
+			if (invItem != null)
+			{
+				useFuelItem(invItem, pPlayer, pLevel);
+				return true;
+			}
+		}
+		return false;
+	}
+	
+	public ItemStack hasFuel(Player pPlayer){
```
kirderf1, line 160: Perhaps `findFuel()`?

src/main/java/com/mraof/minestuck/item/StructureScannerItem.java
```diff
@@ -2,84 +2,180 @@
...
+				if (!pLevel.isClientSide()){
+					MutableComponent message = Component.translatable("message.temple_scanner.off");
+					pEntity.sendSystemMessage(message.withStyle(ChatFormatting.DARK_GREEN));
 				}
-			} else
+			}
+		}
+	}
+	
+	public boolean checkFuelNeeded(Player pPlayer, Level pLevel)
```
kirderf1, line 145: This function might need some form of renaming.
Also, if I interpret the code correctly, it currently prevents a scanner item with `fuelItem == null` from being turned on. Does that seem right? If so, that needs fixing if we're gonna maintain that `fuelItem` can be null.

kirderf1 (changes requested):

src/main/java/com/mraof/minestuck/item/StructureScannerItem.java
VivianZane: Resolved; now it returns true if it's null, and only returns false when it isn't null and the player's inventory doesn't have any fuel in it.

Renamed functions to the following:
- hasFuel => findFuel
- checkFuelNeeded => getFuel
- useFuelItem => useFuel

src/main/java/com/mraof/minestuck/client/ClientProxy.java
```diff
@@ -118,7 +121,10 @@ public static void init(final FMLClientSetupEvent event)
...
 		ItemProperties.register(MSItems.STONE_TABLET.get(), new ResourceLocation(Minestuck.MOD_ID, "carved"), (stack, level, holder, seed) -> StoneTabletItem.hasText(stack) ? 1 : 0);
 		ItemProperties.register(MSItems.MUSIC_SWORD.get(), new ResourceLocation(Minestuck.MOD_ID, "has_cassette"), (stack, level, holder, seed) -> MusicPlayerWeapon.hasCassette(stack) ? 1 : 0);
-		ItemProperties.register(MSItems.BOOMBOX_BEATER.get(), new ResourceLocation(Minestuck.MOD_ID, "has_cassette"), (stack, level, holder, seed) -> MusicPlayerWeapon.hasCassette(stack) ? 1 : 0);
+    ItemProperties.register(MSItems.BOOMBOX_BEATER.get(), new ResourceLocation(Minestuck.MOD_ID, "has_cassette"), (stack, level, holder, seed) -> MusicPlayerWeapon.hasCassette(stack) ? 1 : 0);
+		
+		ItemProperties.register(MSItems.TEMPLE_SCANNER.get(), new ResourceLocation("angle"), new CompassItemPropertyFunction((level, stack, entity) -> {if(stack.hasTag() && stack.getTag().contains("TargetLocation")) {return GlobalPos.of(level.dimension(), NbtUtils.readBlockPos(stack.getTag().getCompound("TargetLocation")));} else {return null;}}));
```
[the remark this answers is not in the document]
VivianZane: ^‿^ ) Hohoh, I think that's the first time I ever had to import something that wasn't a standard library! Done and dusted!

src/main/java/com/mraof/minestuck/item/StructureScannerItem.java
```diff
@@ -2,84 +2,187 @@
...
 	
-	public StructureScannerItem(Properties properties, TagKey<Structure> structure, Supplier<Item> fuelItem)
+	public StructureScannerItem(Properties properties, TagKey<Structure> structure, @Nullable Supplier<Item> fuelItem)
 	{
 		super(properties);
 		this.structure = structure;
 		this.fuelItem = fuelItem;
 	}
 	
+	public static GlobalPos setAngleTag(Level level, ItemStack stack)
```
kirderf1, line 45: Rename to `getTargetLocation()` and annotate with `@Nullable`?

src/main/java/com/mraof/minestuck/item/StructureScannerItem.java
kirderf1: I'd only argue against the "checkFuelNeeded => getFuel" rename. I do have an idea though. If you're fine with it, I could implement it myself and commit it directly to the PR.
My idea is to first convert the `if(getFuel(pPlayer, pLevel) || pPlayer.isCreative())` check in `use()` into an early return, and to then try inlining the `getFuel()` function into `use()`.

Are you fine with me doing this?

kirderf1 (changes requested):

kirderf1:
@VivianZane is it fine by you that I commit the requested changes directly to your branch to finish off this PR?

VivianZane:
Oh, yes, you can absolutely do that! Sorry, I meant to respond after I noticed your second reply, but then... er, I forgot.

PR #512, opened 2023-09-21
Land type extensions for worldgen features

kirderf1:
Introduces "Land type extensions" as a type of datapack json which allows for adding worldgen features, carvers and mob spawns to given biome types in lands with a specific land type (based on the json file name).

Information on the format for datapack makers and others should be written to the github wiki for minestuck after this PR is merged.

While it's also possible to add an extension that calls `MobSpawnSettings.addMobCharge()`, I haven't done so due to not knowing if it would be of interest at all ti extension makers.


kirderf1:
The loading issue has been fixed, so now it is a question about how much cleanup, adjustments, documentation etc. is to be done before we decide to merge.

src/main/java/com/mraof/minestuck/world/gen/LandChunkGenerator.java
```diff
@@ -128,4 +132,33 @@ private static boolean hasGatePlacement(ChunkGeneratorStructureState state, Hold
...
+		
+		
+		public LazyBiomeSettings(Function<LandTypeExtensions, LandCustomBiomeSettings> constructor)
+		{
+			this.constructor = constructor;
+		}
+		
+		public LandCustomBiomeSettings get()
+		{
+			return Objects.requireNonNull(this.value);
```
Dweblenod, line 154: Crash can occur here when a Land which previously had no custom mobs is suddenly given some. 
Tracing its usage back to line 107.

`else return customBiomeSettings.get().customMobSpawnsFor(biome).getMobs(category);`

kirderf1: I don't think I am following you. I'm not sure lands are ever "suddenly given custom mobs", with land mob spawns only being set up/loaded at either server start or land creation, with no regard to what it might've been before. I also don't think there being 0 or 1 custom mobs will make a difference anywhere other than in the return of `getMobs()`.
Did you get a crash at that point? If so, the issue may actually be something else.

PR #520, opened 2023-11-07
Checking for fake player where PlayerSavedData.getData() is called

Dweblenod:
Many of the places where PlayerSavedData.getData() was called did not account for a FakePlayer entity, which resulted in server side crashes when the game tries to get information such as echeladder rung/consort reputation/grist cache.

### DIFF
--- a/src/main/java/com/mraof/minestuck/player/PlayerSavedData.java
+++ b/src/main/java/com/mraof/minestuck/player/PlayerSavedData.java
@@ -100,13 +100,17 @@ public final class PlayerSavedData extends SavedData
 	
 	public PlayerData getData(PlayerIdentifier player)
 	{
-		Objects.requireNonNull(player);
-		if (!dataMap.containsKey(player))
+		if(player != null)
 		{
-			PlayerData data = new PlayerData(this, player);
-			dataMap.put(player, data);
-			setDirty();
+			if (!dataMap.containsKey(player))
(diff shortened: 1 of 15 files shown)


kirderf1:
With this issue boiling down to `IdentifierHandler.encode()` potentially returning null, a more thorough but also more extensive solution could be to return an optional player identifier instead. Any code that gets an identifier from a player will then have to handle this explicitly. I think we should do that eventually, but it is quite a big change, meaning that I don't necessarily think it has to happen right now.

src/main/java/com/mraof/minestuck/player/PlayerSavedData.java
```diff
@@ -100,13 +100,17 @@ public static PlayerData getData(PlayerIdentifier player, MinecraftServer server
 	
 	public PlayerData getData(PlayerIdentifier player)
```
kirderf1, line 101: I'd suggest keeping this function as it was before, and instead add a null check and null result in `PlayerSavedData.getData(ServerPlayer)`

kirderf1 (changes requested):

src/main/java/com/mraof/minestuck/player/PlayerSavedData.java
Dweblenod: Gotcha

PR #521, opened 2023-11-08
Credits and license for music tracks

kirderf1:
This is a start to building up both a credits file and a license file, starting off with all music tracks in the mod at this time.

The tracks by Catboss have been kept under the same license as the albums that they're sourced from (that being the "Land of Fans and Music" albums).
For all other tracks (which have been composed specifically for Minestuck), I have asked for and gotten permission from their authors to license them under CC BY-NC-SA 4.0.

### DIFF
--- a/ASSETS_LICENSE.md
+++ b/ASSETS_LICENSE.md
@@ -1,0 +1,2 @@
+
+`dance_stab_dance.ogg` and `emissary_of_dance.ogg` by Mai Yishan (Catboss) are licensed under [CC BY-NC-SA 3.0](https://creativecommons.org/licenses/by-nc-sa/3.0/). These are the original work modified into a different data format.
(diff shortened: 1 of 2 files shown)


kirderf1:
I've lately wanted to change my approach from building up the credits and license file in parallel to instead focus more on the credits file first.

Dweblenod:
Is there anything with this PR that you would like my help with while it is still a draft?

kirderf1:
Not a lot, I think.
I am trying to decide if I'm gonna work more on the credits file in this PR, or if I instead should get this merged, and work on the credits file in a new PR. If you have any opinions on it, that would be nice.
Otherwise, the only thing would be to give an opinion on the format I've used so far.

kirderf1:
Okay, I've decided to continue with the credits file on a new PR, which means that this should be ready to be reviewed and merged.

PR #523, opened 2023-11-11
Cheese lands Blocks

GideonHead:
Cheese Land Blocks, including 14 Blocks, 9 Stairs, 9 Slabs, and 5 walls. Also added and chaneged some datagen methods for walls.


Dweblenod:
Does this have all the blocks you intend for it to have or will there be more?

Dweblenod (commented):
The chhurpi ores are missing the tool type and level of tool needed to harvest them

kirderf1:
I'm assuming that Rose made some of those textures since you added her to the changelog. Did Rose make all or just some of those new textures?

PR #525, opened 2023-11-13
Data generation for simple assets

kirderf1:
So I aim to eventually license the assets folder (with any needed exceptions) under creative commons.
However I think that some asset files are not suitable under creative commons. These include `sounds.json` as well as certain types of model files (those that configure a preexisting model rather than defining a new one, or are otherwise extremely simple).
Also, as I am thoroughly going through all targeted assets for licensing and crediting, it's really nice to have to go through fewer files and to skip these ones in particular.

Solution? Replace those files with code-generated ones, because:
- Those files end up in a different folder, making it easy to not apply the CC license to them.
- Assets that I believe is not suitable for a CC license correspond really well to assets suitable for datageneration.

In this PR I attempt to keep all model files as identical to what they were before as possible, even when there's a more convenient solution by restructuring or renaming. This is done because the sheer scope of this PR would otherwise hide such changes.
Note: the uv values on the ladder models were automatically removed from the model because they matched the default uv values. This should be the only change to the model files alongside them being moved and reformatted.


src/main/java/com/mraof/minestuck/data/MSBlockStateProvider.java
```diff
@@ -0,0 +1,1215 @@
...
+			modelId = powered ? modelId.withSuffix("_powered") : modelId.withSuffix("_unpowered");
+			topTexture = powered ? topTexture.withSuffix("_powered") : topTexture.withSuffix("_unpowered");
+			
+			if(direction.getAxis() == Direction.Axis.Y)
+				return models().cubeColumn(modelId.getPath(),
+						texture("redstone_machine_block"),
+						topTexture);
+			else
+				return models().getExistingFile(modelId);
+		}, BlockStateProperties.POWER);
```
Dweblenod, line 834: On first glance, this and the other machine/redstone blocks appears to be more convoluted than the manual method. Is there a certain advantage you feel this brings to more complex blocks like these?

kirderf1: Yeah, I can see that some of these are problematic in their current state.
Once I'm at the end, I'll do a second sweep to keep, improve or get rid of these more complex model generation blocks. I can poke you to go through what's left once I've done so.

kirderf1:
@Dweblenod I think that I've gone through enough blocks and items for now. There's more that can be done, but I'd like save them for a separate PR and make sure that what I've got here is fine.
Feel free to look through the class file changes and see if anything looks off or unnecessarily complicated.

Dweblenod (approved):
I have no complaints I can find! This is an exciting PR :)

kirderf1:
Alright, great!

PR #526, opened 2023-11-20
Pogo lance

Dweblenod:
New weapon designed and textured by pavizi!

Uses a new OnHitEffect that flings the attacker and target in opposite directions from where the attacker was looking and dismounts the attacker in the process.

Includes a new packet sent to clients to facilitate moving players around when only the server side is accessible.


src/main/java/com/mraof/minestuck/network/ClientMovementPacket.java
```diff
@@ -0,0 +1,60 @@
...
+		double moveY = buffer.readDouble();
+		double moveZ = buffer.readDouble();
+		
+		return new ClientMovementPacket(playerID, moveX, moveY, moveZ);
+	}
+	
+	@Override
+	public void execute()
+	{
+		Entity entity = Minecraft.getInstance().level.getEntity(playerID);
```
kirderf1, line 54: Assuming that the mocement of the packet should only apply to the player that receives it, you could get the player from the minecraft instance instead of from an entity id.

src/main/java/com/mraof/minestuck/item/weapon/OnHitEffect.java
```diff
@@ -334,6 +336,32 @@ static OnHitEffect enemyKnockback(float knockback)
...
+			
+			target.push(targetVec.x, targetVec.y, targetVec.z);
+			
+			//dismount the attacker
+			if(attacker.getVehicle() != null)
+				attacker.dismountTo(attacker.getX(), attacker.getY(), attacker.getZ());
+			
+			attacker.push(attackerVec.x, attackerVec.y, attackerVec.z);
+			
+			if(attacker instanceof Player player)
```
kirderf1, line 357: Might as well cast directly to `ServerPlayer` if you then need it for `sendToPlayer()`

kirderf1 (approved):

PR #527, opened 2023-11-25
Remove some unused models and textures

kirderf1:
As I cleared out files from the regular assets folder during #525, I noticed some that had gone out of use without being removed. Here's me going through and removing unused stuff I happen to spot.


kirderf1:
I'll look back on the other listed models later. But I do think that I have a clear motivation for the aspect log barks.
At some point, log blocks had four possible states: x-axis, y-axis, z-axis and no-axis. The no-axis one, aka log bark blocks, was eventually removed from vanilla logs (and thus also from mod logs) at some point, assumedly with the addition of wood blocks. Those bark log models are simply a remnant from then.

kirderf1:
Both polished coarse stone and vine oak planks refer to textures that don't exist in the mod. When also considering the simplicity of the models, I'd argue that there's no value preserving those two for a moment when they might be added.

kirderf1:
I don't think the jumper block extension item model is worth preserving, but I could see a point in doing so for the item texture. If you want it preserved, I suggest that I instead move it to the "unused textures" folder. Does that seem fine?

PR #531, opened 2023-11-27
Temple scanner polish

kirderf1:
While looking at changing the temple scanner model with its overrides, I noticed some other changes that I wanted to make, which eventually built up to this PR.

I also gave a try at replacing the target position with just an angle that was calculated and updated server-side in time with the power drain, but I did not get a satisfying result, so I left it like this instead.


Dweblenod (approved):
Looks good and worked fine in some simple testing! 

It is nice to have the power item property and not be relying on durability (and its blue!), it is also nice not to have the equip animation keep replaying, also the new status messages are useful

kirderf1:
Alright, then I'll merge it in a moment!

PR #532, opened 2023-11-29
Parent model for some similar blocks

kirderf1:
Extracts some common elements and display data shared between block models to create `cube_up_mirrored.json` and `logic_gate_cube.json`. The block specific models are now a bit more suitable for data generation, the logic gate blocks in particular.

While I wanted to make some more specific models with fewer texture variables, there weren't enough remaining similarities to go by.


kirderf1:
I took a flight around the [debug mode](https://minecraft.wiki/w/Debug_mode) without spotting anything odd, though I wouldn't apot any subtle differences that way. That said, I think that I was pretty thorough in this process, so I'm pretty confident that this is fine.

PR #541, opened 2023-12-21
Data driven computer Themes

Dweblenod:
Changing it so that computer themes are determined by files in a resource pack, allowing players to add their own custom themes (computer wallpaper and text color) to their computer independently from the server. 

Everyone will still has access to the existing officially created computer themes. If one player does not have the additions another player has to their resource pack, then they will see the default wallpaper and text color. Players have the option to choose a Theme from the list that is available based off of their own resource pack.


src/main/java/com/mraof/minestuck/computer/theme/ComputerThemeManager.java
```diff
@@ -0,0 +1,99 @@
...
+		Optional<ResourceLocation> potentialPath = themes.stream().filter(theme ->
+				theme.getThemeName().equals(name)).findAny().map(ComputerTheme::getTexturePath);
+		return potentialPath.orElse(ComputerTheme.DEFAULT_TEXTURE_PATH);
+	}
+	
+	public int findTextColor(String name)
+	{
+		Optional<Integer> potentialColor = themes.stream().filter(theme -> theme.getThemeName().equals(name)).findAny().map(ComputerTheme::getTextColor);
+		return potentialColor.orElse(ComputerTheme.DEFAULT_TEXT_COLOR);
+	}
```
kirderf1, line 86: How about a function that takes a theme name and returns a text component or translation key?

src/main/java/com/mraof/minestuck/computer/theme/ComputerThemeManager.java
```diff
@@ -0,0 +1,99 @@
...
+@Mod.EventBusSubscriber(modid = Minestuck.MOD_ID, bus = Mod.EventBusSubscriber.Bus.MOD, value = Dist.CLIENT)
+public class ComputerThemeManager extends SimpleJsonResourceReloadListener implements PreparableReloadListener
+{
+	private static final Logger LOGGER = LogManager.getLogger();
+	private static final Gson GSON = new GsonBuilder().registerTypeAdapter(ComputerTheme.class, new ComputerTheme.Serializer()).create();
+	
+	public static final String PATH = "minestuck/computer_themes";
+	public static final String PATH_W_SLASH = PATH + "/";
+	
+	private List<ComputerTheme> themes;
```
kirderf1, line 35: Since computer themes are supposed to be uniquely identified by name, I think it makes more sense for this to be a map with the name/id as its key.

src/main/java/com/mraof/minestuck/computer/theme/ComputerThemeManager.java
```diff
@@ -0,0 +1,99 @@
...
+import org.apache.logging.log4j.Logger;
+
+import java.util.*;
+
+/**
+ * Helps acquire json files in assets/minestuck/minestuck/computer_themes/
+ * ComputerThemes are data driven resource pack files that determine a computers wallpaper and text color.
+ */
+@Mod.EventBusSubscriber(modid = Minestuck.MOD_ID, bus = Mod.EventBusSubscriber.Bus.MOD, value = Dist.CLIENT)
+public class ComputerThemeManager extends SimpleJsonResourceReloadListener implements PreparableReloadListener
```
kirderf1, line 27: During #512 I had an idea to separate the reload listener "loader" from the data fields and query functions, to instead make it an inner class (see `LandTypeExtensions.Loader`). I think the same pattern could be good here.

src/main/java/com/mraof/minestuck/computer/theme/ComputerTheme.java
```diff
@@ -0,0 +1,73 @@
...
+ */
+public class ComputerTheme
+{
+	private static final Logger LOGGER = LogUtils.getLogger();
+	
+	public static final int DEFAULT_TEXT_COLOR = 0x404040;
+	public static final ResourceLocation DEFAULT_TEXTURE_PATH = new ResourceLocation(Minestuck.MOD_ID, "textures/gui/theme/default.png");
+	
+	private final ResourceLocation texturePath;
+	private final ConstantInt textColor;
```
kirderf1, line 26: I'm not sure there's much of a point to using `ConstantInt` in favor of a regular `int` unless you're abstracting over `IntProvider`. I suggest simplifying this by replacing it with a regular int.

src/main/java/com/mraof/minestuck/computer/theme/ComputerTheme.java
```diff
@@ -0,0 +1,73 @@
...
+import net.minecraft.util.GsonHelper;
+import net.minecraft.util.valueproviders.ConstantInt;
+import org.slf4j.Logger;
+
+import java.lang.reflect.Type;
+
+/**
+ * This class is used to define json files that will be used to determine what wallpaper and text color appears on in-game computers.
+ */
+public class ComputerTheme
```
kirderf1, line 18: This class is currently almost a record type. I suggest turning it into a record type.

kirderf1, line 18: I'd like to suggest a rename to `ComputerThemeData` since this class strictly deals with the data of a theme.

src/main/java/com/mraof/minestuck/computer/theme/ComputerTheme.java
```diff
@@ -0,0 +1,73 @@
...
+		
+		@Override
+		public JsonElement serialize(ComputerTheme themeData, Type type, JsonSerializationContext context)
+		{
+			JsonObject json = new JsonObject();
+			json.add("texture_location", ResourceLocation.CODEC.encodeStart(JsonOps.INSTANCE, themeData.texturePath).getOrThrow(false, LOGGER::error));
+			json.add("text_color", ConstantInt.CODEC.encodeStart(JsonOps.INSTANCE, themeData.textColor).getOrThrow(false, LOGGER::error));
+			json.add("theme_name", Codec.STRING.encodeStart(JsonOps.INSTANCE, themeData.themeName).getOrThrow(false, LOGGER::error));
+			return json;
+		}
```
kirderf1, line 71: Considering how you're doing the serializing with individual codecs for each part, I believe that it would be nicer to define a codec for computer themes as a whole, and then use that to serialize and deserialize. You'd be using `RecordCodecBuilder` to define the codec, which sohuld have plenty of examples around.

kirderf1 (changes requested):

kirderf1 (approved):

PR #545, opened 2024-01-06
Skaianet and Sburb connection cleanup

kirderf1:
The goal is to separate `SburbConnection` into two groups of data:
- `ActiveConnection`, which has references to the computers used in the connection. Relevant for entering editmode.
- `SburbPlayerData` for any data that is more closely tied to the player/that should not disappear when the computer connection is closed. Includes entry status and primary player connections.

Removes some unused, poorly used or low-value features to simplify related code:
- Locking a sburb connection or predefined data to a specific session. Removed as it wasn't really used at this point.
- Locking a session so that no additional players could join a session. Was meant to be part of the sburb session command (the predecessor to sburb predefine which had a focus on predefining which players belonged to the same session, unlike the simplified predefine that exists today).
- Session name, which was used to identify sessions with commands. Was part of the sburb session command, but ceased being used with the simplified sburb predefine.
- The hard requirement that two identical titles shouldn't exist in the same session. While it does still make sense to avoid duplicate titles, I don't think it is worth the code to make it a hard requirement.
- Session size limit of 144 players. It was added to support the hard requirement of no duplicate titles, so it doesn't make sense to keep when removing said hard requirement.

The code surrounding connections is also becoming less object-oriented and more data-oriented.
Spaghetti code is being untangled, redirects are shortened.
`SkaianetHandler` got split in parts, predefine data was moved out of sessions.
The internal api for connections was updated to be more reliable to be used by code outside the `skaianet` package. Some code that didn't quite belong in the package gets moved closer to their caller.

Backwards compatibility is in place, and has been tested in a very limited capacity.


src/main/java/com/mraof/minestuck/computer/editmode/DeployList.java
```diff
@@ -349,55 +363,79 @@ public static DeployEntry getEntryForName(String name, EntryLists entryList)
...
 		}
 		return nbt;
 	}
 	
+	private static long lastDay;
+	
+	@SubscribeEvent
+	public static void serverStarting(ServerStartingEvent event)
+	{
+		ConsortDialogue.serverStarting();
```
Dweblenod, line 420: The method called here is in what I would have assumed is an irrelevant class and is empty

src/main/java/com/mraof/minestuck/skaianet/InfoTracker.java
```diff
@@ -96,92 +99,100 @@ void requestInfo(ServerPlayer player, PlayerIdentifier p1)
...
+		return new SkaianetInfoPacket.LandChains(createLandChains());
 	}
 	
 	private List<List<ResourceKey<Level>>> createLandChains()
 	{
 		List<List<ResourceKey<Level>>> landChains = new ArrayList<>();
 		
 		Set<ResourceKey<Level>> checked = new HashSet<>();
-		skaianet.sessionHandler.getConnectionStream().forEach(c -> populateLandChain(landChains, checked, c));
+		skaianet.players().forEach(player -> populateLandChain(landChains, checked, player));
```
Dweblenod, line 124: A server side crash can occur here when /debuglands is used, the details are in discord

src/main/java/com/mraof/minestuck/skaianet/SburbPlayerData.java
```diff
@@ -0,0 +1,274 @@
...
+import java.util.HashSet;
+import java.util.Optional;
+import java.util.Random;
+import java.util.Set;
+import java.util.function.Supplier;
+
+/**
+ * Contains various player-specific data that were originally stored in {@code SburbConnection}.
+ * @author kirderf1
+ */
```
Dweblenod, line 35: I think this javadoc does a good job of describing how it relates to the old code but doesnt help much to summarize what kind of player specific data is stored in relation to sburb

src/main/java/com/mraof/minestuck/skaianet/SessionHandler.java
```diff
@@ -40,120 +70,227 @@ public abstract class SessionHandler
...
-		return cClient.isPresent() && sClient == sServer && (MinestuckConfig.SERVER.allowSecondaryConnections.get() || cClient.get() == cServer.orElse(null))	//Reconnect within session
-				|| !cClient.isPresent() && !serverActive && !(sClient != null && sClient.locked) && !(sServer != null && sServer.locked);	//Connect with a new player and potentially create a main connection
+		return SkaianetData.get(server).sessionHandler;
 	}
 	
-	Session getSessionForConnecting(PlayerIdentifier client, PlayerIdentifier server) throws MergeResult.SessionMergeException
+	/**
+	 * An implementation that only allows for a single, global session, which all players are assumed to be part of
+	 */
+	private static final class Global extends SessionHandler
```
Dweblenod, line 119: Since I havent messed around with it before, have you tested the global config with these changes?

src/main/java/com/mraof/minestuck/skaianet/SkaianetTests.java
```diff
@@ -0,0 +1,58 @@
...
+import net.minecraft.gametest.framework.GameTest;
+import net.minecraft.gametest.framework.GameTestHelper;
+import net.minecraftforge.gametest.GameTestHolder;
+import net.minecraftforge.gametest.PrefixGameTestTemplate;
+
+import java.util.Optional;
+
+@PrefixGameTestTemplate(false)
+@GameTestHolder(Minestuck.MOD_ID)
+public final class SkaianetTests
```
Dweblenod, line 15: Proposing a javadoc description of the file be added!

Dweblenod (commented):
First run at reviewing! 

The cleanup is great to see :) Im still trying to parse the degree to which this impacts code but I am interested to see if this makes it simpler to access Session and Connection related data which could be important to future content such as quests

I have mentioned similar concerns in the past (often after noticing crashes due to FakePlayer entities) but I still do not quite "get" why its beneficial to increase how often our mod might throw errors.

CHANGELOG.md
```diff
@@ -6,13 +6,23 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
...
+- Minor changes to what is shown in the data checker
+- Duplicate titles are now allowed in the same session
+- Pre-entry players no longer play a part in grist gutter capacity,
+and they no longer receive grist from the gutter
+
 ### Fixed
 
 - Metal Boats no longer crash the game when dispensed
+- Fixes some niche issues with connections
+- `/sburbpredefine` now gives an error when applied to a player that already has a land
```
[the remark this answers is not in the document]
kirderf1: It's not a runtime exception that might crash the game, but a command exception, a special type of exception that aborts the command and gives a feedback message to the command executor.

Previously for players with a land, the command would predefine data as normal. That was bad because that data would then never actually be applied to the player (because their land and title has already been set), even if the command reported it as a success.

src/main/java/com/mraof/minestuck/skaianet/SburbPlayerData.java
kirderf1: Yeah, that's pretty much all it is. It's not much more than another group of player data alternative to that in `PlayerData`. I just didn't more the data out of `SkaianetData` for better backwards-compatibility, and didn't split up the data further out of convenience.

I'm not sure if anything more can be said in the javadoc right now. I do see this as somewhat temporary though, since I've been thinking of breaking this apart using neoforge data attachments much like what I suggested for `PlayerData` in the discord.

src/main/java/com/mraof/minestuck/skaianet/SessionHandler.java
kirderf1: Tested a little bit just now, but nothing thorough. I might look into making a few game tests for `SessionHandler`.

src/main/java/com/mraof/minestuck/command/DebugLandsCommand.java
```diff
@@ -32,7 +43,71 @@ public static void register(CommandDispatcher<CommandSourceStack> dispatcher)
...
+					break;
+				
+				PlayerIdentifier fakePlayer = IdentifierHandler.createNewFakeIdentifier();
+				
+				connections.setPrimaryConnection(fakePlayer, lastIdentifier);
+				createAndSetLand(land, fakePlayer, player.server);
+				
+				lastIdentifier = fakePlayer;
+			}
+		}
```
Dweblenod, line 97: This is a rather subjective comment but this block seems awkwardly written to me. It is not hard to read, just seems like their may be a more straightforward solution

src/main/java/com/mraof/minestuck/network/computer/SkaianetInfoPacket.java
```diff
@@ -2,146 +2,117 @@
...
-						list.add(null);
-					else list.add(ResourceKey.create(Registries.DIMENSION, new ResourceLocation(landName)));
-				}
-				packet.landChains.add(list);
-			}
-		} else
+			buffer.writeCollection(landChains, (buffer1, list) ->
+					buffer1.writeCollection(list, (buffer2, land) ->
+							buffer2.writeOptional(Optional.ofNullable(land),
+									FriendlyByteBuf::writeResourceKey)));
```
Dweblenod, line 98: While very efficient in terms of the number of lines it takes up, it takes a while to unpack. Extraction to a method that names what is going on or just adding a comment may be helpful

src/main/java/com/mraof/minestuck/network/computer/SkaianetInfoPacket.java
```diff
@@ -2,146 +2,117 @@
...
+		}
+		
+		public static LandChains decode(FriendlyByteBuf buffer)
 		{
-			packet.playerId = buffer.readInt();
+			List<List<ResourceKey<Level>>> landChains = buffer.readList(buffer1 ->
+					buffer1.readList(buffer2 ->
+							buffer2.readOptional(buffer3 ->
+									buffer3.readResourceKey(Registries.DIMENSION)
+							).orElse(null)));
```
Dweblenod, line 107: Same thoughts as my comment in encode()

src/main/java/com/mraof/minestuck/world/GateHandler.java
```diff
@@ -92,44 +90,60 @@ private static GlobalPos findPosNearLandGate(ServerLevel level)
...
+				.flatMap(clientPlayer -> {
+					SburbPlayerData clientPlayerData = SburbPlayerData.get(clientPlayer, level.getServer());
+					return Optional.ofNullable(clientPlayerData.getLandDimensionIfEntered());
+				});
+		
+		if(clientLandOptional.isEmpty())
+		{
+			//player.sendMessage(new TranslationTextComponent(MISSING_LAND));
+			return null;
+		}
```
Dweblenod, line 111: Is the MISSING_LAND message intended to eventually be used? Is there another reason it is being kept around?

src/main/java/com/mraof/minestuck/world/GateHandler.java
```diff
@@ -92,44 +90,60 @@ private static GlobalPos findPosNearLandGate(ServerLevel level)
...
+		Optional<ResourceKey<Level>> serverLandOptional = SburbConnections.get(level.getServer()).primaryPartnerForClient(landPlayer.get())
+				.flatMap(serverPlayer -> {
+					SburbPlayerData serverPlayerData = SburbPlayerData.get(serverPlayer, level.getServer());
+					return Optional.ofNullable(serverPlayerData.getLandDimensionIfEntered());
+				});
+		if(serverLandOptional.isEmpty())
+		{
+			//player.sendMessage(new TranslationTextComponent(MISSING_LAND));
+			return null;
+		}
```
Dweblenod, line 143: same as other message regarding MISSING_LAND

Dweblenod (commented):
I really like all the javadoc on the new files!

CHANGELOG.md
```diff
@@ -6,14 +6,25 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 
 ## [Unreleased]
 
+### Changed
+
+- Minor changes to what is shown in the data checker
+- Duplicate titles are now allowed in the same session
+- Pre-entry players no longer play a part in grist gutter capacity,
+and they no longer receive grist from the gutter
+
```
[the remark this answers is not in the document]
kirderf1: Good point!

src/main/java/com/mraof/minestuck/command/DebugLandsCommand.java
kirderf1: I agree. While I do think it's overall better than it was before, the looping/iteration here can still be made better.

src/main/java/com/mraof/minestuck/world/GateHandler.java
kirderf1: I had left it untouched to avoid deciding anything on it and focus more on other changes.
Looking at it now, it seems that it was originally commented out due to a code refactoring that left this code in a function without a player parameter, and I didn't think it was appropriate to add it there.
I'm not sure how much value the message holds, so I'll go ahead and remove the commented messages and the `MISSING_LAND` key.

src/main/java/com/mraof/minestuck/network/computer/SkaianetInfoPacket.java
kirderf1: True. Lambdas within lambdas like this is not the best. As you say, method extraction might be suitable here.

PR #551, opened 2024-01-20
boat item code cleanup

InfectedGriffon:
While working on the dispenser behaviors, I noticed that metal boats caused a stack overflow when trying to dispense them. This fixes that, cleans up and de-nests the code a bit, and removes the `BoatProvider` interface by using metal boat types directly.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,14 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 
 ## [Unreleased]
 
+### Fixed
+
+- Metal Boats no longer crash the game when dispensed
+
+### Contributors for this release
+
+- hadean
+
 ## [1.20.1-1.11.2.0] - 2024-01-13
 
 ### Added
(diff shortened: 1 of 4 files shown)


kirderf1 (changes requested):
Alright, nice!
With it fixing an issue, you should add that to the changelog.

I'm a bit unsure about the replacement of `BoatProvider` with `MetalBoatEntity.Type`, since it makes the item restricted to boats of the `MetalBoatEntity` kind.
I suppose it doesn't make any difference either way right now since that's the only kind of boat we have right now, so I'll leave it up to you. But I do have some feedback depending on what you choose:
- If you stick to using type inside the item, I think that the item should be renamed to make it clear that it's only used for a specific type of boat. Maybe `MetalBoatItem`?
- If you revert back to the `BoatProvider`, you could still make the constructor more compact if you add a helper function to the type enum that can be used as a method reference to replace
`(stack, world, x, y, z) -> new MetalBoatEntity(world, x, y, z, MetalBoatEntity.Type.IRON)`
with `MetalBoatEntity.Type.IRON::createEntity`.

kirderf1:
Oh right, implementing the interface directly on the type is an option too. I like that choice here over using it as a functional interface. Great thinking!

kirderf1 (approved):

PR #552, opened 2024-01-24
AspectTreeBlocks class

glubtier:
Migrated blocks related to aspect trees to `AspectTreeBlocks` + updated references to these blocks. Also updated methods in `MSBlocks` from `private` to `protected`, for use by `AspectTreeBlocks`. But I know there are other ways this could be done, so let me know if you'd like a different way of implementing.


CHANGELOG.md
```diff
@@ -9,6 +9,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 ### Added
 
 - Spawn egg items for main Underling types, Consorts, and Carapacians
+- All aspect tree-related blocks moved to `AspectTreeBlocks`
```
kirderf1, line 12: I'd leave this out since it's an internal code change and does not matter in any way for end users (AKA players).

glubtier:
> Have you tried it out in-game? I'm guessing that static initialization need to be triggered for the class to work properly.

I did and I didn't see any issues, tried the blocks out and everything worked. But I can change it if you like.

CHANGELOG.md
glubtier: ✅

kirderf1:
Oh right, static initialization happens right now because the block fields are used in `MSItems`. Speaking of, what do you think about moving aspect wood items from `MSItems` to `AspectTreeBlocks` next to their respective blocks?

glubtier:
I don't have a strong opinion, so that sounds fine!

kirderf1:
Alright great!

It would be nice to do something similar for aspect wood datagen as well, but doing that from scratch is not at all as straightforward as the registry object fields.

Also, with the blocks and items next to each other, it seems like a good idea to try and add a helper that combines them. I'm thinking of a helper function that registers both and returns a record with fields for the two registry objects.

Is this something you want to look into? Or leave for someone else to do or for a future PR? I'm fine with merging this with what's done so far, so I'll leave it up to you whether to go further with this PR or not.

glubtier:
> Is this something you want to look into? Or leave for someone else to do or for a future PR? I'm fine with merging this with what's done so far, so I'll leave it up to you whether to go further with this PR or not.

I'd prefer to merge this PR and then either I or someone else (if they're really chomping at the bit) can look into this later! I just want to make sure I have time to actually work on the trees before 4/13. :)

kirderf1 (approved):

PR #555, opened 2024-01-27
Customizable and Lockable Transportalizers

InfectedGriffon:
loot tables and GUIs and packets oh my


src/main/java/com/mraof/minestuck/blockentity/TransportalizerBlockEntity.java
```diff
@@ -212,21 +210,36 @@ private static boolean isDimensionForbidden(Level level)
...
+		return !id.isEmpty();
+	}
+	
 	public String getId()
 	{
 		return id;
 	}
 	
-	public void setId(String id)
+	public void setId(String id, Player player)
```
kirderf1, line 223: For packet security, I recommend returning with no change if the id has already been set instead of replacing it. Or if the transportalizer is locked.

src/main/java/com/mraof/minestuck/blockentity/TransportalizerBlockEntity.java
kirderf1, line 1: Make sure that the destination id can't be changed once the transportalizer is locked.

src/main/java/com/mraof/minestuck/item/block/TransportalizerItem.java
```diff
@@ -22,12 +27,30 @@ public TransportalizerItem(Block blockIn, Properties builder)
...
 	protected boolean updateCustomBlockEntityTag(BlockPos pos, Level level, @Nullable Player player, ItemStack stack, BlockState state)
 	{
-		if(stack.hasCustomHoverName() && stack.getHoverName().getString().length() == 4)
+		BlockEntity be = level.getBlockEntity(pos);
+		if(stack.hasTag() && be instanceof TransportalizerBlockEntity transportalizer)
 		{
-			BlockEntity be = level.getBlockEntity(pos);
-			if(be instanceof TransportalizerBlockEntity transportalizer)
-				transportalizer.setId(stack.getHoverName().getString().toUpperCase());
+			transportalizer.setId(stack.getTag().getString("idString"), player);
```
kirderf1, line 33: This will now not set the id if it is already in use. However I'd argue that it should behave as it did before: set the id regardless of if it is used, and set the transportalizer to not be active if the id is already used.

There would then be two different scenarios/functions:
- Setting the id from player input/packet
- Copying over the id from item nbt

src/main/java/com/mraof/minestuck/network/TransportalizerPacket.java
kirderf1, line 1: Perhaps we should add some check on code lengths in the packet execution?

kirderf1 (changes requested):
Thanks for the work so far!

Do you think we should accumulate this into the 4/13 update or not? @Dweblenod if I can get your opinion as well.

src/main/java/com/mraof/minestuck/network/TransportalizerPacket.java
InfectedGriffon: what do you mean by this?

kirderf1: The gui ensures that inputted codes are of length 4, but the packet is currently not verifying that.
Because packets from clients should generally not be assumed to be valid, we should check that the code is of valid length before using it in the packet.

src/main/java/com/mraof/minestuck/blockentity/TransportalizerBlockEntity.java
```diff
@@ -212,21 +215,58 @@ private static boolean isDimensionForbidden(Level level)
...
+	 */
 	public void setId(String id)
 	{
-		if(level != null && !level.isClientSide)
+		if(level != null && !level.isClientSide && !id.isEmpty())
 		{
+			var data = TransportalizerSavedData.get(level);
 			GlobalPos location = GlobalPos.of(level.dimension(), worldPosition);
-			if(active && !this.id.isEmpty())
-				TransportalizerSavedData.get(level).remove(this.id, location);
```
kirderf1, line 226: Removing this code is kinda bad. I suggest that you either:
- Add it back, so that `setId()` can replace an already-set id without causing issues.
- Add a `this.hasId()` check so that `setId()` will only set the id if an id has not already been set.

kirderf1 (changes requested):

Dweblenod:
> Thanks for the work so far!
> 
> Do you think we should accumulate this into the 4/13 update or not? @Dweblenod if I can get your opinion as well.

I think this makes sense to accumulate for 4/13, yes!

src/main/java/com/mraof/minestuck/blockentity/TransportalizerBlockEntity.java
```diff
@@ -212,22 +215,41 @@ private static boolean isDimensionForbidden(Level level)
...
-		}
+		if(level == null || level.isClientSide ||this.locked || this.hasId() || id.isEmpty())
+			return;
+		
+		this.id = id;
+		GlobalPos location = GlobalPos.of(level.dimension(), worldPosition);
+		BlockState state = level.getBlockState(worldPosition);
+		this.setChanged();
+		level.sendBlockUpdated(worldPosition, state, state, 0);
+		this.active = TransportalizerSavedData.get(level).set(id, location);
```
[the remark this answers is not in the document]
InfectedGriffon: mb, that was an artifact of when it returned it instead of setting it there.

Dweblenod (approved):
Some testing in game does not reveal any bugs. I also have no issues from a gameplay aspect, so this has my approval!

kirderf1 (approved):

PR #556, opened 2024-01-28
Tree stump feature

Dweblenod:
Adds a new tree stump Feature that generates in the normal biome of the Forest Land Type.

Structure created by ScarabOasis with edits by Dweblenod


src/main/java/com/mraof/minestuck/world/gen/feature/tree/TreeStumpFeature.java
```diff
@@ -0,0 +1,66 @@
...
+	}
+	
+	private boolean isEvenTerrain(FeaturePlaceContext<NoneFeatureConfiguration> context)
+	{
+		WorldGenLevel level = context.level();
+		BlockPos pos = context.origin();
+		int minY = Integer.MAX_VALUE;
+		int maxY = Integer.MIN_VALUE;
+		
+		for(BlockPos floorPos : BlockPos.betweenClosed(pos, pos.offset(8, 0, 8)))
```
kirderf1, line 50: > `BlockPos.betweenClosed(pos, pos.offset(8, 0, 8))`

I'm guessing that this is supposed to match up with the space of the placed structure. Do you know if it actually matches up?

Dweblenod: I cant actually confirm with certainty, but I based it off of minWorldHeightInSize. If it isnt taking rotation into effect, it still seems to do the job it was intended to do

kirderf1: Took a closer look at `AbstractTemplateFeature`, and yeah it does seem to match it correctly. But apparently it's not supposed to. The position used for placing the template and which is passed into `pickY()` and `minWorldHeightInSize()` is supposed to be offset so that the template is centered on the origin, but due to a mistake in `AbstractTemplateFeature.place()`, that position actually ended up just being the origin pos, which places the template with a corner on the origin and makes the area used here match up with the template area on accident.

So that's a separate issue to be fixed in `AbstractTemplateFeature`. But more importantly for this PR, I think that code in this class shouldn't be dependent on how the template is placed in `AbstractTemplateFeature`, which means some sort of changes here.

I do have an idea I'd like to try out for that, and I think I would like to do it in my own PR. With that in mind, I am fine merging this without any additional changes in this PR.

kirderf1 (approved):

PR #557, opened 2024-02-01
Introduce TemplatePlacement as a helper object for template features

kirderf1:
The new class wraps around the placement of a template to help avoid basic positioning mistakes when writing template placement code, while also simplifying the process. This replaces the need for `AbstractTemplateFeature` which has consequently been removed/inlined.

Also fixes some minor issues with placement (see commit messages). I don't think these fixes are significant enough to put in the changelog.


Dweblenod (approved):
Seeing as not all the features that extended AbstractTemplateFeature made use of all its functionality, and not everything might make sense to extend it in the future in order to make use of one piece of functionality or another, this seems like a good move. In-game I havent noticed any changes to generation

PR #559, opened 2024-02-03
Iron-lass Remodel

KingBalDoro:
- the item class now has all the components needed for animation rendering in real time.
- the armor has two animation states, flight, and non-flight
- fully textured, and remodeled armor
- Model, animation and texture by DORO


src/main/java/com/mraof/minestuck/client/model/armor/IronLassArmorModel.java
```diff
@@ -0,0 +1,31 @@
...
+import com.mraof.minestuck.Minestuck;
+import com.mraof.minestuck.item.armor.IronLassArmorItem;
+import net.minecraft.resources.ResourceLocation;
+import software.bernie.geckolib.model.GeoModel;
+
+public class IronLassArmorModel extends GeoModel<IronLassArmorItem>
+{
+	/**
+	 * Model made by DORO with edits by Riotmode
+	 */
```
kirderf1, line 12: This is not where the model is defined. Keep such credit comments in the json file(s) instead only.

src/main/java/com/mraof/minestuck/client/renderer/armor/IronLassArmorRenderer.java
kirderf1, line 1: I don't think this class serves any meaningful purpose. I suggest that you use `GeoArmorRenderer` directly instead.

kirderf1 (approved):

PR #563, opened 2024-02-11
Data driven dialogue

Dweblenod:
A full rework of how dialogue is handled. Key changes are:
- Dialogue is data driven now, meaning new dialogue options can be created via datapack
- Dialogue now occurs in a gui/screen instead of through chat


src/main/java/com/mraof/minestuck/util/Dialogue.java
```diff
@@ -0,0 +1,496 @@
...
+			Type(TriPredicate<LivingEntity, Player, Pair<String, String>> conditions)
+			{
+				this.conditions = conditions;
+			}
+		}
+		
+		private final Type type;
+		private final String content;
+		@Nullable
+		private final String contentExtra;
```
kirderf1, line 168: I'm not a great fan of keeping these in their unparsed form. I suggest some polymorphism here, with each condition type having a class with fields for their content parsed into the right types. Same goes for the triggers.

src/main/java/com/mraof/minestuck/util/DialogueManager.java
kirderf1, line 1: With `LandTypeExtensions` I had the idea to define the loader as an inner class as opposed to making the same class the data storer and loader. What do you think of using the same code pattern here?
The main benefit is that the object that stores the data doesn't have to be created before the data has been loaded. Applied here, that means that we can make sure that once you have an `DialogueManager` object, that guarantees that the `dialogues` inside have been properly loaded and won't be null.

kirderf1, line 1: There's plenty of things that can be null here that haven't been marked as such. For starters, you should add nullability annotations. Then, for each of these cases consider:
- Potentially returning an optional instead of a nullable value.
- Potentially throwing an exception instead of returning null.

kirderf1 (changes requested):

src/main/java/com/mraof/minestuck/util/DialogueManager.java
```diff
@@ -0,0 +1,87 @@
...
+		if(dialogues.isEmpty())
+			return null;
+		else
+			return dialogues.stream().toList().get(rand.nextInt(dialogues.size()));
+	}
+	
+	public Dialogue getDialogue(ResourceLocation location)
+	{
+		Optional<Dialogue> potentialDialogue = dialogues.stream().filter(dialogue ->
+				dialogue.getPath().equals(location)).findAny();
```
kirderf1, line 73: Since this is one of the primary ways that dialogue is looked up, it seems reasonable to have an id-to-dialogue map, potentially replacing the dialogues list.

src/main/java/com/mraof/minestuck/entity/consort/ConsortEntity.java
```diff
@@ -77,6 +80,7 @@ public class ConsortEntity extends AnimatedPathfinderMob implements MenuProvider
 	
 	private final AnimatableInstanceCache cache = GeckoLibUtil.createInstanceCache(this);
 	private final EnumConsort consortType;
+	private Dialogue dialogue;
```
kirderf1, line 83: Since dialogue can be reloaded mid-game, this should likely not be storing a reference to the dialogue object directly. I suggest storing the dialogue path instead.

src/main/java/com/mraof/minestuck/entity/DialogueEntity.java
```diff
@@ -0,0 +1,34 @@
...
+import net.minecraft.resources.ResourceLocation;
+import net.minecraft.world.entity.LivingEntity;
+
+public interface DialogueEntity
+{
+	String DIALOGUE_NBT_TAG = "Dialogue";
+	
+	Dialogue getDialogue();
+	
+	void setDialogue(String location);
```
kirderf1, line 15: I suggest changing this function to take in a resource location or the dialogue itself, to take away the responsibility of the "string -> resource location" parse or the dialogue lookup away from the `DialogueEntity` implementor + make it possible to set the other things directly.

(Unless you want to update all `setDialogue()` functions immediately, you could then add back an overload to `setDialogue()` that takes a string and then parses it to call the other function)

src/main/java/com/mraof/minestuck/entity/consort/ConsortEntity.java
```diff
@@ -450,6 +477,25 @@ public static boolean canConsortSpawnOn(EntityType<ConsortEntity> entityType, Le
 		return true;
 	}
 	
+	@Override
+	public Dialogue getDialogue()
+	{
+		if(dialogue != null)
+			return dialogue;
+		else
+			dialogue = getDialogueFromEntity(this);
```
kirderf1, line 486:
```suggestion
		if(dialogue != null)
			return dialogue;
		
		dialogue = getDialogueFromEntity(this);
```

The `else` makes it look like it only might run with the code outside and after the `else`, when it in fact will always run with that code.

kirderf1:
```suggestion
		if(dialogue == null)
			dialogue = getDialogueFromEntity(this);
```
This is another way of doing it, which would make it match the if-statement below. I'm not entirely sure what to prefer.

src/main/java/com/mraof/minestuck/entity/consort/ConsortEntity.java
```diff
@@ -141,7 +145,20 @@ protected InteractionResult mobInteract(Player player, InteractionHand hand)
 	{
 		if(this.isAlive() && !player.isShiftKeyDown() && eventTimer < 0)
 		{
-			if(!level().isClientSide && player instanceof ServerPlayer serverPlayer)
+			Dialogue dialogue = getDialogue();
+			
+			//TODO there is server side data that needs to be made accessible client side
```
kirderf1, line 150: One option that I think could work great is to send a server-to-client packet with the necessary dialogue info, which then displays the screen from there.

The alternative would be to try and sync the entity dialogue whenever it changes to nearby players, and have dialogue start directly at the client-side. I think that's the worse option since it's slightly harder to sync the right data, and harder to add extra server-side logic to dialogue start, even if it means that the dialogue screen can pop up earlier without a back-and-forth with the server-side.

src/main/java/com/mraof/minestuck/network/DialoguePacket.java
```diff
@@ -0,0 +1,54 @@
...
+		return new DialoguePacket(trigger, buffer.readInt());
+	}
+	
+	@Override
+	public void execute(ServerPlayer player)
+	{
+		Entity entity = Minecraft.getInstance().level.getEntity(entityID);
+		if(entity instanceof LivingEntity livingEntity && entity instanceof DialogueEntity)
+		{
+			trigger.testConditions(livingEntity, player);
```
kirderf1, line 51: I'm guessing you're aware of it already, but this will need some sort of validation for the trigger from the packet, or better yet, a packet with info to look up the trigger for a response from the dialogue server-side.

This is less of a concern right now, and more a concern towards finalizing this feature/PR.

src/main/java/com/mraof/minestuck/entity/DialogueEntity.java
```diff
@@ -0,0 +1,34 @@
...
+		CompoundTag entityNBT = entity.getPersistentData();
+		if(entityNBT.contains(DIALOGUE_NBT_TAG))
+		{
+			String storedDialogue = entityNBT.getString(DIALOGUE_NBT_TAG);
+			
+			return dialogueFromLocation(storedDialogue);
+		}
+		
+		return null;
+	}
```
kirderf1, line 28: Not sure what the purpose of this is since I see no matching write to nbt.

kirderf1 (changes requested):

src/main/java/com/mraof/minestuck/util/Dialogue.java
```diff
@@ -0,0 +1,496 @@
+package com.mraof.minestuck.util;
+
+import com.google.gson.*;
+import com.ibm.icu.impl.Pair;
```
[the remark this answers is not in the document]
Dweblenod: Should be fixed now!

src/main/java/com/mraof/minestuck/util/Dialogue.java
```diff
@@ -0,0 +1,496 @@
...
+		}
+		
+		//TODO since activation of these conditions occurs from a client packet to the server, we may want to check validity
+		public void testConditions(LivingEntity entity, Player player)
+		{
+			type.conditions.accept(entity, player, Pair.of(content, getContentExtra()));
+		}
+	}
+	
+	public static <T extends Enum<T>> boolean enumExists(Class<T> enumClass, String enumName)
```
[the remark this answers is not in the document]
Dweblenod: Im not sure I see a good way to do this

src/main/java/com/mraof/minestuck/client/gui/DialogueScreen.java
```diff
@@ -0,0 +1,192 @@
...
+		for(int i = 0; i < filteredResponses.size(); i++)
+		{
+			Dialogue.Response response = filteredResponses.get(i);
+			String responseMessage = response.response();
+			int yPositionOffset = 20 * i;
+			
+			Component buttonComponent = Component.translatable(responseMessage);
+			
+			ExtendedButton entryButton = new ExtendedButton(xOffset + 20, yOffset + 40 + yPositionOffset, 190, 14, buttonComponent,
+					button -> clickResponse(responseMessage));
```
kirderf1, line 92: Assuming that `this.dialogue` or `this.dialogue.responses()` won't change as long as this screen exists, you could pass `response` directly into `clickResponse()` without going the roundabout way through `responseMessage`.

src/main/java/com/mraof/minestuck/entity/dialogue/condition/Condition.java
```diff
@@ -0,0 +1,840 @@
...
+import net.minecraftforge.registries.tags.ITag;
+
+import javax.annotation.Nullable;
+import java.util.*;
+import java.util.function.Function;
+import java.util.stream.Collectors;
+
+/**
+ * A Condition controls whether a piece of dialogue will show up
+ */
```
kirderf1, line 47: This javadoc need to be updated.

PR #565, opened 2024-02-12
Carved Bush

glubtier:
Have a Carved Bush. As Aki is working on the overall Wood terrain glow-up, I didn't put the bush in the changelog. Presumably, the "WOOD_TERRAIN_BLOCKS" tag will also be updated to reflect whatever the terrain blocks are. I've already provided the updates for the tag, to accommodate fixes Aki is working on for the Wooden Cactus. (Real PR) (NOT CLICKBAIT)

Texture by Glubtier.


kirderf1:
Eh, I think it should still be added to the changelog right now just in case. Aki could just combine it with a future changelog entry if it becomes relevant. Better than risking it getting left out on accident later.

Datagen also need to be done and committed, which I would've gone in and solved myself if everything else was clear. Otherwise I see no problem merging this.

glubtier:
I did run data... twice... and made sure it was committed......... IntelliJ is going to be the start of my Joker era. I'll see what's wrong.

10/4 on the changelog.

kirderf1 (approved):

PR #566, opened 2024-02-19
Ironlass remodel addendum

KingBalDoro:
edited UVs and texture mapping by riotmode


kirderf1:
@KingBalDoro Has this model gotten the revisions that it was waiting for, so that the PR is now ready to be merged, or are there any more changes that need to be pushed?

PR #567, opened 2024-02-20
Parameterizing renderers and using defaulted model class

kirderf1:
The three model classes `UnderlingModel`, `ConsortModel` and `PawnModel` were made to work with multiple entity types each by checking the type of the entity in the model function. However because each model object and renderer object are created for one entity type specifically, it made more sense to parameterize the renderer based on the entity type and then use that.

From there, it became fitting to use `DefaultedEntityGeoModel` for these models. Since the model classes weren't doing beyond what `DefaultedEntityGeoModel` could do at that point, I replaced said model classes entirely. However I could consider adding them back to make it easier for *if* we in the future want to do anything extra that requires extending the model class with our own.


src/main/java/com/mraof/minestuck/client/renderer/entity/UnderlingRenderer.java
```diff
@@ -29,13 +30,19 @@
...
+	
+	public UnderlingRenderer(EntityRendererProvider.Context context, String underlingName)
 	{
 		// this renderer does two simple things :
-		super(context, new UnderlingModel<>()); // render the entity with the base layer merged with a grist texture
-		this.addRenderLayer(new UnderlingDetailsLayer(this)); // render a second layer with details (eyes, mouth, etc)
+		// render the entity with the base layer merged with a grist texture
+		super(context, new DefaultedEntityGeoModel<>(Minestuck.id("underlings/" + underlingName), true));
+		// render a second layer with details (eyes, mouth, etc)
+		this.addRenderLayer(new UnderlingDetailsLayer<>(this, underlingName));
```
Dweblenod, line 44: Weve had an issue since our Geckolib release with the extra layers on underlings not turning red when the entity is injured. If you happen to know of a quick solution then now may be the time to implement it. Otherwise, no big deal!

Dweblenod (approved):
Looks good! Just had a single comment to make. Nice to see we can cut down on code

src/main/java/com/mraof/minestuck/client/renderer/entity/UnderlingRenderer.java
kirderf1: Found a fix to this. While it's very small, I think I'd prefer committing it separately. I'll add it directly to the 1.20.1 branch.

PR #568, opened 2024-02-23
Fix fall damage for minestuck fluids

kirderf1:
Adds a mixin injection to apply fall distance resetting from minestuck fluids at the appropriate time. I have tested it in-dev, but I have not tested it outside yet. That has to be done too to make sure that this is correctly set up with remapping/obfuscation in mind.
Note: with the mixin additions on the gradle side of things, people need to regenerate their runs with this change.


Dweblenod:
Still havent gotten a chance to check it out in game, sorry! In the meantime I am noticing the conflict with CHANGELOG.md. Glad we will be finally free from this bug

Dweblenod (approved):

PR #570, opened 2024-02-24
Prevent oversized stacks when captchaloguing slots

### LINKED ISSUE #554: Stacks of 128 or more items are invisible in the sylladex
The items are still stored in the sylladex and will work as expected, except they are not rendered nor clickable (only for the tree modus). The card also cannot be removed by right-clicking it.

sylladex showing chainmail leggings, 127 chests, and 2 chainmail leggings separated by an empty space

The empty card in the picture is a stack of 128 chests

kirderf1:
Closes #554.


Dweblenod (approved):
It works, yippee!

PR #574, opened 2024-03-01
Aspect tree data class

kirderf1:
Builds upon #552 by creating a new class `AspectTreeBlocksData` to hold all data definitions for aspect tree blocks. This is meant to give two benefits:
- Makes sure that all data generation for these blocks can be found in the same class.
- Thin out the data provider classes where these data definitions were moved from.

While one could try to push it even further so that data for the same block is found in the same code block, I chose not to go that far since I didn't think that it would be worth the extra effort right now.


kirderf1:
I considered moving creative tabs stuff to `AspectTreeBlocks`, but ended up not going through with it.
Because of how item order matters with creative tabs, I think there's a stronger reason to continue adding items for the same tab in the same place.

PR #584, opened 2024-03-24
Made predetermined Captchas Data oriented

tuffythebeloved:
Added a Predetermined Captcha class that loads predetermined captchas from user Datapacks.


src/main/java/com/mraof/minestuck/alchemy/PredeterminedCaptchaLoader.java
```diff
@@ -0,0 +1,96 @@
...
+	@Override
+	protected Map<String, ResourceLocation> prepare(ResourceManager resourceManager, ProfilerFiller profiler)
+	{
+		Map<String, ResourceLocation> captchaData = new HashMap<>();
+		
+		for(String namespace : resourceManager.getNamespaces())
+		{
+			parseCaptchaCodesFromNamespace(resourceManager, namespace).ifPresent(parsedData -> {
+				for(String captcha : parsedData.keySet())
+					captchaData.put(captcha, parsedData.get(captcha));
```
kirderf1, line 44: Overlap in captchas and items between namespaces could be troublesome. Captchas and items need to be checked against the predefined ones that has already been added to solve that.

src/main/java/com/mraof/minestuck/alchemy/PredeterminedCardCaptchas.java
```diff
@@ -0,0 +1,20 @@
...
+import net.minecraft.world.item.Item;
+
+
+public class PredeterminedCardCaptchas
+{
+	public static BiMap<Item, String> predefinedCardMap = HashBiMap.create();
+	
+	public PredeterminedCardCaptchas()
+	{
+	}
```
kirderf1, line 14: This is a constructor that you're not making use of.

src/main/java/com/mraof/minestuck/alchemy/PredeterminedCardCaptchas.java
```diff
@@ -0,0 +1,20 @@
+package com.mraof.minestuck.alchemy;
+
+import com.google.common.collect.BiMap;
+import com.google.common.collect.HashBiMap;
+import net.minecraft.world.item.Item;
+
+
+public class PredeterminedCardCaptchas
+{
+	public static BiMap<Item, String> predefinedCardMap = HashBiMap.create();
```
kirderf1, line 10: I get the impression that this field shouldn't really be public.
Try lowering the visibility and adding some functions that control how this map may be accessed.

src/main/java/com/mraof/minestuck/alchemy/CardCaptchas.java
```diff
@@ -63,7 +71,22 @@ public static String getCaptcha(Item item, MinecraftServer server)
...
+			captchas.captchasMap.inverse().remove(captcha);
+		}
+		
+		if(PredeterminedCardCaptchas.predefinedCardMap.inverse().containsKey(captcha))
+		{
+			return PredeterminedCardCaptchas.predefinedCardMap.inverse().get(captcha);
+		}
+		else
+		{
+			return MSExtraData.get(server).getCardCaptchas().captchasMap.inverse().get(captcha);
```
kirderf1, line 88:
```suggestion
			return captchas.captchasMap.inverse().get(captcha);
```

kirderf1, line 88: When reaching this point, I think it'd be nice to also check for overlap of the item that was found.

src/main/java/com/mraof/minestuck/alchemy/CardCaptchas.java
kirderf1, line 1: When checking if a freshly generated captcha has already been generated, also check predefined captchas.

src/main/java/com/mraof/minestuck/alchemy/CardCaptchas.java
```diff
@@ -49,7 +48,16 @@ public static String getCaptcha(Item item, MinecraftServer server)
...
+		{
+			captchas.captchasMap.remove(item);
+		}
+		
+		if(PredeterminedCardCaptchas.predefinedCardMap.containsKey(item))
+		{
+			return PredeterminedCardCaptchas.predefinedCardMap.get(item);
+		}
+		else if(captchas.captchasMap.containsKey(item))
 			return captchas.captchasMap.get(item);
```
kirderf1, line 61: When reaching this point, I think it'd be nice to also check for overlap of the captcha that was found.

src/main/java/com/mraof/minestuck/alchemy/CardCaptchas.java
```diff
@@ -49,7 +48,16 @@ public static String getCaptcha(Item item, MinecraftServer server)
 		MSExtraData data = MSExtraData.get(server);
 		CardCaptchas captchas = data.getCardCaptchas();
 		
-		if(captchas.captchasMap.containsKey(item))
+		if(PredeterminedCardCaptchas.predefinedCardMap.containsKey(item) && captchas.captchasMap.containsKey(item))
+		{
+			captchas.captchasMap.remove(item);
```
kirderf1, line 53: Remember to call `data.setDirty();` whenever making changes to generated captchas. (also in `getItemFromCaptcha()`)

kirderf1 (changes requested):

src/main/generated/resources/data/minestuck/minestuck/captcha_codes.json
kirderf1, line 1: This file should go in the normal resources folder (`src/main/resources/...`) instead of the generated resources folder.

src/main/java/com/mraof/minestuck/alchemy/CardCaptchas.java
```diff
@@ -49,8 +47,28 @@ public static String getCaptcha(Item item, MinecraftServer server)
...
+			if(captchas.captchasMap.containsKey(item))
+			{
+				captchas.captchasMap.remove(item);
+				data.setDirty();
+			}
+			return PredeterminedCardCaptchas.GetData().get(item);
+		}
+		else if(captchas.captchasMap.containsKey(item))
+		{
+			if(PredeterminedCardCaptchas.GetData().containsKey(item))
```
kirderf1, line 61: This condition is checked earlier in this function and does not need repeating.

src/main/java/com/mraof/minestuck/alchemy/CardCaptchas.java
```diff
@@ -63,7 +81,31 @@ public static String getCaptcha(Item item, MinecraftServer server)
...
+			if(captchas.captchasMap.inverse().containsKey(captcha))
+			{
+				captchas.captchasMap.inverse().remove(captcha);
+				data.setDirty();
+			}
+			return PredeterminedCardCaptchas.GetData().inverse().get(captcha);
+		}
+		else
+		{
+			if(PredeterminedCardCaptchas.GetData().inverse().containsKey(captcha))
```
kirderf1, line 98: This condition is checked earlier in this function and does not need repeating.

src/main/java/com/mraof/minestuck/alchemy/CardCaptchas.java
```diff
@@ -117,17 +152,12 @@ private String createCaptchaForItem(Item item, long seed)
 		String cutHash = shuffledHash.substring(shuffledHash.length() - 16); //last 16 characters of hash
 		String captcha = captchaFromHash(cutHash);
 		
-		if(captchasMap.containsValue(captcha))
+		if(captchasMap.containsValue(captcha) || PredeterminedCardCaptchas.GetData().containsValue(captcha))
```
kirderf1, line 155: I see that generated captchas are also checked in `generateBackupCaptcha()`. Should we check predetermined captchas there too?

src/main/java/com/mraof/minestuck/alchemy/CardCaptchas.java
```diff
@@ -49,8 +47,28 @@ public static String getCaptcha(Item item, MinecraftServer server)
...
 		CardCaptchas captchas = data.getCardCaptchas();
 		
-		if(captchas.captchasMap.containsKey(item))
-			return captchas.captchasMap.get(item);
+		if(PredeterminedCardCaptchas.GetData().containsKey(item))
+		{
+			if(captchas.captchasMap.containsKey(item))
+			{
+				captchas.captchasMap.remove(item);
+				data.setDirty();
```
kirderf1, line 55: I had the thought that this should probably be logged as a warning. Same thing for the corresponding conflict removal in `getItemFromCaptcha()`.

src/main/java/com/mraof/minestuck/alchemy/PredeterminedCaptchaLoader.java
```diff
@@ -0,0 +1,107 @@
...
+						}
+						else if(captchaData.containsKey(captcha))
+						{
+							LOGGER.error("Code: {} is already assigned to Item: {}!", captcha, captchaData.get(captcha));
+						}
+						else
+						{
+							captchaData.put(entry.getKey(), entry.getValue());
+						}
+					}
```
kirderf1, line 62: Would it be a good idea to extract this as a separate function?

src/main/java/com/mraof/minestuck/alchemy/PredeterminedCaptchaLoader.java
```diff
@@ -0,0 +1,107 @@
...
+	
+	@Override
+	protected void apply(Map<String, Item> data, ResourceManager resourceManager, ProfilerFiller profiler)
+	{
+		ImmutableBiMap.Builder<Item, String> predefinedCards = ImmutableBiMap.builder();
+		
+		for(Map.Entry<String, Item> entrySet : data.entrySet())
+		{
+			predefinedCards.put(entrySet.getValue(), entrySet.getKey());
+			LOGGER.atInfo().log(entrySet.getValue() + " was added to card with string: " + entrySet.getKey());
```
kirderf1, line 76: Seems like a remnant line from some debugging, which ought to be removed.

src/main/java/com/mraof/minestuck/alchemy/PredeterminedCardCaptchas.java
```diff
@@ -0,0 +1,21 @@
...
+import com.google.common.collect.BiMap;
+import com.google.common.collect.HashBiMap;
+import net.minecraft.world.item.Item;
+
+
+public class PredeterminedCardCaptchas
+{
+	private static BiMap<Item, String> predefinedCardMap = HashBiMap.create();
+	
+	public static void SetData(BiMap<Item, String> predefinedCards)
```
kirderf1, line 12:
```suggestion
	static void SetData(BiMap<Item, String> predefinedCards)
```
Reduced visibility because it should only be accessed from `PredeterminedCaptchaLoader`.
Although if you turn `PredeterminedCaptchaLoader` into an inner class of `PredeterminedCardCaptchas` (similar to `ComputerThemes.Loader` or `DialogueNodes.Loader`), then you can even use private access to set it.

src/main/java/com/mraof/minestuck/alchemy/PredeterminedCardCaptchas.java
kirderf1, line 1: The new checkstyle action caught on to it, but apparently didn't run in the right way to make a review:
The functions here use *PascalCase* names, while the standard for all java functions is to use *camelCase* names. Rename `GetData()` ->`getData()` and `SetData()` -> `setData()`, and it should be fine.

kirderf1, line 1: A lot of similar datapack-loaded static data structures of ours take extra measures to only be accessible once loaded, throwing an exception when accessed at an inappropriate time. There may be a point to doing something like that here too.

kirderf1 (changes requested):
Please write out a short description of your changes in the PR description if you can.
Also merge 1.20.1 and add a changelog entry for this.

src/main/java/com/mraof/minestuck/alchemy/PredeterminedCardCaptchas.java
```diff
@@ -0,0 +1,132 @@
...
+import java.util.Map;
+import java.util.Objects;
+import java.util.Optional;
+
+@ParametersAreNonnullByDefault
+@MethodsReturnNonnullByDefault
+@Mod.EventBusSubscriber(modid = Minestuck.MOD_ID, bus = Mod.EventBusSubscriber.Bus.FORGE)
+public class PredeterminedCardCaptchas
+{
+	private static BiMap<Item, String> predefinedCardMap = HashBiMap.create();
```
kirderf1, line 38:
```suggestion
	@Nullable
	private static BiMap<Item, String> predefinedCardMap;
```

src/main/java/com/mraof/minestuck/alchemy/PredeterminedCardCaptchas.java
```diff
@@ -0,0 +1,132 @@
...
+			}
+			
+			PredeterminedCardCaptchas.setData(predefinedCards.build());
+		}
+		
+		@SubscribeEvent
+		public static void onResourceReload(AddReloadListenerEvent event)
+		{
+			event.addListener(new Loader());
+		}
```
kirderf1, line 88: This doesn't seem to be working since it's not directly inside the class with the `@EventBusSubscriber` annotation. A simple solution is to move it up to `PredeterminedCardCaptchas`.

kirderf1 (changes requested):

Dweblenod (approved):
Works well! 

The only bug I encountered is a cosmetic one, where the displayed/stored captcha for a card is not updated to match changes to its code. Returning it to the sylladex updates it correctly, and even with the wrong captcha displayed it still works with the totem lathe/punch designix.

Considering the complexity I dont think its necessarily worth it but there is room to make a data generator for this.

kirderf1 (approved):
There may be an error in the CI execution, but that error is unrelated to this. This looks good to go by me.

PR #587, opened 2024-03-28
Response messages and player message type

kirderf1:
Adds a message type for messages from the player, and adds a way to attach messages to a response which will then show up at the top of the dialogue screen following that response.

The two use-cases I had in mind for this is:
- Define player response text at the response itself instead of the node that the response leads to.
- Add a simpler way to add text that shows up when choosing a response, but not when dialogue starts in the node that follows. Like when adding text that describes an action and sets a new dialogue entrypoint, but without that description showing up when starting dialogue over from that entrypoint.

I have also demonstrated both of these use-cases. The first one which I've changed response texts to match what the corresponding text was in the old system.


Dweblenod (approved):
I think some of the dialogue needs reviewed but that can occur in a separate PR!

PR #590, opened 2024-04-03
Add dialogue commands for testing and reviewing dialogue

kirderf1:
Two new commands:
- `/review_dialogue <entity_type> <dialogue_category>` spawns entities of the given type for every dialogue in the given category (going by the starting node of all randomly selectable dialogue)
- `/set_dialogue <entity> <dialogue_id>` sets the dialogue node for the given entity

Both of these commands sets the dialogue such that it is kept on reset.


src/main/java/com/mraof/minestuck/command/ReviewDialogueCommand.java
```diff
@@ -0,0 +1,93 @@
...
+	private static int spawnDialogueEntities(CommandSourceStack source, EntityType<?> entityType, RandomlySelectableDialogue.DialogueCategory category) throws CommandSyntaxException
+	{
+		Collection<Dialogue.SelectableDialogue> dialogueCollection = RandomlySelectableDialogue.instance(category).getAll();
+		
+		ServerLevel level = source.getLevel();
+		BlockPos pos = BlockPos.containing(source.getPosition());
+		
+		for(Dialogue.SelectableDialogue dialogue : dialogueCollection)
+		{
+			pos = pos.east(3);
```
Dweblenod, line 59: I think this should be 2 instead. It takes a while to FLY from one end of the dialogue to the other so it could do with being more compact

src/main/java/com/mraof/minestuck/command/ReviewDialogueCommand.java
```diff
@@ -0,0 +1,93 @@
...
+			
+			BlockPos signPos = pos.north().below();
+			level.setBlock(signPos, Blocks.OAK_WALL_SIGN.defaultBlockState().setValue(WallSignBlock.FACING, Direction.NORTH), Block.UPDATE_ALL);
+			if(level.getBlockEntity(signPos) instanceof SignBlockEntity sign)
+			{
+				SignText signText = new SignText().setHasGlowingText(true).setColor(DyeColor.WHITE)
+						.setMessage(0, Component.literal(dialogue.dialogueId().getNamespace()));
+				String[] lines = dialogue.dialogueId().getPath().split("/", 3);
+				for(int i = 0; i < lines.length; i++)
+					signText = signText.setMessage(i + 1, Component.literal(lines[i]));
```
Dweblenod, line 84: I assume because they count as a single word, the names of long dialogues such as "a_little_lampshading" and "no_tower_treasure" get cut off

Dweblenod (commented):
Set Dialogue command looks good to go and there are just a few minor changes that might be worth getting made for Review Dialogue but otherwise this looks good!

src/main/java/com/mraof/minestuck/command/ReviewDialogueCommand.java
kirderf1: A bit unfortunate. Maybe it'd be possible to switch to something other than signs, like attaching a name to the entity like a name tag. That said, I don't think it is a particularly important issue in this specific situation.

Dweblenod (approved):

PR #591, opened 2024-04-04
Fix save format for effect in area effect block

kirderf1:
The registry int ids are not necessarily consistent between play sessions or save files, which makes it a bad idea to use it for serialization (with the potential exception of network serialization).

The area effect block was doing this with its effect, which is bad for both save files and structure templates. This pr changes this to use the string id instead, while keeping backwards-compatibility.


Dweblenod (approved):
Looks good in game!

PR #592, opened 2024-04-05
Avoid/catch problems setting up a StructureBlockRegistry

kirderf1:
Fixes a problem that @akisephila encountered. When an exception is thrown in `ILandType.registerBlocks()`, the error ended up not being logged, thus making it harder to find the actual problem. This PR fixes this issue in two ways:
- Uses by `TitleLandType.isAspectCompatible()` have been replaced by terrain land type tags.
- When the structure block registry is set up for `LandGenSettings`, any exceptions are now caught and logged.


Dweblenod (commented):
I am unsure how to really bug test this, but I can give my approval now for the general contents

kirderf1:
If one really want to test this, you could try out `TitleLandType.isAspectCompatible()` by predefining land types, and creating a land/loading a save file with a land to test the `LandGenSettings` part. One could try it as-is to see that everything works fine normally, and one could purposefully make a land throw an exception in `landType.registerBlocks()` and see that the exception is handled properly.

Dweblenod:
Using /sburbpredefine, it does correctly prevent a disallowed land type combo from being used, however it does not give a warning and automatically reassigns the offending land type

PR #594, opened 2024-04-06
Helper functions for common crafting recipes

kirderf1:
Introduces `CommonRecipes` with helper functions for creating common recipes during data-generation.

The new helper functions are also being used where appropriate.
Cruxite recipes originally had raw cruxite as their unlock criteria, but not the actual block ingredient. For the recipes involved in this PR, I also added the block ingredient as an alternative unlock criteria.

I was originally thinking of going over additional recipes, like doors, trapdoors, fences and fence gates. But I decided to cut it short in order to focus on other things with 4/13 around the corner.


Dweblenod (approved):
Now thats some good stuff : ) Its clear from the generated folder that there hasnt been too many changes while switching to these helper functions so I wont worry about testing it in game.

The only thing I could think of would be to add more templates to CommonRecipes.

PR #595, opened 2024-04-06
Rain land tower structure

kirderf1:
Adds a new tower building as a structure to rain lands.
Credit for the structure template goes to ScarabOasis.


Dweblenod:
Still waiting on the button in the currently inescapable room, but otherwise it seems ready to merge!

PR #600, opened 2024-04-11
CommandWithFlags Trigger

Dweblenod:
A stretch goal for 4/13 (can take lower priority)! Its just useful in situations where more than one set of flags can result in a mess for command usage.

Below is an example of usage, a modified version of mushroom_pizza. A command is triggered when the last response option is taken. It starts off as "execute as @s in [flag0] run tp ~[flag1] ~ ~". [flag0] corresponds to command.0.minecraft:overworld and [flag1] corresponds to command.1.500. 

- If neither flag is fulfilled, the command runs as "execute as @s in  run tp ~ ~ ~".
- If the first flag is fulfilled, the command runs as "execute as @s in minecraft:overworld run tp ~ ~ ~"
- If the second flag is fulfilled, the command runs as "execute as @s in  run tp ~500 ~ ~"
- If both flags are fulfilled, the command runs as "execute as @s in minecraft:overworld run tp ~500 ~ ~"

```json
{
  "node": {
    "message": "minestuck.dialogue.consort.mushroom_pizza.start",
    "responses": [
      {
        "message": "set dimension flag",
        "next_dialogue": {
          "id": "minestuck:consort/mushroom_pizza/on",
          "player_message": "minestuck.dialogue.consort.mushroom_pizza.start.on.reply",
          "set_as_entrypoint": false
        },
        "triggers": [
          {
            "type": "minestuck:set_flag",
            "flag": "command.0.minecraft:overworld",
            "player_specific": true
          }
        ]
      },
      {
        "message": "set coord flag",
        "triggers": [
          {
            "type": "minestuck:set_flag",
            "flag": "command.1.500",
            "player_specific": true
          }
        ]
      },
      {
        "message": "minestuck.dialogue.consort.mushroom_pizza.start.off",
        "next_dialogue": {
          "id": "minestuck:consort/mushroom_pizza/off",
          "player_message": "minestuck.dialogue.consort.mushroom_pizza.start.off.reply",
          "set_as_entrypoint": false
        },
        "triggers": [
          {
            "type": "minestuck:command_with_flags",
            "command": "execute as @s in [flag0] run tp ~[flag1] ~ ~",
            "flag_amount": 2
          }
        ]
      }
    ]
  }
}
```

### DIFF
--- a/src/main/java/com/mraof/minestuck/entity/dialogue/Trigger.java
+++ b/src/main/java/com/mraof/minestuck/entity/dialogue/Trigger.java
@@ -30,6 +30,7 @@ import net.minecraftforge.registries.ForgeRegistries;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 
+import java.util.ArrayList;
 import java.util.List;
 import java.util.Optional;
 import java.util.Set;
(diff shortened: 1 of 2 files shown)


kirderf1:
I'm not so confident in making flags hold extra data like that. I might suggest something else, but I would need to see a few practical use cases of this to do that.

Dweblenod:
> I'm not so confident in making flags hold extra data like that. I might suggest something else, but I would need to see a few practical use cases of this to do that.

In the end, I created this as a means to allow for more dynamic uses of commands. The main use I had in mind was actually related to the public server, where dialogue options are used to pick the Land the player gets. I wasnt planning on making it this complex, but with that scenario I imagined what would be required in order to make a command that does everything /sburbpredefine allows (title land type / terrain land type / class / aspect)

kirderf1:
The idea I have in mind then is a player-specific "string key to string value" map. A trigger that sets a value for a key, and the command trigger which pulls the value for a certain key.

Depending on how much you want it ready for the public server, I could step in and make the changes that I am thinking of before tomorrow.

Dweblenod:
I think this would be too much given the hour, it can be a goal for later!

kirderf1:
Deleted the `quest_systems` branch since that project was finished up and merged into 1.20.1. But this PR was apparently targeting that branch, and was closed as a result. The PR should probably be reopened or remade, targeting some other branch.

PR #610, opened 2024-04-20
PlayerData data attachments

kirderf1:
The goal is to convert most (potentially all) data in `PlayerData` into attachments, as well as doing general code cleanup on code adjacent to this data.

I'm also considering renaming `PlayerData` to reflect it being available even when the player is offline.

The same treatment could be given to `SburbPlayerData` (and moving some data from it over to `PlayerData`), but I'll leave it for a future PR to limit the scope of this one.

As Minestuck hasn't been released for Minecraft 1.20.4 yet, I am not attempting to keep backwards-compatibility for these changes.


kirderf1:
This should be ready for review.

I didn't get any specific ideas for renaming `PlayerData`/`PlayerSavedData`. If anyone has any thoughts, feel free to add them through a review.

src/main/java/com/mraof/minestuck/item/BoondollarsItem.java
```diff
@@ -29,9 +29,11 @@ public BoondollarsItem(Properties properties)
 	@Override
 	public InteractionResultHolder<ItemStack> use(Level level, Player playerIn, InteractionHand handIn)
 	{
-		if(!level.isClientSide && !(playerIn instanceof FakePlayer))
+		if(playerIn instanceof ServerPlayer serverPlayer)
```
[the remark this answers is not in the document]
kirderf1: No. A `FakePlayer` is definitely an instance of `ServerPlayer` still.
Rather, it has become redundant in some places by actually checking the return value of  `PlayerSavedData.getData(ServerPlayer)`/`PlayerData.get(ServerPlayer)`. I changed the nullable return of that function to an optional, and along with it also changed a lot of places to use `ifPresent()` instead of throwing, like here. Because the data is null/missing specifically for fake players, this works out to make most `FakePlayer` checks redundant.

Dweblenod (commented):
Looks like there are some backwards compatibility issues. I created a world while on the main branch, saved the game, and opened it back up under this branch. The following got reset:
- Grist cache
- Echeladder rung
- Porkhollow

I also noticed some issues with getting the players Title

src/main/java/com/mraof/minestuck/alchemy/GristGutter.java
Dweblenod, line 1: The changes actually seem to fix the bug with the gutter modifier not seeming to be saved! Yippee!

kirderf1: Yeah! But only here on latest. #612.

src/main/java/com/mraof/minestuck/player/Title.java
```diff
@@ -1,79 +1,128 @@
...
-				}
-				
-				if(potionLevel > 0)
-				{
-					player.addEffect(new MobEffectInstance(aspectEffects[aspect.ordinal()], 600, potionLevel - 1));
-					LOGGER.debug("Applied aspect potion effect to {}", player.getDisplayName().getString());
-				}
-			}
+			player.addEffect(new MobEffectInstance(ASPECT_EFFECTS.get(aspect).effect(), 600, potionLevel - 1));
+			LOGGER.debug("Applied aspect potion effect to {}", player.getDisplayName().getString());
```
Dweblenod, line 125: The log message applies twice at the same time, making me think the effect is also being applied twice

kirderf1: That might be `ServerEventHandler` listening to the player tick event and handling both the start and end phase of the tick.

src/main/java/com/mraof/minestuck/inventory/captchalogue/CaptchaDeckHandler.java
Dweblenod, line 1: Modus and items held within sylladex seem to be getting reset upon reload

src/main/java/com/mraof/minestuck/util/MSAttachments.java
```diff
@@ -0,0 +1,87 @@
...
+			() -> AttachmentType.serializable(restricted(() -> new ItemStackHandler(1), ItemStack.class)).build());
+	public static final Supplier<AttachmentType<MusicPlaying>> MUSIC_PLAYING = REGISTER.register("music_playing",
+			() -> AttachmentType.builder(restricted(MusicPlaying::new, LivingEntity.class)).build());
+	
+	public static final Supplier<AttachmentType<EditTools>> EDIT_TOOLS = REGISTER.register("edit_tools",
+			() -> AttachmentType.builder(restricted(EditTools::new, Player.class)).build());
+	public static final Supplier<AttachmentType<MSFluidType.LastFluidTickData>> LAST_FLUID_TICK = REGISTER.register("last_fluid_tick",
+			() -> AttachmentType.builder(restricted(MSFluidType.LastFluidTickData::new, LivingEntity.class)).build());
+	public static final Supplier<AttachmentType<DialogueComponent.CurrentDialogue>> CURRENT_DIALOGUE = REGISTER.register("current_dialogue",
+			() -> AttachmentType.builder(restricted(DialogueComponent.CurrentDialogue::new, ServerPlayer.class)).build());
```
Dweblenod, line 65: Comment on layout, I think it would make sense to group the following together
- edit tools and editmode locations
- current dialogue and consort reputation

Dweblenod (approved):

PR #616, opened 2024-05-01
Block Teleporter

Dweblenod:
A block for teleporting other blocks! A redstone block intended to be used in dungeon settings but which can be obtained in survival.

Block textures are by sipherNil except block_teleport_front.png which is an edited version of block_teleport_side.png by Dweblenod


src/main/java/com/mraof/minestuck/blockentity/redstone/BlockTeleporterBlockEntity.java
```diff
@@ -0,0 +1,119 @@
...
+		super.saveAdditional(compound);
+		
+		compound.putInt("offsetX", teleportOffset.getX());
+		compound.putInt("offsetY", teleportOffset.getY());
+		compound.putInt("offsetZ", teleportOffset.getZ());
+	}
+	
+	public static BlockPos clampPos(int x, int y, int z)
+	{
+		return new BlockPos(Math.min(Math.max(x, -32), 32), Math.min(Math.max(y, -32), 32), Math.min(Math.max(z, -32), 32));
```
kirderf1, line 105:
```suggestion
		return new BlockPos(Mth.clamp(x, -32, 32), Mth.clamp(y, -32, 32), Mth.clamp(z, -32, 32));
```

src/main/java/com/mraof/minestuck/network/BlockTeleporterPacket.java
```diff
@@ -0,0 +1,64 @@
...
+	
+	@SuppressWarnings("resource")
+	@Override
+	public void execute(ServerPlayer player)
+	{
+		if(player.level().isAreaLoaded(beBlockPos, 0))
+		{
+			if(player.level().getBlockEntity(beBlockPos) instanceof BlockTeleporterBlockEntity blockTeleporter)
+			{
+				if(Math.sqrt(player.distanceToSqr(beBlockPos.getX() + 0.5, beBlockPos.getY() + 0.5, beBlockPos.getZ() + 0.5)) <= 8)
```
kirderf1, line 53: Could you combine and invert these?

src/main/java/com/mraof/minestuck/network/BlockTeleporterPacket.java
```diff
@@ -0,0 +1,64 @@
...
+	{
+		BlockPos offsetPos = buffer.readBlockPos();
+		BlockPos beBlockPos = buffer.readBlockPos();
+		
+		return new BlockTeleporterPacket(offsetPos, beBlockPos);
+	}
+	
+	@SuppressWarnings("resource")
+	@Override
+	public void execute(ServerPlayer player)
```
kirderf1, line 47: Like when the block is activated and the screen opened, I think that creative shock should be checked here.

src/main/java/com/mraof/minestuck/network/BlockTeleporterPacket.java
```diff
@@ -0,0 +1,64 @@
...
+
+import com.mraof.minestuck.Minestuck;
+import com.mraof.minestuck.blockentity.redstone.BlockTeleporterBlockEntity;
+import net.minecraft.core.BlockPos;
+import net.minecraft.network.FriendlyByteBuf;
+import net.minecraft.resources.ResourceLocation;
+import net.minecraft.server.level.ServerPlayer;
+import net.minecraft.world.level.block.state.BlockState;
+
+public class BlockTeleporterPacket implements MSPacket.PlayToServer
```
kirderf1, line 11: Make this a record?

kirderf1 (changes requested):
I've got a PR on its way after #610 that cleans up packets. In it I've already applied the comments I made on `BlockTeleporterPacket` to other packets, so you shouldn't do the same on any packets other than `BlockTeleporterPacket`.

kirderf1 (approved):

PR #617, opened 2024-05-02
Particle Accelerator Block

InfectedGriffon:
currently implemented as a non-functional block entity for ease of updating


src/main/java/com/mraof/minestuck/blockentity/machine/ParticleAcceleratorBlockEntity.java
kirderf1, line 1: This does seem redundant in its current state. What's the plan for this block?

Dweblenod:
Since the block entity mini-game (discussed in Discord) is still under development but the rest of this is finished, I recommend removing the block entity for now and submitting as is!

Dweblenod:
@InfectedGriffon is there still a plan for this? Can just the block be added?

PR #621, opened 2024-05-09
Split general chest loot table into reusable parts

kirderf1:
Takes the loot pools of the loot table `minestuck:chests/medium_basic` and separates out their loot entries to form new loot tables `minestuck:chests/weapon_item`, `minestuck:chests/supply_item`, `minestuck:chests/misc_item` and `minestuck:chests/rare_item`. These are then referenced from the original loot table.

These new loot tables can then be reused if desired when creating new loot tables.

Also removes the item damage function from shuriken entries because shurikens do not use durability.


Dweblenod (approved):
Looks good!

PR #623, opened 2024-05-10
Holopad-focused code cleanup

kirderf1:
Removes the unused hologram entity (the current version of the holopad functions through its block entity renderer) and some code associated with it, and performs some general code cleanup on both the block and block entity.

I also changed the holopad to display a generic object instead of the contained item when the card isn't punched, instead of just when the card is empty. This is to make the holopad behave as if it was specifically reading the item off of the punched holes of the card. Not sure whether to add this to the changelog or not.


src/main/java/com/mraof/minestuck/block/machine/HolopadBlock.java
```diff
@@ -34,21 +33,11 @@
 @MethodsReturnNonnullByDefault
 public class HolopadBlock extends MachineBlock implements EntityBlock
 {
-	public static final Map<Direction, VoxelShape> SHAPE = createRotatedShapes(2, 0, 1, 14, 6, 13);
-	public static final Map<Direction, VoxelShape> COLLISION_SHAPE;
-	//todo these are unused?
-	protected static final AABB HOLOPAD_TOP_AABB = new AABB(3/16F, 6/16F, 2.6/16F, 13/16F, 7/16F, 12.6/16F);
-	protected static final AABB HOLOPAD_CARDSLOT_AABB = new AABB(4/16F, 0F, 13.8/16F, 12/16F, 10.1/16F, 15.94/16F);
+	public static final Map<Direction, VoxelShape> SHAPE = new CustomVoxelShape(new double[]{2, 0, 1, 14, 6, 13}).createRotatedShapes();
+	public static final Map<Direction, VoxelShape> COLLISION_SHAPE = new CustomVoxelShape(new double[]{4, 0, 14, 12, 10, 16}, new double[]{3, 6, 3, 13, 7, 13}).createRotatedShapes();
```
Cibernet83, line 37: Collision Shape doesn't include the main Interaction Shape cube, meaning that you can clip through the Holopad from below.

src/main/java/com/mraof/minestuck/block/machine/HolopadBlock.java
```diff
@@ -34,21 +33,11 @@
 @MethodsReturnNonnullByDefault
 public class HolopadBlock extends MachineBlock implements EntityBlock
 {
-	public static final Map<Direction, VoxelShape> SHAPE = createRotatedShapes(2, 0, 1, 14, 6, 13);
-	public static final Map<Direction, VoxelShape> COLLISION_SHAPE;
-	//todo these are unused?
-	protected static final AABB HOLOPAD_TOP_AABB = new AABB(3/16F, 6/16F, 2.6/16F, 13/16F, 7/16F, 12.6/16F);
-	protected static final AABB HOLOPAD_CARDSLOT_AABB = new AABB(4/16F, 0F, 13.8/16F, 12/16F, 10.1/16F, 15.94/16F);
+	public static final Map<Direction, VoxelShape> SHAPE = new CustomVoxelShape(new double[]{2, 0, 1, 14, 6, 13}).createRotatedShapes();
```
Cibernet83, line 36: I believe the Holopad's Interaction Shape should match that of its Collision. Or at the very least add a box for the back part where the Captchalogue Card is supposed to be inserted.

Cibernet83 (changes requested):
https://github.com/lunar-sway/minestuck/blob/731f9f0a8a8620f12098547a0717204269134cad/src/main/java/com/mraof/minestuck/client/renderer/blockentity/HolopadRenderer.java#L29
`57.295776F` is representative of 180/pi, or `Mth.RAD_TO_DEG`. Considering `Axis.rotation` is handled in radians this value doesn't exactly make sense, perhaps we could optimize or make the code more legible by going with a different value?

https://github.com/lunar-sway/minestuck/blob/main/src/main/java/com/mraof/minestuck/block/CustomVoxelShape.java
`CustomVoxelShape` is only used to create easily rotatable Collision and Interaction shapes, perhaps it could be refactored to `RotatableVoxelShape` to better represent its purpose.

https://github.com/lunar-sway/minestuck/blob/main/src/main/resources/assets/minestuck/models/block/holopad_has_card.json
The Holopad's card-containing model currently has a Solid Render Type rather than Cutout, Making the Captchalogue card present in the model have a stray white pixel, we could either update its Render Type or change the model to separate the card's cubes into two.

src/main/java/com/mraof/minestuck/block/machine/HolopadBlock.java
kirderf1: I've combined the two, and also made small further adjustments.

kirderf1:
> https://github.com/lunar-sway/minestuck/blob/main/src/main/java/com/mraof/minestuck/block/CustomVoxelShape.java `CustomVoxelShape` is only used to create easily rotatable Collision and Interaction shapes, perhaps it could be refactored to `RotatableVoxelShape` to better represent its purpose.

Messing around with `CustomVoxelShape` is a bit out of scope for this, and I would also rather not increase the scope of this PR at this time.

Dweblenod (approved):

Cibernet83 (approved):

PR #624, opened 2024-05-11
Util package cleanup

kirderf1:
The goal was to clean up and move things out of the util package, attempting to get rid of the package entirely.

I wasn't able to get rid of the package entirely, but I did thin it out a lot.

Unneeded things were removed, code was cleaned up, and classes were move to relevant existing packages.


src/main/java/com/mraof/minestuck/util/CoordPair.java
Dweblenod, line 1: While it seems usable for wider applications, the only current uses I can find are in the world package. Perhaps this could go there?

kirderf1: If I remember correctly, it's used for feature worldgen specifically. Putting it in the world package would make some sense, yeah. However, I do not feel fully confident in it for some reason. Because of that, I'm wondering if I should wait and see if the "util" package goes away entirely before moving it there.

Dweblenod (approved):
Im good with this being merged, although I will remind you that you were considering moving CoordPair to the world package

PR #626, opened 2024-06-30
Structure Block Registry Processor

Dweblenod:
Made changes to StructureBlockRegistryProcessor which allows StructureBlockRegistry to be called within finalizeProcess(). This makes it viable for use directly in a `template_pool`, meaning the structure block registry is suitable for jigsaw structures!

An example structure has been included for review purposes. It will need to be removed before merging.


src/main/resources/data/minestuck/worldgen/processor_list/structure_block_registry.json
```diff
@@ -0,0 +1,7 @@
+{
+  "processors": [
+    {
+      "processor_type": "minestuck:block_entity"
```
Dweblenod, line 4: I wanted to ask, why was this called block_entity instead of something like block_registry?

src/main/java/com/mraof/minestuck/world/gen/feature/StructureBlockRegistryProcessor.java
```diff
@@ -1,47 +1,49 @@
...
-		this.blocks = blocks;
 	}
 	
-	public static StructureBlockRegistryProcessor from(FeaturePlaceContext<?> context)
-	{
-		return new StructureBlockRegistryProcessor(StructureBlockRegistry.getOrDefault(context.chunkGenerator()));
-	}
-	
-	@Nullable
-	@Override	//TODO figure out blockpos difference
```
Dweblenod, line 38: Should this TODO be included in the new code?

src/main/resources/data/minestuck/worldgen/processor_list/structure_block_registry.json
kirderf1: I can't imagine that it's anything but an accident. Go ahead and rename it.

src/main/java/com/mraof/minestuck/world/gen/feature/StructureBlockRegistryProcessor.java
kirderf1: From what I recall and since the two position fields were never used in this processor, that TODO was just about figuring out more meaningful names for `blockPos` and `blockPos2`. I think that's solved in the new code, even if I still don't fully know what the technical difference between the two are. You can leave it out.

src/main/java/com/mraof/minestuck/world/gen/feature/StructureBlockRegistryProcessor.java
```diff
@@ -1,47 +1,49 @@
...
+import net.minecraft.world.level.chunk.ChunkGenerator;
+import net.minecraft.world.level.levelgen.structure.templatesystem.*;
 
-import javax.annotation.Nullable;
 import javax.annotation.ParametersAreNonnullByDefault;
+import java.util.List;
 
 @ParametersAreNonnullByDefault
 @MethodsReturnNonnullByDefault
 public class StructureBlockRegistryProcessor extends StructureProcessor
```
kirderf1, line 18:
```suggestion
public final class StructureBlockRegistryProcessor extends StructureProcessor
```
Shall we make it final while at it?

src/main/java/com/mraof/minestuck/world/gen/feature/StructureBlockRegistryProcessor.java
Dweblenod: Based on your explanation, I would theorize that the second block pos is used by the gravity processor. Regardless, I will leave it out!

kirderf1 (approved):
Code looks all good. Do you think there's a point in noting this as a datapack feature in the changelog?

kirderf1:
With everything else fine, the test structure ought to be removed.

Dweblenod:
> With everything else fine, the test structure ought to be removed.

Removed the structure files but did keep the (unused) processor_list under the assumption it can be used in the near future

PR #628, opened 2024-07-06
Structure block registry expansion

Dweblenod:
Creates new registry entries, updates existing entries in Lands with our newer blocks, and does some refactoring/reorganizing. Also fixes issues where jigsaw blocks would not get factored into the structure block registry


Dweblenod (commented):
Will need to remove the structure tester but should otherwise be essentially ready!

Dweblenod:
Should have a CHANGELOG addition, as what blocks are being placed down in Lands have changed

PR #632, opened 2024-08-11
Add helper functions to replace long repetitive methods

Dweblenod:


src/main/java/com/mraof/minestuck/data/loot_table/MSChestLootTables.java
```diff
@@ -862,4 +863,19 @@ public static ResourceLocation locationForTitle(Supplier<TitleLandType> landType
 		ResourceLocation landName = Objects.requireNonNull(LandTypes.TITLE_REGISTRY.getKey(landType.get()));
 		return baseLoot.withSuffix("/title/" + landName.getNamespace() + "/" + landName.getPath());
 	}
+	
+	public LootItemFunction.Builder defaultDamageRange()
+	{
+		return damageRange(0.75F, 1.0F);
+	}
```
kirderf1, line 870: I don't think calling it `default` makes sense in this context, even if it's the only range that is being used.
I suggest something more descriptive, renaming it to (something with) "slightly damaged" or "randomly slightly damaged".

kirderf1 (commented):
Looks good, should be good once the function name comment has been resolved.

src/main/java/com/mraof/minestuck/data/loot_table/MSChestLootTables.java
Dweblenod: Renamed it to minorDamageRange

kirderf1 (approved):

PR #634, opened 2024-09-06
Ciber's 1.20.4 patches

Cibernet83:
This pull request contains a handful of general fixes, as well as some minor changes to the Lotus Flower

Changes:
- Added Teleporter block to the Minestuck Main creative tab
- The Lotus Flower and Edit Mode Cursor are no longer considered as `LivingEntity`, meaning that they're no longer affected by damage or mob effects
- Adjusted the hitbox size for Lotus Flowers
- The Lotus Time Capsule now spawns a Lotus Flower when placed
- Lotus Time Capsule blocks now remove any Lotus Flowers above them when destroyed

Fixes:
- Restored the fix for dialogue buttons constantly playing their hover sound when selected
- `maxTicks` and `noBlockCollision` variables for `BouncingProjectileEntity` and `ReturningProjectileEntity` are now properly synced with the client. This means that chakrams and pinballs no longer immediately disappear visually, and the Umbral Infiltrator no longer jitters when its aspect ability is active
- Ogres, Basilisks, and Giclopes now use horizontal delta to determine whether or not their walking animations should play, allowing their idle walk animation to play as intended


src/main/java/com/mraof/minestuck/client/renderer/entity/ServerCursorRenderer.java
```diff
@@ -34,4 +37,10 @@ public RenderType getRenderType(ServerCursorEntity animatable, ResourceLocation
 	{
 		return RenderType.entityTranslucentCull(getTextureLocation(animatable));
 	}
+	
+	@Override
+	protected void applyRotations(ServerCursorEntity animatable, PoseStack poseStack, float ageInTicks, float rotationYaw, float partialTick)
+	{
+		poseStack.mulPose(Axis.YP.rotationDegrees(180.0F - animatable.getYRot()));
+	}
```
kirderf1, line 45: Does this have to do with the change from LivingEntity to Entity for ServerCursorEntity?

CHANGELOG.md
```diff
@@ -13,10 +13,19 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 
 ### Changed
 - Structure Block Registry Processor can now be used for data generated structures through a processor_list
+- Adjusted the hitbox size for Lotus Flowers
+- The Lotus Time Capsule now spawns a Lotus Flower when placed
+- Lotus Time Capsule blocks now remove any Lotus Flowers above them when destroyed
+
+### Fixed
+- The Umbral Infiltrator no longer jitters when thrown by a void player
```
kirderf1, line 21: Was this also a problem in 1.20.1?

kirderf1 (commented):
Just a few questions that I want to check, which will probably not require any additional changes.

src/main/java/com/mraof/minestuck/client/renderer/entity/ServerCursorRenderer.java
Cibernet83: Yeah, it does. `GeoEntityRenderer` only handles render rotations for `LivingEntity` inheritors, so the change required us to do so manually.

CHANGELOG.md
Cibernet83: Sure is. I got on the most recent public build to test it out and it appears to be the case.

kirderf1 (approved):

PR #635, opened 2024-09-07
Attribute-Dependent Underling Animations

Cibernet83:
- Walking animation speed for `AnimatedPathfinderMob`s is now adjusted by their Movement Speed attribute
- Gave Underlings and Pawns the Attack Speed attribute: it influences how fast their attack animations should play


src/main/java/com/mraof/minestuck/entity/ai/attack/AnimatedAttackWhenInRangeGoal.java
```diff
@@ -75,7 +78,7 @@ public void tick()
 	{
 		super.tick();
 		
-		if(time == phasedAnimation.getContactStartTime())
+		if(time == Math.floor(phasedAnimation.getContactStartTime(phasedAnimation.getAnimationSpeed(entity))))
```
kirderf1, line 81:
```suggestion
		if(time == phasedAnimation.getContactStartTime(phasedAnimation.getAnimationSpeed(entity)))
```
No need to floor an int.

src/main/java/com/mraof/minestuck/entity/animation/PhasedMobAnimation.java
```diff
@@ -13,47 +20,61 @@ public class PhasedMobAnimation
...
-		return contactStart;
+		return (int) Math.round(contactStart / speed);
 	}
 	
-	public int getRecoveryStartTime()
+	public int getRecoveryStartTime(double speed)
 	{
-		return recoveryStart;
+		return (int) Math.round(recoveryStart / speed);
 	}
```
kirderf1, line 78: I see calls like `phasedAnimation.getContactStartTime(phasedAnimation.getAnimationSpeed(entity))` in a lot of places. It would be nice with an entity overload for these so that you only need to call on `phasedAnimation` once:
`phasedAnimation.getContactStartTime(entity)`
Depending on where the double overloads would still be used, they could potentially be privated or even replaced.

src/main/java/com/mraof/minestuck/entity/underling/OgreEntity.java
```diff
@@ -122,16 +127,17 @@ public void initiationPhaseStart(MobAnimation.Action animation)
 	@Override
 	public void registerControllers(AnimatableManager.ControllerRegistrar controllers)
 	{
-		controllers.add(new AnimationController<>(this, "walkArmsAnimation", OgreEntity::walkArmsAnimation).setAnimationSpeed(0.3));
-		controllers.add(new AnimationController<>(this, "walkAnimation", OgreEntity::walkAnimation).setAnimationSpeed(0.3));
-		controllers.add(new AnimationController<>(this, "attackAnimation", OgreEntity::attackAnimation).setAnimationSpeed(0.5));
+		controllers.add(new AnimationController<>(this, "walkArmsAnimation", OgreEntity::walkArmsAnimation).setAnimationSpeed(0.5));
+		controllers.add(new AnimationController<>(this, "walkAnimation", OgreEntity::walkAnimation).setAnimationSpeed(0.5));
+		controllers.add(new AnimationController<>(this, "attackAnimation", OgreEntity::attackAnimation).setAnimationSpeed(ATTACK_ANIMATION_SPEED));
 		controllers.add(new AnimationController<>(this, "deathAnimation", OgreEntity::deathAnimation).setAnimationSpeed(0.85));
```
kirderf1, line 133: I wonder if `setAnimationSpeed()` even matters here for those controllers where we now update the animation speed inside the animation state handlers.

src/main/java/com/mraof/minestuck/entity/animation/PhasedMobAnimation.java
```diff
@@ -64,26 +85,36 @@ public MobAnimation getAnimation()
...
 	
+	public double getAnimationSpeed(PathfinderMob entity)
+	{
+		if(speedModifyingAttribute == null)
+			return baseAnimationSpeed;
+		
+		AttributeInstance attributeInstance = entity.getAttribute(speedModifyingAttribute);
+		return (attributeInstance != null ? attributeInstance.getValue() : 1) * baseAnimationSpeed;
+	}
```
kirderf1, line 116: It looks like you've tried tuning when the speed may change in the animation state handlers, with walk animations updating continuously while attack animations only change while an attack is not ongoing. But does that actually get reflected for stuff controlled by `PhasedMobAnimation`? What happens in here if the speed changes mid-animation?

kirderf1 (changes requested):

kirderf1:
Oh and just found something that may be relevant here: `animationController.setAnimationSpeedHandler(...)`

src/main/java/com/mraof/minestuck/entity/animation/PhasedMobAnimation.java
Cibernet83: Good catch, I went ahead and changed it so its speed now only gets updated at the start of the AI goal

Dweblenod (commented):
Overall looks pretty clean! 

The animation speed parameter in the creation of the PhasedMobAnimation is a bit awkward. We now have an attribute for attack speed (which is 1 by default for all our entities besides Pawns as of writing) and a separate value controlling animation speed. For a given mob, all the attacks have the same attack animation speed value anyways, making it seem like it holds a similar role to the attribute. Is there any way they could be combined? What is the benefit to having both?

Dweblenod (commented):
Aside from the unaddressed increase in pawn attack animation speed, I think I am ready to approve this PR!

Cibernet83:
The reason Pawns had their attack speed increased was to accommodate for their weapon's own attack speed penalty, since it affects the attack speed attribute directly.

Dweblenod (approved):

kirderf1 (approved):

PR #638, opened 2024-09-30
Update terrain ladders and bookshelves to use the appropriate CommonR…

glubtier:
- Bookshelves now use CommonRecipes.bookshelfRecipe()
-- The previous recipes would create the Vanilla bookshelf rather than the specific wood. This update allows the set-specific shelves to be crafted.
- Ladders now use CommonRecipes.ladderRecipe()
-- The previous recipes worked, and there is no change on the user's end, so this has not been mentioned in the CHANGELOG


kirderf1 (approved):

kirderf1:
@Dweblenod do you want to look at this before it is merged?

Dweblenod:
> @Dweblenod do you want to look at this before it is merged?

If its already been checked in game, Im good

kirderf1:
I have not checked it in-game myself. But the code change is simple enough, so I don't think that it needs to be checked anymore than what glubtier might have already done.

PR #639, opened 2024-09-30
Fixed Area Effect and Summoner Blocks crashing the game

Cibernet83:
- Summoner Blocks and Area Effect Blocks no longer crash the game when right-clicked with a Spawn Egg and Potion respectively.
- Localized the set Entity Type and Potion Name for both blocks' status message.
- Added to changelog fix implemented by the packet cleanup, where in previous versions of the mod right-clicking a Summoner or Area Effect Block with their corresponding items wouldn't actually update their settings.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -15,4 +15,3 @@
 /src/main/generated/resources/.cache/
 /.metadata
 /logs/
-java_pid29320.hprof
--- a/.gitignore
+++ b/.gitignore
@@ -15,3 +15,4 @@
 /src/main/generated/resources/.cache/
 /.metadata
 /logs/
+java_pid29320.hprof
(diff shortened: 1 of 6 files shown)


.gitignore
```diff
@@ -15,3 +15,4 @@
 /src/main/generated/resources/.cache/
 /.metadata
 /logs/
+java_pid29320.hprof
```
kirderf1, line 18: Do you know what this is?

Cibernet83: it's a file that showed up when i started the 1.21 port. figured i'd leave the gitignore change it in just in case

kirderf1: I gave it a search, and it is likely a log file (specifically a heap dump) from a java crash (from JVM, not from the program), _probably_ caused by an out of memory error. I would guess that the number is just an arbitrary identifier for that specific crash. You should be fine to get rid of it, or move it out of the project folder if you want to give it a closer look later.

kirderf1 (approved):
@Dweblenod if you think that this needs testing, feel free to give it a look first. Otherwise, feel free to merge this immediately.

Cibernet83:
Just so you're aware, I tested my changes before making the PR, and it all seems to be working as intended! Further testing wouldn't hurt if necessary.

Dweblenod (approved):

PR #649, opened 2025-01-05
Fix and remove smaller todo comments

kirderf1:
Goes through current todo and fixme comments, fixing or implementing a selection of them based on how small and straightforward I considered them to be. Also removed some comments that either did not need an action, was already solved, or were under-specified feature suggestions. I've also done some code cleanup related to code with a todo, even when the cleanup is not related to the todo itself.

There are still plenty of todos left after this. In this pass, I stayed away from a lot of game design related todos out of a lack of interest for design decisions.


src/main/java/com/mraof/minestuck/client/ClientProxy.java
```diff
@@ -213,28 +217,50 @@ public boolean addDestroyEffects(BlockState state, Level level, BlockPos pos, Pa
...
+			}
+			
+			@Override
+			public void modifyFogRender(Camera camera, FogRenderer.FogMode mode, float renderDistance, float partialTick, float nearDistance, float farDistance, FogShape shape)
+			{
+				RenderSystem.setShaderFogStart(-8);
+				RenderSystem.setShaderFogEnd(48F / 0.2F);
+				RenderSystem.setShaderFogShape(FogShape.SPHERE);
+			}
+		}, MSFluids.WATER_COLORS_TYPE);
```
Dweblenod, line 260: Underwater colors work but fog is seemingly not present

src/main/java/com/mraof/minestuck/client/ClientProxy.java
Dweblenod, line 1: I doubt this is new to these changes but I noticed that mobs seemingly cannot pathfind in our fluids, making them stuck even in 1 block deep sources

src/main/java/com/mraof/minestuck/Minestuck.java
```diff
@@ -69,8 +69,6 @@ public Minestuck(IEventBus eventBus, ModContainer modContainer)
 		modContainer.registerConfig(ModConfig.Type.CLIENT, MinestuckConfig.clientSpec);
 		modContainer.registerConfig(ModConfig.Type.SERVER, MinestuckConfig.serverSpec);
 		
-		//GeckoLib.initialize(eventBus); FIXME
-
```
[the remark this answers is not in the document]
kirderf1: No.
From comparing versions, it seems that what `GeckoLib.initialize()` did is now handled in the geckolib mod constructor.

src/main/java/com/mraof/minestuck/client/ClientProxy.java
kirderf1: Ah. good catch! I'll see if I can figure out what's going on with that.

kirderf1: Fix pushed.

Dweblenod (approved):

PR #650, opened 2025-01-05
Additional Resources

Dweblenod:
A new location to hold everything endorsed by Minestuck but which is not being directly integrated/supported for newer game versions.

Includes new Traditional Chinese (zh_tw) resource packs made by 4seer2 and previously posted Official Compat datapacks made by Dweblenod and Vennelli.


kirderf1:
I'm not so sure about putting these in `src/main/resources`. By having them there, I believe that they'll be copied into the mod jar. But they won't actually be used by the mod when they're there.
I suggest having a top-level folder instead, like the `unused textures` folder. They are then easier to find when navigating the github page.

Dweblenod:
Remove CHANGELOG entry, add something to README instead

PR #652, opened 2025-01-27
Data Driven Echeladder

Dweblenod:
Makes multiple aspects of the Echeladder and Rungs data driven, tweaks how Echeladder experience is gained, and makes other minor tweaks related to the Echeladder.

- There is no longer a threshold under which certain experience sources will stop contributing. After the incoming exp has been modified for the given rung, there is a 1 in 2/exp chance that 1 exp will be added.
- Each rung now directly stores:
  - Required exp to complete it
  - Boondollars gained upon completion
  - Grist capacity
  - A list of Aspect effects available
  - A list of Echeladder specific attributes, now including the underling damage modifier and underling protection modifier
  - The text color and background color when presented in the GUI
- Echeladder related attributes now have linear scaling instead of exponential. With this change there are minor tweaks to every rung and 
- Total number of rungs can be increased or decreased
- There are now customizable Echeladder exp providers. Some vanilla mobs were already hard coded to contribute, those that did had some minor tweaks to their amounts, and now more mobs will provide exp by default. Advancements can now also contribute
- Any rungs with a sufficiently dark background color in the Echeladder gui now have a white progress bar, to improve visibility
- "aspectEffects" config value has been removed


kirderf1:
This looks cool and nice! What more do you plan on doing before putting this out of draft?

And to make a suggestion: How about creating a way to add attributes and effects outside of the `rungs.json` list?
That is, something to give these outcomes:
- A way to group together attribute/effect definitions separated from the "required"/"core" rung data. (separating the attributes/effects from the core rungs helps making the attribute/effect data stand out and give an overview to them)
- A way to add rung attributes/effects by adding new data file(s). (helps datapacks add attributes/effects without redefining the whole `rungs.json` file and without datapacks overriding each other)

Dweblenod:
>     * A way to group together attribute/effect definitions separated from the "required"/"core" rung data. (separating the attributes/effects from the core rungs helps making the attribute/effect data stand out and give an overview to them)

Interesting! I was debating whether to move attributes to a separate file since at least by default they are only used on the first rung. Are you envisioning a something like that


>     * A way to add rung attributes/effects by adding new data file(s). (helps datapacks add attributes/effects without redefining the whole `rungs.json` file and without datapacks overriding each other)

Perhaps something like a "rung override"? The below would be a separate file that adds another max health attribute starting at rung 5
`{
 "rung_number": 5,
 "attributes": [
      {
        "attribute": "minecraft:generic.max_health",
        "id": "custom:additional_boost",
        "operation": "add_value",
        "rung_multiplier": 5000,
        "starting_value": 0.0
      }
 ],
 "aspect_effects": [
 ]
}`

Dweblenod:
Ah also I think I will mark this as ready for review once I have added the fall damage attribute

src/main/java/com/mraof/minestuck/client/gui/playerStats/EcheladderScreen.java
```diff
@@ -147,135 +122,176 @@ public void render(GuiGraphics guiGraphics, int mouseX, int mouseY, float partia
...
-			if(rung > Echeladder.RUNG_COUNT)
+			guiGraphics.blit(guiEcheladder, xOffset + 90, yOffset + 175 + scroll - i * RUNG_Y, 7, 242, 146, RUNG_Y);
+			
+			int y = yOffset + 177 + scroll - i * RUNG_Y;
+			int rung = scrollIndex / RUNG_Y + i;
+			if(rung > (finalRung + 1))
 				break;
 			
-			int textColor = 0xFFFFFF;
+			//TODO broke the animation for previous rung flashing
```
Dweblenod, line 174: Having trouble identifying the source

src/main/java/com/mraof/minestuck/client/gui/playerStats/EcheladderScreen.java
```diff
@@ -147,135 +122,176 @@ public void render(GuiGraphics guiGraphics, int mouseX, int mouseY, float partia
...
+			int textColor = currentRung >= rung ? textColor(rung) : 0xFFFFFF;
```
kirderf1, line 175: Looks like text color no longer takes into account `showLastRung`. This should be fixed.

src/main/java/com/mraof/minestuck/client/gui/playerStats/EcheladderScreen.java
```diff
@@ -37,105 +37,80 @@ public class EcheladderScreen extends PlayerStatsScreen
...
 	
 	@Override
 	public void render(GuiGraphics guiGraphics, int mouseX, int mouseY, float partialTicks)
 	{
-		updateScrollAndAnimation(mouseX, mouseY);
+		updateScrollAndAnimation(mouseY);
 		
 		int speedFactor = MinestuckConfig.CLIENT.echeladderAnimation.get().getSpeed();
 		int currentRung;
-		boolean showLastRung = true;
```
kirderf1, line 83: This was changed to a field, but in the process does not get properly reset. Before, it would be true unless explicitly reassigned by the code below. Now, it is whatever it was last tick (which may be false) unless explicitly reassigned in getCurrentRung(). This might be your rung animation issue.

src/main/java/com/mraof/minestuck/player/Rungs.java
kirderf1, line 1: Concern: this data is loaded server-side and is used client-side, but without being explicitly synced from server-side to client-side. This design problem needs to be resolved somehow.

src/main/java/com/mraof/minestuck/client/gui/playerStats/EcheladderScreen.java
kirderf1, line 1: Nice to see some of the constants tidied up and used more!

kirderf1 (changes requested):

src/main/java/com/mraof/minestuck/client/gui/playerStats/EcheladderScreen.java
```diff
@@ -37,105 +37,80 @@ public class EcheladderScreen extends PlayerStatsScreen
...
-	private static final int[] textColors  = new int[] {  0xFDFF2B,   0x404040,   0xB6FF00,   0x775716,   0xFFFFFF,   0xFF6A00,   0x0094FF,   0x3F3F3F,   0x007F7F,   0xB200FF,   0x7B9CB5,   0x6D9A00,   0x219621,   0x7F743F,   0xFF7F7F,   0xAF0A8C,   0x2A9659,   0xFFD8F2,   0xFFFFFF,   0xDAFF7F,   0x3459BC,   0xDFE868,   0x2AA3D3,   0x4C4C4C,   0x00D318,   0x6F22A5,   0xC4AA29,   0x237C00,   0x2D2D2D,   0xFF6721,   0x8487E0,   0x000000,   0xADADAD,   0xE24400,   0xE27609,   0x0A08A0,   0xC6A623,   0x38C151,   0xF9A640,   0x368E4A,   0xFFFFFF,   0x00A1C1,   0x6B699E,   0x2D2D2D,   0x18117A,   0xFF9028,   0x1F7C8E,   0xE25012,   0x9721E0,   0x000000};
 	
-	private static final int ladderXOffset = 163, ladderYOffset = 25;
-	private static final int rows = 12;
+	private static final int LADDER_X_OFFSET = 163, ladderYOffset = 25;
```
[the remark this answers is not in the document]
Dweblenod: Dont see any uses of that variable anymore so removing it instead!

src/main/java/com/mraof/minestuck/player/EcheladderExpSource.java
```diff
@@ -0,0 +1,207 @@
...
+import java.util.Objects;
+import java.util.function.Function;
+
+@ParametersAreNonnullByDefault
+@EventBusSubscriber(bus = EventBusSubscriber.Bus.GAME)
+public interface EcheladderExpSource
+{
+	Codec<EcheladderExpSource> CODEC = EcheladderExpSources.REGISTRY.byNameCodec().dispatch(EcheladderExpSource::codec, Function.identity());
+	
+	List<EcheladderExpSource> INSTANCE_LIST = new ArrayList<>();
```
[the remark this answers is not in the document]
Dweblenod: Moved them, but am confused as to whether the access modifier should be changed to "protected"? If so, IntelliJ suggests I make EcheladderExpSources not final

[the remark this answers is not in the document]
kirderf1: Moved some more stuff and lowered visibility.

src/main/java/com/mraof/minestuck/client/gui/playerStats/EcheladderScreen.java
kirderf1: Fixed.

src/main/java/com/mraof/minestuck/client/gui/playerStats/EcheladderScreen.java
kirderf1: Fixed.

src/main/java/com/mraof/minestuck/player/Rungs.java
kirderf1: Handled by introducing CLientRungData.

src/main/java/com/mraof/minestuck/client/gui/playerStats/EcheladderScreen.java
kirderf1: Probably fixed now.

PR #662, opened 2025-04-07
Wave function collapse

kirderf1:
Contains the main wfc system, and a demo prospit dimension showing it off.


Dweblenod (commented):
Is the plan to replace the data (biome/structure/dimension) appended with `wfc_demo` with the variants created in `glubtier:dream-moons`? If so would it make sense for this MR to target `lunar-sway:dream-moons` and revert the naming scheme?

src/main/java/com/mraof/minestuck/world/gen/structure/wfc/WFC.java
```diff
@@ -0,0 +1,320 @@
...
+					StructurePiece piece = entry.constructor().apply(pos);
+					if(piece != null)
+						pieceAccessor.addPiece(piece);
+				}
+				
+				@Override
+				public void logNoEntries(WFCUtil.CellPos cellPos)
+				{
+					BlockPos pos = transform.toBlockPos(cellPos);
+					WFC.LOGGER.warn("No entries possible at {}!", pos);
```
Dweblenod, line 255: Will recommend that this be turned into a debug message or hiding behind a config. Can cause a lot of clutter

kirderf1:
> Is the plan to replace the data (biome/structure/dimension) appended with `wfc_demo` with the variants created in `glubtier:dream-moons`? If so would it make sense for this MR to target `lunar-sway:dream-moons` and revert the naming scheme?

Not really. I've not meant for the prospit_wfc_demo stuff to be reused as-is by glubtiers dimensions. It's only included as a demonstration until the real thing has been set up.

src/main/java/com/mraof/minestuck/world/gen/structure/wfc/WFC.java
kirderf1: Principally, I would disagree with changing the log level as this is a very concrete hint towards a problem with the palette.

Practically, I don't think it matters right now. If the WFC is integrated and used properly later and there's a lot of clutter, we should primarily fix the WFC palette. If the problems at that point are acceptable but there is still clutter, then I'd be more fine with changing this at that point.

PR #667, opened 2025-04-28
1.21.1 patches

Dweblenod:
Various fixes


src/main/java/com/mraof/minestuck/inventory/musicplayer/CassetteContainerMenu.java
kirderf1, line 1: Looks solid and seems to work. Good job!

kirderf1 (commented):
While testing it out, I noticed that the cassette tooltip that names the songs were also lost during the migration to 1.21. I'll leave it up to you whether to fix it here, or leave it for someone to fix outside of this PR.

Dweblenod:
Addressed comments!

PR #668, opened 2025-04-28
added bunkers <3

saviourom:
added all three bunkers, prospit, durse, and imp. Made it so that imp bunkers are more common than prospit and durse ones, and just an FYI they are rather small due to the limitations of base structure blocks.


src/main/java/com/mraof/minestuck/world/gen/feature/StructureBlockRegistryProcessor.java
```diff
@@ -28,6 +29,13 @@ private StructureBlockRegistryProcessor()
 	{
 	}
 	
+	@Override
+	public StructureTemplate.StructureEntityInfo processEntity(LevelReader world, BlockPos seedPos, StructureTemplate.StructureEntityInfo rawEntityInfo, StructureTemplate.StructureEntityInfo entityInfo, StructurePlaceSettings placementSettings, StructureTemplate template)
+	{
+		
+		return super.processEntity(world, seedPos, rawEntityInfo, entityInfo, placementSettings, template);
+	}
+
```
Dweblenod, line 38: Remove

PR #673, opened 2025-05-23
Fix weapons having the wrong durability

### LINKED ISSUE #666: Weapons have the wrong durability
It seems weapon tiers are overriding the separate durability. For example the blacksmith hammer has 250 durability (like iron tools), but [this line](https://github.com/lunar-sway/minestuck/blob/main/src/main/java/com/mraof/minestuck/item/MSItems.java#L55) makes it seem like it should have 450.

medsal15:
Fixes #666


Dweblenod:
These latest PRs are much appreciated!

PR #675, opened 2025-06-04
Fix farmine not working as expected

### LINKED ISSUE #672: Farmine does not work
Regardless of the amount of similar blocks nearby, farmine will always break a 3x3x3 cube centered on the block mined

medsal15:
Farmine was bugged and would only mine a 3x3x3 cube of the same block, ignoring radius and block limit.

The added tag, `minestuck:farmine_break_blacklist`, replaces the hardcoded list of forbidden blocks. It only contains obsidian by default, and limits breaking to 3x3x3.

This solves the issue, buffing all farmine weapons, except for the Coppice Crusher.

I've also fixed an error in WeaponItem where setting a DestroyBlockEffect would not work when a RightClickBlockEffect is set.

Closes #672

### DIFF
--- a/src/main/java/com/mraof/minestuck/item/weapon/FarmineEffect.java
+++ b/src/main/java/com/mraof/minestuck/item/weapon/FarmineEffect.java
@@ -13,9 +13,12 @@ import net.minecraft.world.level.block.entity.BlockEntity;
 import net.minecraft.world.level.block.state.BlockState;
 import org.apache.commons.lang3.tuple.Pair;
 
+import static com.mraof.minestuck.util.MSTags.Blocks.FARMINE_BREAK_BLACKLIST;
+
 import java.util.Comparator;
 import java.util.HashMap;
 import java.util.HashSet;
+import java.util.LinkedList;
 import java.util.PriorityQueue;
 
 /**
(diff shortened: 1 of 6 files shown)


Dweblenod (commented):
FarmineEffect on the whole could use some restructuring, such as by extracting methods, performing early returns/continues, removing unused sections, and improving general flow of destroyBlock(). Do you feel comfortable refactoring in these ways, or would you prefer that be handled at a separate time by us?

src/main/java/com/mraof/minestuck/item/weapon/FarmineEffect.java
```diff
@@ -98,7 +98,7 @@ public void onDestroyBlock(ItemStack stack, Level level, BlockState blockState,
 		} else
 		{
 			//If the block is unacceptable or there's a harvestTool mismatch, cap out at a basic 3x3 area
-			if(farMineForbiddenBlocks.contains(block)
+			if(blockState.is(FARMINE_BREAK_BLACKLIST)
 					|| stack.getDestroySpeed(blockState) < ((WeaponItem) stack.getItem()).getEfficiency())
```
Dweblenod, line 102:
```suggestion
					|| blockState.getBlock().defaultDestroyTime() > ((WeaponItem) stack.getItem()).getEfficiency())
```

src/main/java/com/mraof/minestuck/item/weapon/FarmineEffect.java
Dweblenod, line 1: Please remove the unused removeAssociation()

Dweblenod, line 1: It would be wise to use "Reformat Code", which can be found under the Code tab at the top of IntelliJ

PR #679, opened 2025-07-03
Make unbreakable katana and crowbar actually unbreakable

medsal15:
Turns out a maximum durability of -1 will break the item on its first use


Dweblenod:
Add changelog and this is good to go!

PR #681, opened 2025-07-11
Conditional Rung Gating

Dweblenod:
Uses the existing Condition infrastructure to replace the Entry rung config. Now each rung can require a unique player specific Condition in order to progress.

### DIFF
--- a/src/main/java/com/mraof/minestuck/client/gui/playerStats/EcheladderScreen.java
+++ b/src/main/java/com/mraof/minestuck/client/gui/playerStats/EcheladderScreen.java
@@ -38,7 +38,6 @@ public class EcheladderScreen extends PlayerStatsScreen
 	
 	private static final ResourceLocation guiEcheladder = ResourceLocation.fromNamespaceAndPath("minestuck", "textures/gui/echeladder.png");
 	
-	private static final int LADDER_X_OFFSET = 163;
 	private static final int RUNG_Y = 14;
 	private static final int VISIBLE_RUNG_COUNT = 12;
 	
@@ -81,12 +80,9 @@ public class EcheladderScreen extends PlayerStatsScreen
 		rungBars.clear();
 		for(int i = 0; i <= ClientRungData.getFinalRungIndex(); i++)
 		{
-			Component name = Component.translatable(I18n.exists("echeladder.rung." + i) ? I18n.get("echeladder.rung." + i) : "Rung " + (i + 1));
+			Component name = I18n.exists("echeladder.rung." + i) ? Component.translatable("echeladder.rung." + i) : Component.literal("Rung " + (i + 1));
 			
-			Optional<String> tooltip = Optional.empty();
-			String description = ClientRungData.getData(i).description().orElse("");
-			if(!description.isEmpty())
-				tooltip = Optional.of(description);
+			Optional<String> tooltip = ClientRungData.getData(i).description();
 			
 			RungBar rungBar = new RungBar(xOffset + 90, yOffset + 175 - i * RUNG_Y, 146, RUNG_Y, name, tooltip, i);
 			rungBars.add(rungBar);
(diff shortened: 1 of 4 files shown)


src/main/java/com/mraof/minestuck/client/gui/playerStats/EcheladderScreen.java
```diff
@@ -65,39 +67,60 @@ public EcheladderScreen()
...
 		super.init();
-		scrollIndex = Mth.clamp((ClientPlayerData.getRung() - 8) * RUNG_Y, 0, maxScroll);
 		animatedRung = Math.max(animatedRung, lastRung);    //If you gain a rung while the gui is open, the animated rung might get higher than the lastRung. Otherwise they're always the same value.
 		fromRung = lastRung;
 		lastRung = ClientPlayerData.getRung();
+		
+		rungBars.clear();
+		for(int i = 0; i < ClientRungData.getFinalRungIndex() + 1; i++)
+		{
+			Component name = Component.translatable(I18n.exists("echeladder.rung." + i) ? I18n.get("echeladder.rung." + i) : "Rung " + (i + 1));
```
kirderf1, line 84:
```suggestion
			Component name = I18n.exists("echeladder.rung." + i) ? Component.translatable("echeladder.rung." + i) : Component.literal("Rung " + (i + 1));
```

src/main/java/com/mraof/minestuck/client/gui/playerStats/EcheladderScreen.java
```diff
@@ -65,39 +67,60 @@ public EcheladderScreen()
...
+			
+			Optional<String> tooltip = Optional.empty();
+			String description = ClientRungData.getData(i).description().orElse("");
+			if(!description.isEmpty())
+				tooltip = Optional.of(description);
```
kirderf1, line 89: Seems excessive. Is there a particular reason that you structured it this way?
```suggestion
			Optional<String> tooltip = ClientRungData.getData(i).description();
```

src/main/java/com/mraof/minestuck/client/gui/playerStats/EcheladderScreen.java
```diff
@@ -65,39 +67,60 @@ public EcheladderScreen()
...
 		
 		super.render(guiGraphics, mouseX, mouseY, partialTicks);
 		
 		drawTabs(guiGraphics);
 		
-		drawLadder(guiGraphics, currentRung);
+		rungBars.forEach(rungBar -> {
+			rungBar.setY(rungBar.initY + getScrollMod());
+			rungBar.updateVisibility();
+		});
```
kirderf1, line 115: Better to do this before the `super.render()` call, so that they're always up to date when they render.

src/main/java/com/mraof/minestuck/client/gui/playerStats/EcheladderScreen.java
```diff
@@ -332,27 +321,66 @@ public boolean mouseClicked(double xcor, double ycor, int mouseButton)
 		{
 			if(ycor >= yOffset + 35 && ycor < yOffset + 42)
 			{
-				scrollIndex = Mth.clamp(scrollIndex + RUNG_Y, 0, maxScroll);
-				isScrolling = true;
+				scroll = maxScroll;
```
kirderf1, line 324: Click-based scrolling seems kinda off right now. It also looks to have changed from what it was before (though it did seem to be lacking before too, in the sense that only the up/down buttons were press-able, and not the rest of the scrollbar)

src/main/java/com/mraof/minestuck/client/gui/playerStats/EcheladderScreen.java
```diff
@@ -332,27 +321,66 @@ public boolean mouseClicked(double xcor, double ycor, int mouseButton)
...
+			this.initY = initY;
+			this.rung = rung;
+			tooltip.ifPresent(string -> setTooltip(Tooltip.create(Component.translatable(string))));
+		}
+		
+		@Override
+		public void renderWidget(GuiGraphics guiGraphics, int mouseX, int mouseY, float partialTick)
+		{
+			int x = getX();
+			int y = getY();
```
kirderf1, line 353: The y variable is well-used, but I think x could be used more too, basically anywhere where xOffset is used in this function.

src/main/java/com/mraof/minestuck/client/gui/playerStats/EcheladderScreen.java
Dweblenod: Yeah, im not sure where it broke. This PR does nothing to change functionality and it remains the case that only the arrows can be clicked, which sends you to the very top or bottom. Otherwise you must use mouse scrolling

Dweblenod:
Addressed comments!

src/main/java/com/mraof/minestuck/client/gui/playerStats/EcheladderScreen.java
kirderf1: I'll leave this to be solved separately then.

PR #703, opened 2025-10-27
Pair of fixes

### LINKED ISSUE #705: Transportalizers become "inactive" after being broken and re-placed, rendering them unusable
To be honest I don't know if this is intentional or not seeing as this has an entire nbt tag associated with it but it still is an issue that gets in the way of the game. When you break a transportalizer and re-place it in the world, the tag "active" gets set to 0b. Visually when you open the transportalizer menu you can see that the text showing the transportalizer's ID is red. This signifies if the transportalizer is active or not. The issue with this is that even though transportalizers keep nbt data upon breaking and re-placing them, you still can't use it due to the "active" tag getting set to 0... If this is intentional then I would suggest maybe giving the player some way of toggling this state without the /data command. Although again I'm not really sure about the intent here. Regardless, it gets in the way.. like if you make a transportalizer you CAN NOT move it or else it just breaks forever..

Image

^ notice the "active" tag here. the next image is after you break and re-place it

Image

you can see this without using the data command by simply viewing the UI of the transportalizer

Image

Image

medsal15:
Fix captchalogue cameras consuming durability but not giving cards when the player inventory is full
Fix transportalizers not working after mining and placing them

Closes #705


Dweblenod:
Excellent work!

PR #736, opened 2026-05-11
Create uranium power capability

medsal15:
Creates a capability for uranium power, working just like Neoforge's EnergyStorage capability. Also converts minestuck machines to use it instead of `UraniumPowered`.

Additionally removes the hardcoded fuel value for Raw Uranium by making it data-driven. The datamap is similar to furnace fuels, and includes the Uranium Block for 10 times the fuel.

Includes a few helpful additions, such as:
- `MachineUraniumHandler` which is a pre-made uranium power handler.
- `UraniumPower.getUraniumPower` which returns the uranium power of a given item stack and `UraniumPower.hasUraniumPower` which checks if it exists.
- `UraniumPowerSlot` which replaces some `InputSlot`s and only allows items that can provide uranium power.

`UraniumPowered` can be entirely removed if needed, but is only marked as deprecated because my mod uses it.


Dweblenod:
Ooh quality! Shouldnt the uranium block only be 9 times as powerful?

medsal15:
I chose 10 times to be similar to coal and coal blocks, which smelt 8 and 80 items respectively.

Dweblenod:
The various machines that implement the uranium capability store their uranium value in the form of a `short`, but the capability handles it as an `int`. Since ints have more storage capacity than shorts, what is your opinion on change the shorts to ints?

May need to add a check to loadAdditional() to make sure they dont throw an error checking for the wrong type

PR #737, opened 2026-05-11
Pickable laptops

Fl1s:
For convenience and to differentiate it from basic computers, laptops can now be easily picked up with the ability to keep the disks, themes, and active connections


Dweblenod:
Nice work! I am noticing however that the Commit history is outdated here. In the future, make sure to Fetch to keep the Remote branches up to date, and then create your new branch from the Remote version.

PR #740, opened 2026-05-13
[Pickable Laptops]

Fl1s:


CHANGELOG.md
```diff
@@ -6,6 +6,10 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 
 ## [Unreleased]
 
+### Changed
+
+- Laptops now pickable
+
```
Dweblenod, line 12: Dont forget to add your name as a contributor for this release! I would also rephrase the message to be something more like "Laptops can now be picked up without losing their contents"

Dweblenod (commented):
Make dropItems() in ComputerBlock protected instead of private, then Override it in LaptopBlock and leave it empty. This should accomplish the same thing as the variable pickedUp,

src/main/java/com/mraof/minestuck/blockentity/ComputerBlockEntity.java
```diff
@@ -57,7 +57,7 @@ public final class ComputerBlockEntity extends BlockEntity implements ISburbComp
 	public static final String DISK_REJECT = "block.minestuck.computer.disk_reject";
 	
 	private static final Logger LOGGER = LogManager.getLogger();
-	private static final int DISK_CAPACITY = 4;
+	private static final int DISK_CAPACITY = 5;
```
Dweblenod: Set this back to 4. We should handle this in only one PR at a time

PR #744, opened 2026-05-19
[Cake Mix]

Fl1s:


Fl1s:
I decided not to add extra alchemy steps because minestuck-compat already does the exact same thing

Dweblenod:
Make an entry to the changelog and this is good to go!

PR #746, opened 2026-05-30
[Server-Player Render]

Fl1s:
* added 'editInvisibility' config that allows to disable server player render in edit mode


Fl1s:
https://github.com/user-attachments/assets/540520f1-b3e0-4d45-9e50-784a5d493617

PR #748, opened 2026-06-01
Improve WeaponItem.Builder to support adding attributes

medsal15:
Adds the ability to give attributes to weapons using WeaponItem.Builder.

Weapons that used the enemyKnockback hit effect now use the Attack Knockback attributes with the same number.


src/main/java/com/mraof/minestuck/item/weapon/WeaponItem.java
```diff
@@ -265,9 +271,27 @@ public Builder addItemUses(int duration, UseAnim action, FinishUseItemEffect...
...
+		public Builder addAttributeModifier(Holder<Attribute> attribute, AttributeModifier modifier)
+		{
+			return this.addAttributeModifier(attribute, modifier, EquipmentSlotGroup.MAINHAND);
+		}
+		
+		public Builder addAttributeModifier(Holder<Attribute> attribute, AttributeModifier modifier, EquipmentSlotGroup slot)
+		{
+			attributeBuilder.add(attribute, modifier, slot);
+			return this;
+		}
```
Dweblenod, line 290: Since all implementations have a lot of similarities, could you make a version specific to knockback?

Dweblenod (commented):
After responding to the other comment I shared, this should be good to go! Nice work as always

PR #750, opened 2026-06-29
Themed buttons

medsal15:
Adds themed buttons to computer themes, with the ability to define new ones.

Additionnally makes the theme selector's buttons use their respective themes.

If the theme has no specific button, the default button textures and colors will be used.

<details>
<summary>Examples</summary>
Themed selector buttons<br>
First page of the theme selector

Disk burner with the Joy theme<br>
Picture of the Disk Burner program, with the option to burn the server disk being highlighted
</details>


Dweblenod (commented):
Really love this! So far I only have one note, I suggest that we keep the assets associated with a theme together. That would be to move the contents at `textures/gui/theme/` and `textures/gui/sprites/theme/` into the same directory.

My thought would be to rename `gui/sprites/theme` to `gui/computer/theme` and eventually move the contents of `gui/desktop_icon` under computer as well.

medsal15:
Unfortunately, due to the way `blitSprite` works, the textures must be under `gui/sprites`. I have no idea whether it is possible to change that.

Dweblenod:
Then for the time being I will consider this ready and merge

PR #751, opened 2026-07-08
Allow Terrain Flatenator to flatten terrain

medsal15:
Affects a 3x3x3 cube around the clicked block. Works the same way as shovels do.


Dweblenod (commented):
Tangentially related to this PR, what are you thoughts on the shovels current durability and grist cost?

medsal15:
For an upgrade to the iron shovel, it is very cheap. Maybe increasing a bit the rust cost could make it more balanced.

Dweblenod:
> For an upgrade to the iron shovel, it is very cheap. Maybe increasing a bit the rust cost could make it more balanced.

I was thinking of going the other way for once, so lets plan on leaving it for now

PR #755, opened 2026-08-11
Data Checker gui improvements

Dweblenod:
Overhauls the Data Checker in the Minestuck GUI to show each Sessions data through an overview of the Incipisphere. Data should be easier to comprehend, denser, and quicker to navigate. 

Fixes minor issues like grist cache button giving poorly formatted command and scroll bar being stuck at very top or bottom.

It is also available to everyone in config by default


src/main/java/com/mraof/minestuck/client/gui/playerStats/DataCheckerScreen.java
```diff
@@ -195,372 +287,242 @@ public boolean mouseReleased(double mouseX, double mouseY, int mouseButton)
...
-		public boolean isButton()
+		private static void appendLandAndPlayer(CompoundTag landNbt, Optional<LandTypePair.Named> oNamed, MutableComponent component)
 		{
-			return true;
-		}
-		@Override
-		public String getName()
-		{
-			return String.format("%s (%d/%d)", name, playersEntered, players);
+			//TODO formatting is wrong for null lands
```
Dweblenod: Last major issue

PR #757, opened 2026-08-23
Fix sylladex voiding items when closed

medsal15:
This only happened when using the minestuck menu keybind.

The mode check can be safely removed, but may cause the editmode inventory to clutter. I still tested to check that no item is dropped.


Dweblenod:
Add to the changelog and this should be good to go!

PR #759, opened 2026-09-17
More cruxite artifacts

medsal15:
Adds 3 new Cruxite Artifacts:
- The shears artifact must be used on a shearable entity or break a block that can is more efficiently broken with swords
- The disk artifact must be used on a computer
- The disc artifact must be used on a jukebox

Changed the way cruxite artifacts are selected for a player. Instead of using an enum, artifacts are selected from the `#minestuck:cruxite_artifacts` tag. This allows addons to create their own artifacts.

Artifact models are now datagenned, and the potion artifact no longer has a fully transparent layer.

A picture of 2 rows of cruxite artifacts, with shears, disk, and disc artifacts in that order. The top artifacts are cyan, and the bottom are light blue.

<details>
<summary>Notes</summary>
If there is no valid artifact obtained, cruxite apples are given as default.

If the obtained artifact does not extend `CruxiteArtifactItem`, a warning is logged in the console. I assume whoever added the item to the tag has implemented something similar.

`onArtifactActivated` is made static, as it does not use any additional logic.

Player artifacts are now stored as a simple itemstack that is copied. Old artifacts should be loaded correctly.
</details>


src/main/java/com/mraof/minestuck/skaianet/SburbPlayerData.java
```diff
@@ -72,12 +75,18 @@ void read(CompoundTag tag)
 			this.hasEntered = tag.getBoolean("has_entered");
 		}
 		
-		this.artifactType = ArtifactType.fromInt(tag.getInt("artifact"));
+		if(tag.contains("artifact", Tag.TAG_INT))
+		{
+			this.artifactType = ArtifactType.fromInt(tag.getInt("artifact")).createItemStack();
+		} else
+		{
+			this.artifactType = ItemStack.parseOptional(pRegistries, tag);
```
Fl1s: Should be added
`tag.getCompound("artifact")`
after `ItemStack.parseOptional(pRegistries, ... <- here);`

src/main/java/com/mraof/minestuck/skaianet/SburbPlayerData.java
```diff
@@ -116,7 +125,13 @@ void readOldData(CompoundTag tag)
 			hasEntered = tag.getBoolean("has_entered");
 		}
 		
-		artifactType = ArtifactType.fromInt(tag.getInt("artifact"));
+		if(tag.contains("artifact", Tag.TAG_INT))
+		{
+			artifactType = ArtifactType.fromInt(tag.getInt("artifact")).createItemStack();
+		} else
+		{
+			artifactType = ItemStack.parseOptional(pRegistries, tag);
```
Fl1s: Should also be `artifactType = ItemStack.parseOptional(pRegistries, tag.getCompound("artifact"));`

src/main/java/com/mraof/minestuck/item/artifact/CruxiteArtifactItem.java
```diff
@@ -12,7 +12,7 @@ public CruxiteArtifactItem(Properties properties)
 		super(properties);
 	}
 	
-	public void onArtifactActivated(ServerPlayer player)
+	public static void onArtifactActivated(ServerPlayer player)
```
Fl1s: Should not be static, cuz it would replace overriding with hiding which will erase ability for subclasses to override their behavior through polimorphism (I suppose)

Fl1s (changes requested):
If the player re-enter the world, they will be able to reuse the artifacts

https://github.com/user-attachments/assets/f7b3b102-3a37-4f49-aebb-fdb768e41b7a

(compressed to be able to upload)

src/main/java/com/mraof/minestuck/item/artifact/CruxiteShearsItem.java
```diff
@@ -0,0 +1,50 @@
...
+				onArtifactActivated(serverPlayer);
+			return InteractionResult.CONSUME;
+		}
+		return InteractionResult.PASS;
+	}
+	
+	@Override
+	public boolean mineBlock(ItemStack stack, Level level, BlockState state, BlockPos pos, LivingEntity miningEntity)
+	{
+		// What do you mean there is no mineable with shears?
```
Fl1s, line 41: Lmao

Fl1s (commented):
Did the bug gone ingame? I will not be able to check it in the next couple of hours...

Otherwise, I see no other issues, good work =]

Fl1s (approved):

